PR #2, opened 2020-09-04
Implement basic topology properties.

jp7677:
Use D3D11/DXVK to loop through all exposed adapters and stop at the first NVIDIA GPU. Get the Vulkan handle  from that (DXVK) adapter to get access to Vulkan device properties.


jp7677:
Some ideas:
- Split adapter class into adapter and a factory/registry that does the initial enumeration, we might refer more than one adapter with this.  
- Define our own NvPhysicalGpuHandle to identify the correct adapter
- Support other Vendors, we might need to come up with a NVIDIA driver version for that.

PR #7, opened 2021-05-04
includes: Replace hand-copied VK headers with submodule

liam-middlebrook:
Replace the manually copied `vk_platform.h` and `vulkan_core.h` files with a
submodule for the [Vulkan-Headers](https://github.com/KhronosGroup/Vulkan-Headers) GitHub repository. Initialize the
submodule to Vulkan-Headers tag `v1.2.177` which is the latest Vulkan spec
release.


jp7677:
Hi @liam-middlebrook thanks a lot for the PR, yeah, this is a much better way to handle the headers. I’m currently away from my machine for a few days, I’ll merge when being back.
Two thoughts crossed my mind:
- there is also the Vulkan lib (vulkan-1.lib), from what I see we would need to refer to and build this from the vulkan loader repository if we wanted to get rid of the hand-copied version too. Do you know of a better way?
- would be cool if we could include the nvapi headers in the future as well using git submodule.

jp7677:
@liam-middlebrook Merged, thanks a lot. I figured that I can just use `LoadLibraryA` to avoid having `vulkan-1.lib` in the repository. Will work on that ;)

PR #10, opened 2021-05-12
Query and store supported Vulkan Device Extensions in NvapiAdapter

liam-middlebrook:
One thought that came up is whether it's better to allocate `m_deviceExtensions` like we currently are, or if it would make more sense to use `std::vector` here so that `~NvapiAdapter()` can remain trivial. What's your preference on this @jp7677 ?

### DIFF
--- a/src/sysinfo/nvapi_adapter.cpp
+++ b/src/sysinfo/nvapi_adapter.cpp
@@ -28,9 +28,32 @@ namespace dxvk {
                 reinterpret_cast<void*>(
                     GetProcAddress(vkModule, "vkGetInstanceProcAddr")));
 
+        auto vkEnumerateDeviceExtensionProperties =
+            reinterpret_cast<PFN_vkEnumerateDeviceExtensionProperties>(
+                reinterpret_cast<void*>(
+                    GetProcAddress(vkModule, "vkEnumerateDeviceExtensionProperties")));
+
         VkInstance vkInstance = VK_NULL_HANDLE;
         VkPhysicalDevice vkDevice = VK_NULL_HANDLE;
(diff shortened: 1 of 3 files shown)


jp7677:
Hi @liam-middlebrook, thanks for the PR, nice. Yeah, I would prefer using a `std::vector`, preferably `std::vector<std::string>`.

liam-middlebrook:
Okay. I updated this to store the extensions in a `std::set<std::string>`, since the set will auto-sort and should have slightly better lookup times for checking if an extension is supported.

jp7677:
Yeah, that’s even better, thanks!

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -28,16 +28,47 @@ namespace dxvk {
...
+            return false;
+
+        std::vector<VkExtensionProperties> extensions(extCount);
+        if (VK_SUCCESS != vkEnumerateDeviceExtensionProperties(vkDevice,
+                                                               nullptr,
+                                                               &extCount,
+                                                               extensions.data()))
+            return false;
 
+        for (VkExtensionProperties &ext : extensions) {
```
jp7677, line 55: What about `for (const auto& ext : extensions)` for consistency? I world also prefer to omit the brace here.

jp7677 (changes requested):
One small nit..

src/sysinfo/nvapi_adapter.cpp
liam-middlebrook: Fixed.

jp7677 (approved):

jp7677:
Merged, thanks!

PR #11, opened 2021-05-19
Update to NvAPI R465 Open Source SDK v01 && implement two NvAPI_GPU functions

liam-middlebrook:
Updates the imported NvAPI files to the latest Open Source SDK release R465 v01

Implement support for the following new NvAPI functions:
- `NvAPI_GPU_GetAdapterIdFromPhysicalGpu`
- `NvAPI_GPU_GetArchInfo`


src/sysinfo/nvapi_adapter.cpp
```diff
@@ -147,7 +157,51 @@ namespace dxvk {
...
+            memcpy(luid, &m_deviceIdProperties.deviceLUID, sizeof(*luid));
+            return true;
+        }
+        return false;
+    }
+
+    bool NvapiAdapter::GetArchitectureId(NV_GPU_ARCHITECTURE_ID *archId) const {
+        // This function only makes sense on NVIDIA GPUs, bail out early if a
+        // non-NVIDIA GPU is detected.
+        if (m_deviceProperties.vendorID != 0x10de)
```
jp7677, line 171: What about moving this test into its own method and use it in the calling method (lNvAPI_GPU_GetArchInfo)? Like:
```c++
if (!adapter-IsNvidiaGpu())
  return NvidiaDeviceNotFound(n);
```
That way we could just return NV_GPU_ARCHITECTURE_ID in this here by declaring it like
```c++
NV_GPU_ARCHITECTURE_ID NvapiAdapter::GetArchitectureId() const
```

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -147,7 +157,51 @@ namespace dxvk {
...
+            return false;
+
+        // Start at Kepler and update as features are detected
+        *archId = NV_GPU_ARCHITECTURE_GK100;
```
jp7677, line 175: We could do this the other way around, starting by testing for Ampere, return corresponding architecture when matching or performing the next test. That way we could skip some test. Thus with above in mind something like: 

```c++
// ...
if (isVkDeviceExtensionSupported(VK_KHR_FRAGMENT_SHADING_RATE_EXTENSION_NAME)
  && m_deviceFragmentShadingRateProperties.primitiveFragmentShadingRateWithMultipleViewports)
  return NV_GPU_ARCHITECTURE_GA100;

// ...
if (isVkDeviceExtensionSupported(VK_NV_SHADING_RATE_IMAGE_EXTENSION_NAME))
  return NV_GPU_ARCHITECTURE_TU100;

...
```

src/sysinfo/nvapi_adapter.h
```diff
@@ -20,13 +20,17 @@ namespace dxvk {
         [[nodiscard]] uint32_t GetGpuType() const;
         [[nodiscard]] uint32_t GetBusId() const;
         [[nodiscard]] uint32_t GetVRamSize() const;
+        [[nodiscard]] bool GetLUID(LUID *luid) const;
+        [[nodiscard]] bool GetArchitectureId(NV_GPU_ARCHITECTURE_ID *archId) const;
 
     private:
-        bool isVkDeviceExtensionSupported(std::string name);
+        bool isVkDeviceExtensionSupported(std::string name) const;
```
jp7677, line 27: Thanks!

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -147,7 +157,51 @@ namespace dxvk {
         return 0;
     }
 
-    bool NvapiAdapter::isVkDeviceExtensionSupported(const std::string name) { // NOLINT(performance-unnecessary-value-param)
+    bool NvapiAdapter::GetLUID(LUID *luid) const {
```
jp7677, line 160: I tried to avoid using pointers when not needed internally, so if possible I would prefer passing `luid` by reference here. Alternatively we could split this method into `HasValidLuid()` and `GetLuid()` and then just return the LUID in `GetLuid()`. But not sure if this would just be boilerplate code. What do you think?

jp7677 (changes requested):
Hi @liam-middlebrook thanks a lot for the updated headers and the implementation! Looks really good overall, Just a few smaller remarks.

jp7677:
PS: Would be even more cool if the header could find a place here https://github.com/NVIDIA , then we could also use `git submodule` to reference them.

vityafx:
> PS: Not sure if this is possible, but it would be even more cool if the headers could find a place here https://github.com/NVIDIA , then we could also use `git submodule` to reference them.

I thought I also tried to propose that, but I couldn't find the headers there myself. I think those are distributed only in archives.

src/sysinfo/nvapi_adapter.cpp
vityafx: Does that data have to be allocated or the pointer can just be set to internally-stored data? IIRC, the latter is possible, because most of the pointers are valid only when the nvapi is initialised, and after deinitialisation, the pointers are said to be able to be different. I don't object to it being copied though, I did that myself in my implementation of it.

src/sysinfo/nvapi_adapter.cpp
vityafx: I think when we initialise the nvapi, we already go over the adapters and can extract all the useful information during the initalisation. If there are no strict requirements to the duration of such, I think, we can just do these checks once there, and here only have a single if and return. Now, every single call to this function will have consequences of many nested calls to vulkan.

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -147,7 +157,51 @@ namespace dxvk {
...
+        if (m_deviceIdProperties.deviceLUIDValid) {
```
vityafx, line 161: That was the part I was the most curious about. I did the same in my implementation, but windows games didn't quite like it. When I tried to mock the windows behaviour of getting the LUID by hardcoding the same value, it worked better. Hence the question is this supposed to work (I am not that much of a graphics developer)? When I was returning the `deviceLUID` received from vulkan the games in wine never liked the value, but when I hardcoded it to like `0x3f7` or other constants I extracted from windows, it worked better. I know it feels like the output should be the same, but I don't know what nvidia driver for windows uses to obtain the LUID. I might need to recheck, but I thought that it simply used directx API to get it, like [that one](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-getadapterluid).  Am I wrong here? Cuz I hope I am and this vulkan's LUID should work exactly the same as in directx/dxgi.

vityafx: I think I was just doing it wrong, `0x3f7` is also returned from here and passes the same checks later.

src/sysinfo/nvapi_adapter.cpp
jp7677: > I think when we initialise the nvapi, we already go over the adapters and can extract all the useful information during the initalisation, that was what I did in my fork. If there are no strict requirements for the duration of such, I think, we can just do these checks once there, and here only have a single if and return. Now, every single call to this function will have consequences of many nested calls to vulkan.

No, this is fine here and no different than e.g. determining the memory size. We already have all the needed information/structures from Vulkan on "our" side (`vulkan-1.dll` is already unloaded here).

src/sysinfo/nvapi_adapter.cpp
jp7677: > I think I was just doing it wrong, `0x3f7` is also returned from here and passes the same checks later.

Thanks for the confirmation that the implementation works. That said, please keep the discussion here focused to the PR itself ;)

jp7677:
> > PS: Not sure if this is possible, but it would be even more cool if the headers could find a place here https://github.com/NVIDIA , then we could also use `git submodule` to reference them.
> 
> I thought I also wanted to propose that, but I couldn't find the headers there myself. I think those are distributed only in archives.

See the commit message, the origin is stated there. The readme also contains a reference. Anyway, having the headers in a public repository is just a nice-to-have.

vityafx:
> > > PS: Not sure if this is possible, but it would be even more cool if the headers could find a place here https://github.com/NVIDIA , then we could also use `git submodule` to reference them.
> > 
> > 
> > I thought I also wanted to propose that, but I couldn't find the headers there myself. I think those are distributed only in archives.
> 
> See the commit message, the origin is stated there. The readme also contains a reference. Anyway, having the headers in a public repository is just a nice-to-have.

And that's what I meant, the sources weren't available on github so that we could have a submodule :(

jp7677:
> > > > PS: Not sure if this is possible, but it would be even more cool if the headers could find a place here https://github.com/NVIDIA , then we could also use `git submodule` to reference them.
> > > 
> > > 
> > > I thought I also wanted to propose that, but I couldn't find the headers there myself. I think those are distributed only in archives.
> > 
> > 
> > See the commit message, the origin is stated there. The readme also contains a reference. Anyway, having the headers in a public repository is just a nice-to-have.
> 
> And that's what I meant, the sources weren't available on github so that we could have a submodule :(

Ah ok, sorry, I slightly misunderstood your statement ;).

src/sysinfo/nvapi_adapter.cpp
liam-middlebrook: Fixed.

src/sysinfo/nvapi_adapter.cpp
liam-middlebrook: Sure. I've changed this in the next revision.

liam-middlebrook:
I'll start a discussion internally to figure out if we'll be able to provide the NvAPI Open Source SDK as a git repository like you're asking for. Thanks for the suggestion.

jp7677 (approved):

src/sysinfo/nvapi_adapter.cpp
liam-middlebrook: I'd prefer to not split this into `HasValidLuid()` and `GetLuid()` as it just seems a bit too piecemeal for my taste (then clients of the adapter need to actively care about the specifics of LUID validity).

To me the tradeoff for code complexity vs performance benefit sways towards just doing the copy. As far as I'm aware `NvAPI_GPU_GetAdapterIdFromPhysicalGpu()` isn't expected to be in the hot path for any applications, and maybe is called a handful of times.

jp7677:
Merged, thanks a lot, also for the surrounding discussion and opinions!

PR #12, opened 2021-05-27
nvapi: Fix version in GetInterfaceVersionString

EsdrasTarsis:


jp7677:
Thanks a lot, nice catch.

PR #17, opened 2021-06-07
Extract some vulkan code into separate class.

jp7677:
Not sure yet if this improves code quality.

### DIFF
--- a/src/sysinfo/nvapi_adapter_registry.cpp
+++ b/src/sysinfo/nvapi_adapter_registry.cpp
@@ -20,7 +20,7 @@ namespace dxvk {
 
     bool NvapiAdapterRegistry::Initialize() {
         m_vulkan = std::make_unique<Vulkan>();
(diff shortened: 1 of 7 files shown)


src/sysinfo/nvapi_adapter_registry.cpp
```diff
@@ -1,10 +1,13 @@
 #include "nvapi_adapter_registry.h"
+#include "vulkan.h"
 #include "../util/util_string.h"
 #include "../util/util_log.h"
 
 namespace dxvk {
 
-    NvapiAdapterRegistry::NvapiAdapterRegistry() = default;
+    NvapiAdapterRegistry::NvapiAdapterRegistry() {
+        m_vulkan = std::make_unique<Vulkan>();
```
jp7677, line 9: Doing this in ‘Initialize’ is probably a better idea.

Issue #5, opened 2021-02-14
GPU utilization and temperatures


jp7677:
Some benchmark also measure GPU utilization and temperature to provide fancy statistics. Unfortunately some (AC-Valhalla) do not cope that well with the fact that DXVK-NVAPI does not implement those and show fancy values. This seem to have no effect on actual gameplay.

![results](https://user-images.githubusercontent.com/17219293/107877534-59ddef00-6ecd-11eb-90e1-c8630b186ae6.png)

[console-output.txt](https://github.com/jp7677/dxvk-nvapi/files/5977741/console-output.txt)

We might stub those methods and return 0. That said I'm unsure if this is a good idea since 'zero' is something very different than 'I don't know'.

An alternative might be to look into getting those value from NVML (https://developer.nvidia.com/nvidia-management-library-nvml) which is present on Linux. That said DXVK-NVAPI is a PE library, using a native library is I guess not that trivial without introducing a Wine dll for NVML.

jp7677:
See also https://github.com/Saancreed/wine-nvml

Saancreed:
I made some progress! :tada:

![NVAPI over NVML](https://user-images.githubusercontent.com/26201033/109394909-bc050e00-7929-11eb-941a-dd00dfdd3ed2.png)

The code isn't very pretty right now (it's more of a PoC) and I'm not sure if I'm mapping the values correctly (especially `NvAPI_GPU_GetDynamicPstatesInfoEx` which the documentation doesn't explain very clearly) but apparently I got enough things right to make Valhalla show real GPU utilization and temperature.

https://github.com/Saancreed/dxvk-nvapi/tree/nvml

jp7677:
@Saancreed Really cool, thanks for sharing your progress! This also needs wine-nvml installed in the prefix, correct?

Saancreed:
Yes, but it can be installed either in the prefix or globally with Wine itself. My repo has Arch PKGBUILD for the latter.

jp7677:
Thanks for the confirmation. I was _almost_ sure that this is indeed a requirement ;). Actually is there even a way to load a Linux native library directly from PE code?

Do you know if this would work for Proton, considering Proton now running inside pressure-vessel/container.

Saancreed:
> Actually is there even a way to load a Linux native library directly from PE code?

No idea, that's a question for Wine people. But considering that there is a push for splitting PE and non–PE code, I think it's quite unlikely.

[...]

PR #20, opened 2021-06-14
Add support for NVIDIA Management Library v2

jp7677:
Based on https://github.com/jp7677/dxvk-nvapi/pull/15
This addresses https://github.com/jp7677/dxvk-nvapi/issues/5

### DIFF
--- a/src/nvapi_gpu.cpp
+++ b/src/nvapi_gpu.cpp
@@ -210,18 +210,16 @@ extern "C" {
                 pDynamicPstatesInfoEx->utilization[3].bIsPresent = 1;
                 pDynamicPstatesInfoEx->utilization[3].percentage = 0;
 
-                for (int i = 4; i < NVAPI_MAX_GPU_UTILIZATIONS; ++i) {
+                for (auto i = 4U; i < NVAPI_MAX_GPU_UTILIZATIONS; i++)
                     pDynamicPstatesInfoEx->utilization[i].bIsPresent = 0;
-                }
 
                 return Ok(n, alreadyLoggedOk);
 
             case NVML_ERROR_NOT_SUPPORTED:
                 pDynamicPstatesInfoEx->flags = 0;
 
-                for (int i = 0; i < NVAPI_MAX_GPU_UTILIZATIONS; ++i) {
-                    pDynamicPstatesInfoEx->utilization[i].bIsPresent = 0;
-                }
+                for (auto& util : pDynamicPstatesInfoEx->utilization)
+                    util.bIsPresent = 0;
 
                 return NotSupported(n);
             
@@ -268,7 +266,7 @@ extern "C" {
                         pThermalSettings->count = 1;
                         pThermalSettings->sensor[0].controller = NVAPI_THERMAL_CONTROLLER_UNKNOWN;
                         pThermalSettings->sensor[0].target = NVAPI_THERMAL_TARGET_GPU;
-                        pThermalSettings->sensor[0].currentTemp = temp;
+                        pThermalSettings->sensor[0].currentTemp = static_cast<int>(temp);
                         pThermalSettings->sensor[0].defaultMaxTemp = 127;
                         pThermalSettings->sensor[0].defaultMinTemp = -256;
                         break;
--- a/src/nvapi_gpu.cpp
+++ b/src/nvapi_gpu.cpp
@@ -1,6 +1,7 @@
 #include "nvapi_private.h"
 #include "nvapi_static.h"
 #include "util/util_statuscode.h"
+#include "util/util_string.h"
 
 extern "C" {
     using namespace dxvk;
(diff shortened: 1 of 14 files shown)


jp7677:
@Saancreed I've put your commits in a separate new branch and applied some work on top of it. I haven't tested yet, but I hope that I didn't yet completely broke it (eyes-closed ;)). Feel free to let me know if it still works for you (or not) if you beat me in time. Thanks again for your work!

src/nvapi_gpu.cpp
```diff
@@ -169,4 +170,139 @@ extern "C" {
...
+
+        if (pDynamicPstatesInfoEx->version != NV_GPU_DYNAMIC_PSTATES_INFO_EX_VER)
+            return IncompatibleStructVersion(n);
+
+        auto adapter = reinterpret_cast<NvapiAdapter*>(hPhysicalGpu);
+        if (!nvapiAdapterRegistry->IsAdapter(adapter))
+            return ExpectedPhysicalGpuHandle(n);
+
+        if (!adapter->HasNvml())
+            return NoImplementation(str::format(n, ": NVML not loaded"), alreadyLoggedNoNvml);
```
Saancreed, line 193: This error can be misleading: because `adapter->HasNvml()` can be false even if NVML was loaded and initialized successfully (due to missing `nvmlDevice_t` pointer). I guess in that case we should probably return `HANDLE_INVALIDATED` but I think it would make more sense to ignore non–Nvidia GPUs instead.

src/sysinfo/nvml.cpp
```diff
@@ -0,0 +1,83 @@
...
+            if (::GetLastError() != ERROR_MOD_NOT_FOUND) // Ignore library not found
+                log::write(str::format("Loading ", nvmlModuleName, " failed with error code ", ::GetLastError()));
+
+            return;
+        }
+
+        nvmlInit_v2 =
+            reinterpret_cast<PFN_nvmlInit_v2>(
+                reinterpret_cast<void*>(
+                    ::GetProcAddress(m_nvmlModule, "nvmlInit_v2")));
```
Saancreed, line 19: I'd suggest checking if none of requested function pointers is `nullptr` before proceeding with `nvmlInit` call.

src/sysinfo/nvml.cpp
```diff
@@ -0,0 +1,83 @@
+#include "nvml.h"
+#include "../util/util_string.h"
+#include "../util/util_log.h"
+
+namespace dxvk {
+    Nvml::Nvml() {
+        const auto nvmlModuleName = "nvml.dll";
+        m_nvmlModule = ::LoadLibraryA(nvmlModuleName);
+        if (m_nvmlModule == nullptr) {
+            if (::GetLastError() != ERROR_MOD_NOT_FOUND) // Ignore library not found
```
Saancreed, line 10: `auto error = ::GetLastError();` to avoid calling it twice perhaps?

src/sysinfo/nvml.cpp
```diff
@@ -0,0 +1,83 @@
...
+            reinterpret_cast<PFN_nvmlDeviceGetUtilizationRates>(
+                reinterpret_cast<void*>(
+                    ::GetProcAddress(m_nvmlModule, "nvmlDeviceGetUtilizationRates")));
+
+        auto result = nvmlInit_v2();
+        if (result != NVML_SUCCESS) {
+            log::write(str::format("NVML loaded but initialization failed with ", ErrorString(result)));
+            ::FreeLibrary(m_nvmlModule);
+            m_nvmlModule = nullptr;
+            return;
```
Saancreed, line 51: This `return;` is redundant.

src/sysinfo/nvml.h
```diff
@@ -0,0 +1,34 @@
...
+
+    private:
+        typedef decltype(&nvmlInit_v2) PFN_nvmlInit_v2;
+        typedef decltype(&nvmlShutdown) PFN_nvmlShutdown;
+        typedef decltype(&nvmlErrorString) PFN_nvmlErrorString;
+        typedef decltype(&nvmlDeviceGetHandleByPciBusId_v2) PFN_nvmlDeviceGetHandleByPciBusId_v2;
+        typedef decltype(&nvmlDeviceGetTemperature) PFN_nvmlDeviceGetTemperature;
+        typedef decltype(&nvmlDeviceGetUtilizationRates) PFN_nvmlDeviceGetUtilizationRates;
+
+        HMODULE m_nvmlModule{};
```
Saancreed, line 26: Why not keep it as `unique_ptr`? It greatly simplifies managing such resources.

src/nvapi_private.h
```diff
@@ -16,6 +16,7 @@
 #include <d3d11_1.h>
 #include <d3d12.h>
 #include <vulkan/vulkan.h>
+#include <../inc/nvml.h>
```
Saancreed, line 19: Usually when including a file with a relative path, `"../path/to/file.h"` is used instead of `<../path/to/file.h>`.

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -173,4 +189,20 @@ namespace dxvk {
...
+
+    std::string NvapiAdapter::NvmlErrorString(nvmlReturn_t result) const {
+        return std::string(m_nvml.ErrorString(result));
+    }
+
+    nvmlReturn_t NvapiAdapter::NvmlDeviceGetTemperature(nvmlTemperatureSensors_t sensorType, unsigned int *temp) const {
+        return m_nvml.DeviceGetTemperature(m_nvmlDevice, sensorType, temp);
+    }
+
+    nvmlReturn_t NvapiAdapter::NvmlDeviceGetUtilizationRates(nvmlUtilization_t *utilization) const {
```
Saancreed, line 205: `*` should probably bind left to stay consistent with header file. Same in `unsigned int *temp` above.

src/sysinfo/nvml.h
```diff
@@ -0,0 +1,34 @@
...
+        PFN_nvmlInit_v2 nvmlInit_v2{};
+        PFN_nvmlShutdown nvmlShutdown{};
+        PFN_nvmlErrorString nvmlErrorString{};
+        PFN_nvmlDeviceGetHandleByPciBusId_v2 nvmlDeviceGetHandleByPciBusId_v2{};
+        PFN_nvmlDeviceGetTemperature nvmlDeviceGetTemperature{};
+        PFN_nvmlDeviceGetUtilizationRates nvmlDeviceGetUtilizationRates{};
```
Saancreed, line 32: Shouldn't these have `m_` prefix too? At least they won't look like calls to linked NVML.

Saancreed (commented):
Looks fine and still appears to work (at least with AC Valhalla, which calls both newly implemented functions), but I have some suggestions / questions I'd like to see answered :slightly_smiling_face:

src/nvapi_private.h
jp7677: That was indeed a typo, addressed thanks.

src/sysinfo/nvapi_adapter.cpp
jp7677: I'm torn about this one, but yeah, let's keep it like the original headers, we are doing the same in all nvapi_* files too. I tried to keep it consistent with c++ style, but often forgot when copying a new method from the headers, so yeah, keeping the original format I guess ensures more consistency.

src/sysinfo/nvml.cpp
jp7677: Nice find, addressed, thanks.

src/sysinfo/nvml.cpp
jp7677: Good idea, done.

src/sysinfo/nvml.cpp
jp7677: Addressed with the `nullptr` checks.

src/sysinfo/nvml.h
jp7677: Yes, renamed.

src/sysinfo/nvml.h
jp7677: For now I wanted to keep the nvml wrapper consistent with the Vulkan wrapper. That said, I do agree and want to return `unique_ptr`here and also in the Vulkan wrapper. I propose to do the separately once this PR is merged.

src/nvapi_gpu.cpp
jp7677: I agree that having more fine grained return codes would be nice. In reality though, I think most consumer only ever check that the return code is OK/not OK, but don't contain code paths for specific error codes. Fine grained error code are important when developing against an API, though I hope that nobody is developing against our nvapi implemenation. So I'm actually Ok here to return `NoImplemenation` for both nvml not available (no wine-nvml) or when something went wrong when loading nvml (e.g. AMD GPU or sandboxed environments like Proton). If this bites us, we can always rethink this topic.

jp7677:
@Saancreed Thanks a lot for the detailed feedback. I've addressed you points, let me know your verdict ;)

Saancreed (approved):

src/nvapi_gpu.cpp
Saancreed: It's not always technically loading/initializing NVML: a scenario where the user has more than one GPU, one from Nvidia and one from another vendor, is not that uncommon. In that case NVML will successfully load and find `nvmlDevice_t` for NV GPU but will fail to do so for the other one. I said that the error can be misleading because user could see "NVML not loaded" in the log which is not the truth, it's just that NVML can't handle non–Nvidia GPUs.

jp7677:
Merged. Thanks a lot @Saancreed for your help to get this in!

PR #21, opened 2021-06-15
Add function for NvAPI_GPU_GetVbiosVersionString

SveSop:
This grabs bios version from nvml for the graphics adapter.

### DIFF
--- a/src/sysinfo/nvapi_adapter.cpp
+++ b/src/sysinfo/nvapi_adapter.cpp
@@ -209,4 +209,8 @@ namespace dxvk {
     nvmlReturn_t NvapiAdapter::NvmlDeviceGetUtilizationRates(nvmlUtilization_t *utilization) const {
         return m_nvml.DeviceGetUtilizationRates(m_nvmlDevice, utilization);
     }
+
+    nvmlReturn_t NvapiAdapter::NvmlDeviceGetVbiosVersion(char* version, unsigned int length) const {
+        return m_nvml.DeviceGetVbiosVersion(m_nvmlDevice, version, length);
+    }
 }
--- a/src/sysinfo/nvapi_adapter.h
+++ b/src/sysinfo/nvapi_adapter.h
@@ -29,6 +29,7 @@ namespace dxvk {
         [[nodiscard]] std::string NvmlErrorString(nvmlReturn_t result) const;
         [[nodiscard]] nvmlReturn_t NvmlDeviceGetTemperature(nvmlTemperatureSensors_t sensorType, unsigned int* temp) const;
         [[nodiscard]] nvmlReturn_t NvmlDeviceGetUtilizationRates(nvmlUtilization_t* utilization) const;
+        [[nodiscard]] nvmlReturn_t NvmlDeviceGetVbiosVersion(char* version, unsigned int length) const;
 
     private:
         Vulkan& m_vulkan;
--- a/src/sysinfo/nvml.cpp
+++ b/src/sysinfo/nvml.cpp
@@ -44,12 +44,18 @@ namespace dxvk {
                 reinterpret_cast<void*>(
                     ::GetProcAddress(m_nvmlModule, "nvmlDeviceGetUtilizationRates")));
 
+        m_nvmlDeviceGetVbiosVersion =
+            reinterpret_cast<PFN_nvmlDeviceGetVbiosVersion>(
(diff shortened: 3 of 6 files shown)


src/sysinfo/nvapi_adapter.cpp
```diff
@@ -209,4 +209,7 @@ namespace dxvk {
     nvmlReturn_t NvapiAdapter::NvmlDeviceGetUtilizationRates(nvmlUtilization_t *utilization) const {
         return m_nvml.DeviceGetUtilizationRates(m_nvmlDevice, utilization);
     }
+    nvmlReturn_t NvapiAdapter::NvmlDeviceGetVbiosVersion(char *version, unsigned int length) const {
```
jp7677, line 212: Please add an empty line between methods.

src/sysinfo/nvml.cpp
```diff
@@ -91,4 +97,7 @@ namespace dxvk {
     nvmlReturn_t Nvml::DeviceGetUtilizationRates(nvmlDevice_t device, nvmlUtilization_t *utilization) const {
         return m_nvmlDeviceGetUtilizationRates(device, utilization);
     }
-}
\ No newline at end of file
+    nvmlReturn_t Nvml::DeviceGetVbiosVersion(nvmlDevice_t device, char *version, unsigned int length) const {
```
jp7677, line 100: Please add an empty line between methods.

src/sysinfo/nvapi_adapter.h
```diff
@@ -29,6 +29,7 @@ namespace dxvk {
         [[nodiscard]] std::string NvmlErrorString(nvmlReturn_t result) const;
         [[nodiscard]] nvmlReturn_t NvmlDeviceGetTemperature(nvmlTemperatureSensors_t sensorType, unsigned int* temp) const;
         [[nodiscard]] nvmlReturn_t NvmlDeviceGetUtilizationRates(nvmlUtilization_t* utilization) const;
+        [[nodiscard]] nvmlReturn_t NvmlDeviceGetVbiosVersion(char *version, unsigned int length) const;
```
jp7677, line 32: Please bind the * to left side, thus `char* version`. I know, the styles are messy in this project, though I would prefer c++ style for our own types. Same for the method implementation.

jp7677 (changes requested):

SveSop:
Yeah, i suppose it is only a "one time info" kind of thing when it comes to vbios, not like frequency or stuff like that.

A wee bit confused about that `char *version` vs `char* version`, since i .. uhm.. kinda have a hard time knowing when it is "our own types"? Not being picky, but could not just all of them change, and just leave the official headers (nvapi.h/nvml.h) intact?

src/sysinfo/nvapi_adapter.h
jp7677: Please do that also for the implementation (in the cpp file.)

jp7677:
> Yeah, i suppose it is only a "one time info" kind of thing when it comes to vbios, not like frequency or stuff like that.
> 
> A wee bit confused about that `char *version` vs `char* version`, since i .. uhm.. kinda have a hard time knowing when it is "our own types"? Not being picky, but could not just all of them change, and just leave the official headers (nvapi.h/nvml.h) intact?

I did that initially, but soon forgot that when adding a new method. So currently I keep directly copied method signatures as-is, thus "not our own", e.g. in the lop level nvapi_* files and the nvml.h/cpp since that is only a wrapper around nvml. Everything else should conform to having * on the left side. As I said, it's a bit messy :).

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -209,4 +209,8 @@ namespace dxvk {
     nvmlReturn_t NvapiAdapter::NvmlDeviceGetUtilizationRates(nvmlUtilization_t *utilization) const {
         return m_nvml.DeviceGetUtilizationRates(m_nvmlDevice, utilization);
     }
+
+    nvmlReturn_t NvapiAdapter::NvmlDeviceGetVbiosVersion(char *version, unsigned int length) const {
```
jp7677, line 213: Could you please change the `*` here too?

jp7677 (changes requested):

jp7677 (approved):

jp7677:
Merged, thanks a lot @SveSop

PR #22, opened 2021-06-17
NvAPI_GPU_GetAllClockFrequencies

SveSop:
Needs a bit of work i think. I am not really happy with the way nvml gets called 3 times for the 3 clocks. 
Since i dont really know much programming, i have not been able to slim this down, but i am sure it can be done.

I do not think we can get away without making nvml grab values 3 times, as the function for nvml is not the same as for nvapi, where nvapi just fills out whatever struct with all available clocks and returns this in `pClkFreqs`. nvml takes the calls in a 1-by-1 manner and only returns `unsigned int clock` (value without a struct?)
Since the structs for `nvmlClockType_t` is not the same as the enum for `NV_GPU_PUBLIC_CLOCK_ID` i cant seem to figure out how to make `nvmlClockType_t NVML_CLOCK_GRAPHICS` point to  `NV_GPU_PUBLIC_CLOCK_ID NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS` in any meaningful way when trying to use that in a for loop or something.

Another culprit would also be if (maybe not very likely) the graphics clock is not available for the adapter, but the memory and/or video is.. this would not be read cos it would be escaped `return NotSupported(n);` from the first `NVML_ERROR_NOT_SUPPORTED` case in `resultgpu`, and thus no more clocks would be read.

Ideas? Sorry for my horrible explanation tho.. hehe.

### DIFF
--- a/src/nvapi_gpu.cpp
+++ b/src/nvapi_gpu.cpp
@@ -348,4 +348,95 @@ extern "C" {
                 return Error(str::format(n, ": ", adapter->NvmlErrorString(result)));
         }
     }
+
+    NvAPI_Status __cdecl NvAPI_GPU_GetAllClockFrequencies(NvPhysicalGpuHandle hPhysicalGpu, NV_GPU_CLOCK_FREQUENCIES *pClkFreqs) {
+        constexpr auto n = "NvAPI_GPU_GetAllClockFrequencies";
+        static bool alreadyLoggedNoNvml = false;
+        static bool alreadyLoggedHandleInvalidated = false;
+        static bool alreadyLoggedOk = false;
+
+        if (nvapiAdapterRegistry == nullptr)
+            return ApiNotInitialized(n);
+
+        if (pClkFreqs == nullptr)
+            return InvalidArgument(n);
+
+        if (pClkFreqs->version != NV_GPU_CLOCK_FREQUENCIES_VER_1 && pClkFreqs->version != NV_GPU_CLOCK_FREQUENCIES_VER_2 && pClkFreqs->version != NV_GPU_CLOCK_FREQUENCIES_VER_3)
+            return IncompatibleStructVersion(n);
+
+        // Only check for CURRENT_FREQ, and not for the other types ie. BOOST or DEFAULT for now
+        if (pClkFreqs->ClockType != NV_GPU_CLOCK_FREQUENCIES_CURRENT_FREQ)
+            return NotSupported(n);
+
+        auto adapter = reinterpret_cast<NvapiAdapter*>(hPhysicalGpu);
+        if (!nvapiAdapterRegistry->IsAdapter(adapter))
+            return ExpectedPhysicalGpuHandle(n);
+
+        if (!adapter->HasNvml())
+            return NoImplementation(str::format(n, ": NVML not loaded"), alreadyLoggedNoNvml);
+
+        if (!adapter->HasNvmlDevice())
+            return HandleInvalidated(str::format(n, ": NVML available but current adapter is not NVML compatible"), alreadyLoggedHandleInvalidated);
+
+        // Reset all clock data for all domains
+        for (auto i = 0U; i < NVAPI_MAX_GPU_PUBLIC_CLOCKS; i++){
+            pClkFreqs->domain[i].bIsPresent = false;
+            pClkFreqs->domain[i].frequency = 0;
+        }
+
+        unsigned int clock;
+        // Seemingly we need to do nvml call on a "per clock unit" to get the clock
+        // Set the availability of the clock to TRUE and the nvml read clock in the nvapi struct
+        auto resultgpu = adapter->NvmlDeviceGetClockInfo(NVML_CLOCK_GRAPHICS, &clock);
+            switch (resultgpu) {
+                case NVML_SUCCESS:
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS].bIsPresent = true;
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS].frequency = (clock * 1000);
+                    break;
+
+                case NVML_ERROR_NOT_SUPPORTED:
+                    break;
+
+                case NVML_ERROR_GPU_IS_LOST:
+                    return HandleInvalidated(n);
+
+                default:
+                    return Error(str::format(n, ": ", adapter->NvmlErrorString(resultgpu)));
+            }
+
+        auto resultmem = adapter->NvmlDeviceGetClockInfo(NVML_CLOCK_MEM, &clock);
+            switch (resultmem) {
+                case NVML_SUCCESS:
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_MEMORY].bIsPresent = true;
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_MEMORY].frequency = (clock * 1000);
+                    break;
+
+                case NVML_ERROR_NOT_SUPPORTED:
+                    break;
+
+                case NVML_ERROR_GPU_IS_LOST:
+                    return HandleInvalidated(n);
+
+                default:
+                    return Error(str::format(n, ": ", adapter->NvmlErrorString(resultmem)));
+            }
+
+        auto resultvid = adapter->NvmlDeviceGetClockInfo(NVML_CLOCK_VIDEO, &clock);
+            switch (resultvid) {
+                case NVML_SUCCESS:
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_VIDEO].bIsPresent = true;
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_VIDEO].frequency = (clock * 1000);
+                    break;
+
+                case NVML_ERROR_NOT_SUPPORTED:
+                    break;
+
+                case NVML_ERROR_GPU_IS_LOST:
+                    return HandleInvalidated(n);
+
+                default:
+                    return Error(str::format(n, ": ", adapter->NvmlErrorString(resultvid)));
+            }
+       return Ok(n, alreadyLoggedOk);
+    }
 }
--- a/src/nvapi_interface.cpp
+++ b/src/nvapi_interface.cpp
@@ -56,6 +56,7 @@ extern "C" {
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetDynamicPstatesInfoEx)
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetThermalSettings)
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetVbiosVersionString)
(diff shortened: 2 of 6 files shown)


jp7677:
Hi @SveSop thanks for the PR and sorry for the late response. Days were a bit hot here, it wasn't really inviting to sit behind my home machine.
I guess this is used by GPU-Caps-Viewer/GPU Shark? Is there anything else that uses those calls? To be honest I'm a bit torn about adding lots of low level GPU methods that are are only useful for GPU monitoring apps. In reality I guess no one will ever use those apps on Linux more than once in Wine (to just see that it works) and always fall-back to native Linux solution, e.g. `nvidia-smi`, for GPU monitoring.

SveSop:
I know Unigine benchmarks (Superposition atleast) uses this, but for gaming it is rather useless. That is why i kinda dropped those from my "lite" branch of the old nvapi.

I understand your thoughts. On one hand, it would be a "more complete api", on the other hand "why bother.. more to maintain that nobody really uses" kind of thing.

I'm fine with closing this PR and maybe just fiddle with it on my on branch for learning... however, i would be grateful for any tips regarding those "is there a way to do this a bit smoother" if you have any :)

PS. (Somewhat unrelated to this PR) I was doing some fiddling with Star Wars - The Old Republic when launched from steam, and for some reason it has a lot of nvapi calls, seemingly unrelated to gaming performance, but nevertheless called. Some i think was trying to use nvapi to set refreshrate. I think there was calls to `NvAPI_SetRefreshRateOverride` and such. Is that something to work on, or would it (imo) maybe open up a can of worms since that kinds SETS things one would not be really in control of? (Ie. dangerous if set too high).

jp7677:
Interesting, do you know if Superposition does something visually with the frequencies? May be they are just collecting everything they can get for reporting. In that case it is I guess not that interesting.

>  if you have any :)

Yeah sure, give me some time for a review.

About setting refresh rates, I don't know if wine is even capable of setting refresh rates. Actually I wouldn't want to mess with those kind of things, chances are indeed that I just break things :).

SveSop:
super
Upper right corner shows gpu info.

SveSop:
As to setting refreshrate, i am not really sure either. It could be that wine just uses whatever refresh the X11 server is running on, and dont touch it.. However, manipulating this with nvidia-smi/nvml could possibly change this "in hardware" with interesting effects running wine or dxvk.
Have not really looked into how that would work, or what it would cause.. and i do fear it could be... problems.

jp7677:
Thanks for the screenshot. Actually seeing it, this convinced me that it is a good idea to add  `NvAPI_GPU_GetAllClockFrequencies`.

About refresh rate, yeah lets leave it like it is now until there is a useful case. This with the assumption that the game can handle that we return `nullptr` for `NvAPI_SetRefreshRateOverride`.

src/nvapi_gpu.cpp
```diff
@@ -348,4 +348,95 @@ extern "C" {
...
+        // Set the availability of the clock to TRUE and the nvml read clock in the nvapi struct
+        auto resultgpu = adapter->NvmlDeviceGetClockInfo(NVML_CLOCK_GRAPHICS, &clock);
+            switch (resultgpu) {
+                case NVML_SUCCESS:
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS].bIsPresent = TRUE;
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS].frequency = static_cast<long unsigned int>(clock * 1000);
+                    break;
+
+                case NVML_ERROR_NOT_SUPPORTED:
+                    return NotSupported(n);
```
jp7677, line 398: I think `Not supported` should just set the domain as not present, thus:
```cpp
pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS].bIsPresent = false;
pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS].frequency = 0;
```

src/nvapi_gpu.cpp
```diff
@@ -348,4 +348,95 @@ extern "C" {
...
+        }
+
+        unsigned int clock;
+        // Seemingly we need to do nvml call on a "per clock unit" to get the clock
+        // Set the availability of the clock to TRUE and the nvml read clock in the nvapi struct
+        auto resultgpu = adapter->NvmlDeviceGetClockInfo(NVML_CLOCK_GRAPHICS, &clock);
+            switch (resultgpu) {
+                case NVML_SUCCESS:
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS].bIsPresent = TRUE;
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS].frequency = static_cast<long unsigned int>(clock * 1000);
```
jp7677, line 394: Do we really need a `static_cast` here?

src/nvapi_gpu.cpp
```diff
@@ -348,4 +348,95 @@ extern "C" {
...
+        if (!adapter->HasNvml())
+            return NoImplementation(str::format(n, ": NVML not loaded"), alreadyLoggedNoNvml);
+
+        if (!adapter->HasNvmlDevice())
+            return HandleInvalidated(str::format(n, ": NVML available but current adapter is not NVML compatible"), alreadyLoggedHandleInvalidated);
+
+        // Reset all clock data for all domains
+        for (auto i = 4U; i < NVAPI_MAX_GPU_PUBLIC_CLOCKS; i++){
+            pClkFreqs->domain[i].frequency = 0;
+            pClkFreqs->domain[i].bIsPresent = FALSE;
```
jp7677, line 384: Please use c++ style, thus `false` (lowercase).

src/nvapi_gpu.cpp
```diff
@@ -348,4 +348,95 @@ extern "C" {
...
+        }
+
+        unsigned int clock;
+        // Seemingly we need to do nvml call on a "per clock unit" to get the clock
+        // Set the availability of the clock to TRUE and the nvml read clock in the nvapi struct
+        auto resultgpu = adapter->NvmlDeviceGetClockInfo(NVML_CLOCK_GRAPHICS, &clock);
+            switch (resultgpu) {
+                case NVML_SUCCESS:
+                    pClkFreqs->domain[NVAPI_GPU_PUBLIC_CLOCK_GRAPHICS].bIsPresent = TRUE;
```
jp7677, line 393: Please use c++ style, thus `true` (lowercase).

src/nvapi_gpu.cpp
```diff
@@ -348,4 +348,95 @@ extern "C" {
...
+
+                case NVML_ERROR_GPU_IS_LOST:
+                    return HandleInvalidated(n);
+
+                default:
+                    return Error(str::format(n, ": ", adapter->NvmlErrorString(resultgpu)));
+            }
+
+        auto resultmem = adapter->NvmlDeviceGetClockInfo(NVML_CLOCK_MEM, &clock);
+            switch (resultmem) {
```
jp7677, line 408: See above.

src/nvapi_gpu.cpp
```diff
@@ -348,4 +348,95 @@ extern "C" {
...
+
+                case NVML_ERROR_GPU_IS_LOST:
+                    return HandleInvalidated(n);
+
+                default:
+                    return Error(str::format(n, ": ", adapter->NvmlErrorString(resultmem)));
+            }
+
+        auto resultvid = adapter->NvmlDeviceGetClockInfo(NVML_CLOCK_VIDEO, &clock);
+            switch (resultvid) {
```
jp7677, line 425: See above.

jp7677 (changes requested):

jp7677:
See comments. I'm not super happy with the current way how we handle the `nvml` result. But your proposal is consistent with other `nvml` calls. So let's go for consistency and I may be I'll refactor this later.

src/nvapi_gpu.cpp
Saancreed: Fwiw `bIsPresent` is actually a bit field:

```c
    struct
    {
        NvU32 bIsPresent:1;         //!< Set if this domain is present on this GPU
        NvU32 reserved:31;          //!< These bits are reserved for future use.
        NvU32 frequency;            //!< Clock frequency (kHz)
    }domain[NVAPI_MAX_GPU_PUBLIC_CLOCKS];
```

I think that setting it to literal value of `0` or `1` (or `0U` / `1U`) would be least surprising here (and that's what I did in `NvAPI_GPU_GetDynamicPstatesInfoEx`).

src/nvapi_gpu.cpp
Saancreed: I think we should be able to skip the cast, I believe that the type of `frequency` (which is `NvU32`) is the same as the return type of `nvmlDeviceGetClockInfo` (which is `unsigned int`) on every platform that matters (and on 64–bit Linux, casting to `unsigned long` would actually make it too large to fit in there).

jp7677: Yes, I also think this is not needed. Usually I'm lazy and just follow what CLion (which uses clang-tidy underneath) warns me about, but I haven't looked at this branch in the IDE yet...

src/nvapi_gpu.cpp
Saancreed: Unless _all_ NVML calls fail with this error, in which case returning `NotSupported` seems to be justified.

SveSop: There is no need to explicitly set bIsPresent to anything, cos i set all to 0/false above with the for loop above - The default state being bIsPresent = false, and frequency = 0. 
Should i just `break:` instead of `return NotSupported(n);`?

jp7677: Yeah, just break seems fine.

src/nvapi_gpu.cpp
```diff
@@ -348,4 +348,95 @@ extern "C" {
...
+            return ExpectedPhysicalGpuHandle(n);
+
+        if (!adapter->HasNvml())
+            return NoImplementation(str::format(n, ": NVML not loaded"), alreadyLoggedNoNvml);
+
+        if (!adapter->HasNvmlDevice())
+            return HandleInvalidated(str::format(n, ": NVML available but current adapter is not NVML compatible"), alreadyLoggedHandleInvalidated);
+
+        // Reset all clock data for all domains
+        for (auto i = 4U; i < NVAPI_MAX_GPU_PUBLIC_CLOCKS; i++){
```
jp7677, line 382: Is an offset of 4 correct here?

src/nvapi_gpu.cpp
SveSop: I would prefer `true` / `false` for logic and human readability tho.. is it "wrong"? Changing to 0U/1U would definitely seem very confusing...

SveSop:
I am not too well versed in how the "resolved" stuff work, but pushed some changes - although i kept true/false.

src/nvapi_gpu.cpp
jp7677: Let’s go for 1/0 for now to keep it consistent with other methods. I’m quite certain that true/false is also correct because it is implicitly casted to 1/0 by the compiler. But consistency first and it is as said above the safest thing to do.

jp7677:
I have pushed this with small tweaks as https://github.com/jp7677/dxvk-nvapi/commit/fb903567bf328f45eac3cf60c32e8728fc624007

PR #23, opened 2021-06-22
vk-headers: Update external/Vulkan-Headers to 1.2.182

rgc183:

Reviewed-by: Liam Middlebrook <lmiddlebrook@nvidia.com>


jp7677:
Thanks for the contribution, looks good.
You are using different usernames for commits and GitHub interaction, this is intended?

rochaudhari:
If we do not need to be fixed now, I can try it later.

jp7677:
> If we do not need to be fixed now, I can try it later.

Sorry, I did not understood what you propose, could you please rephrase?

rochaudhari:
I need to figure out why "You are using different usernames for commits and GitHub interaction" is happening. Setting git user.name did not work. I am planning to look at it later, can we merge this ?

jp7677:
Thanks for the rewording, got it.
Actually I would prefer if you could fix this before merging. Changing this afterwards once this has been landed in master is messy. 
Since this is just one commit, using `git commit ––amend ––author=“...”` and then force push should do the trick.

rochaudhari:
Opened new one: https://github.com/jp7677/dxvk-nvapi/pull/24 issue was due to multiple accounts :/

PR #37, opened 2021-09-07
Implement NvAPI_GPU_GetPstates20 and NvAPI_GPU_GetCurrentPstate

SveSop:
These two NvAPI functions provides the needed info for Unigine Heaven and Unigine Vulkan to display gpu/mem frequency aswell as temperatures.

There does not seem as any way to get voltages from nvml atm.

### DIFF
--- a/src/nvapi_gpu.cpp
+++ b/src/nvapi_gpu.cpp
@@ -334,6 +334,43 @@ extern "C" {
         }
     }
 
+    NvAPI_Status __cdecl NvAPI_GPU_GetCurrentPstate(NvPhysicalGpuHandle hPhysicalGpu, NV_GPU_PERF_PSTATE_ID *pCurrentPstate) {
+        constexpr auto n = "NvAPI_GPU_GetCurrentPstate";
+        static bool alreadyLoggedNoNvml = false;
+        static bool alreadyLoggedHandleInvalidated = false;
+        static bool alreadyLoggedOk = false;
+
+        if (nvapiAdapterRegistry == nullptr)
+            return ApiNotInitialized(n);
+
+        if (pCurrentPstate == nullptr)
+            return InvalidArgument(n);
+
+        auto adapter = reinterpret_cast<NvapiAdapter*>(hPhysicalGpu);
+        if (!nvapiAdapterRegistry->IsAdapter(adapter))
+            return ExpectedPhysicalGpuHandle(n);
+
+        if (!adapter->HasNvml())
+            return NoImplementation(n, alreadyLoggedNoNvml);
+
+        if (!adapter->HasNvmlDevice())
+            return HandleInvalidated(str::format(n, ": NVML available but current adapter is not NVML compatible"), alreadyLoggedHandleInvalidated);
+
+        nvmlPstates_t pState;
+        auto result = adapter->GetNvmlPerformanceState(&pState);
+        switch (result) {
+            case NVML_SUCCESS:
+                *pCurrentPstate = static_cast<NV_GPU_PERF_PSTATE_ID>(pState);
+                return Ok(n, alreadyLoggedOk);
+            case NVML_ERROR_NOT_SUPPORTED:
+                return NotSupported(n);
+            case NVML_ERROR_GPU_IS_LOST:
+                return HandleInvalidated(n);
+            default:
+                return Error(str::format(n, ": ", adapter->GetNvmlErrorString(result)));
+       }
+    }
+
     NvAPI_Status __cdecl NvAPI_GPU_GetAllClockFrequencies(NvPhysicalGpuHandle hPhysicalGpu, NV_GPU_CLOCK_FREQUENCIES *pClkFreqs) {
         constexpr auto n = "NvAPI_GPU_GetAllClockFrequencies";
         static bool alreadyLoggedNotSupported = false;
--- a/src/nvapi_interface.cpp
+++ b/src/nvapi_interface.cpp
@@ -56,6 +56,7 @@ extern "C" {
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetDynamicPstatesInfoEx)
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetThermalSettings)
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetVbiosVersionString)
+        INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetCurrentPstate)
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetAllClockFrequencies)
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_Disp_GetHdrCapabilities)
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_DISP_GetDisplayIdByDisplayName)
--- a/src/sysinfo/nvapi_adapter.cpp
+++ b/src/sysinfo/nvapi_adapter.cpp
@@ -234,6 +234,10 @@ namespace dxvk {
         return m_nvml.DeviceGetVbiosVersion(m_nvmlDevice, version, length);
     }
 
+    nvmlReturn_t NvapiAdapter::GetNvmlPerformanceState(nvmlPstates_t *pState) const {
+        return m_nvml.DeviceGetPerformanceState(m_nvmlDevice, pState);
+    }
+
     nvmlReturn_t NvapiAdapter::GetNvmlDeviceClockInfo(nvmlClockType_t type, unsigned int* clock) const {
         return m_nvml.DeviceGetClockInfo(m_nvmlDevice, type, clock);
     }
--- a/src/sysinfo/nvapi_adapter.h
+++ b/src/sysinfo/nvapi_adapter.h
@@ -30,6 +30,7 @@ namespace dxvk {
         [[nodiscard]] nvmlReturn_t GetNvmlDeviceTemperature(nvmlTemperatureSensors_t sensorType, unsigned int* temp) const;
         [[nodiscard]] nvmlReturn_t GetNvmlDeviceUtilizationRates(nvmlUtilization_t* utilization) const;
         [[nodiscard]] nvmlReturn_t GetNvmlDeviceVbiosVersion(char* version, unsigned int length) const;
+        [[nodiscard]] nvmlReturn_t GetNvmlPerformanceState(nvmlPstates_t* pState) const;
         [[nodiscard]] nvmlReturn_t GetNvmlDeviceClockInfo(nvmlClockType_t type, unsigned int* clock) const;
 
     private:
--- a/src/sysinfo/nvml.cpp
+++ b/src/sysinfo/nvml.cpp
@@ -21,6 +21,7 @@ namespace dxvk {
         m_nvmlDeviceGetTemperature = GetProcAddress<PFN_nvmlDeviceGetTemperature>("nvmlDeviceGetTemperature");
         m_nvmlDeviceGetUtilizationRates = GetProcAddress<PFN_nvmlDeviceGetUtilizationRates>("nvmlDeviceGetUtilizationRates");
         m_nvmlDeviceGetVbiosVersion = GetProcAddress<PFN_nvmlDeviceGetVbiosVersion>("nvmlDeviceGetVbiosVersion");
+        m_nvmlDeviceGetPerformanceState = GetProcAddress<PFN_nvmlDeviceGetPerformanceState>("nvmlDeviceGetPerformanceState");
         m_nvmlDeviceGetClockInfo = GetProcAddress<PFN_nvmlDeviceGetClockInfo>("nvmlDeviceGetClockInfo");
 
         if (m_nvmlInit_v2 == nullptr
@@ -30,6 +31,7 @@ namespace dxvk {
             || m_nvmlDeviceGetTemperature == nullptr
             || m_nvmlDeviceGetUtilizationRates == nullptr
             || m_nvmlDeviceGetVbiosVersion == nullptr
+            || m_nvmlDeviceGetPerformanceState == nullptr
             || m_nvmlDeviceGetClockInfo == nullptr)
             log::write(str::format("NVML loaded but initialization failed"));
         else {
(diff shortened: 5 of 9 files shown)


SveSop:
I wonder if `pPstatesInfo->pstates[0]` should be `pPstatesInfo->pstates[result]` (from GetNvmlPerformanceState) so that it will point to the active pstate somehow?

pstate 0 = "max performance" (`NV_GPU_PERF_PSTATE_ID`), but in case some lower pstate, would the structure be `pPstatesInfo->pstates[2].data.single.freq_kHz` with the pstate's current gpu clock? 

I kinda never seen anything but pstate 0 when running anything 3d related, so it is not easy to verify. Afaik setting "Performance, powersave" or something in the settings does not seem to affect this much in a meaningful way.

Should i experiment a bit with this you think @jp7677 ? I do not really know if what i describe above is true - that if the adapter changes "pstate" the structure will indicate this in that way?

SveSop:
I think i somewhat found an answer to my question above with the `pPstatesInfo->pstates[0]` thing. It seems a bit funky, and i only have a GTX970 running on windows.. that means what is true for that might not be true for everything else.

Basically this: [https://github.com/SveSop/nvtt/commit/3d45566e4aa742d93ae185fb3ed11c6c82eeb1fb](https://github.com/SveSop/nvtt/commit/3d45566e4aa742d93ae185fb3ed11c6c82eeb1fb)
The result from `NvAPI_GPU_GetCurrentPstate` is not directly translatable into the `pstates[]` struct. I suspect something in the line of one adapter having 3 pstates, while another can have 5... My GTX970 had 4.
This code:
[https://github.com/SveSop/nvtt/blob/master/nvtt.c#L268-L272](https://github.com/SveSop/nvtt/blob/master/nvtt.c#L268-L272)
would most likely be different in the case an adapter have 5 pstates or 3... 

The only thing that is certain is that "currentPstate 0" is the same as "pstateID 0", so i think it would be best to "force" that.

Probably not really that interesting to figure out if a card have 20 pstates, as realistically it will be running at pstate 0 when gaming/benching anyway... Just my 2cp :)

jp7677:
From what I read (mostly https://docs.nvidia.com/gameworks/content/gameworkslibrary/coresdk/nvapi/group__gpupstate.html), `NvAPI_GPU_GetPstatesInfo20`/`NvAPI_GPU_GetPstates20` will return a static table, like 
| PState | GPU Frequency | Mem Frequency |
|---|---|---
|  P0 | 1200 | 800 |
|  P1 | 1000 | 600 |
|  P2 | 800 | 500 |

Once having this table, using `NvAPI_GPU_GetCurrentPstate` gives you the current active state (P0, P1 or P2). Having the current state you can read from the table what the current frequencies etc are.  I guess application will fetchs the table with all states and related info just once and then frequently ask what the current state is.

Looking a bit through the NVML documentation, getting the current power state seems indeed match  nicely with `nvmlDeviceGetPowerState`, but I don't see  a way to build the static table with all possible power states and related frequencies etc.

src/nvapi_gpu.cpp
```diff
@@ -260,6 +260,108 @@ extern "C" {
...
+
+        if (!adapter->HasNvml())
+            return NoImplementation(n, alreadyLoggedNoNvml);
+
+        if (!adapter->HasNvmlDevice())
+            return HandleInvalidated(str::format(n, ": NVML available but current adapter is not NVML compatible"), alreadyLoggedHandleInvalidated);
+
+        nvmlPstates_t pState;
+        auto result = adapter->GetNvmlPerformanceState(&pState);
+        *pCurrentPstate = static_cast<NV_GPU_PERF_PSTATE_ID>(result);
```
jp7677, line 287: You are casting the NVML result code (Success, failed etc) into the current perf state, this probably explains why you are only seeing pstate 0 ;)
You need to check that result is sucessful  like below and then cast `pState` to `pCurrentPstate`.

src/nvapi_gpu.cpp
```diff
@@ -260,6 +260,108 @@ extern "C" {
...
+        if (!adapter->HasNvmlDevice())
+            return HandleInvalidated(str::format(n, ": NVML available but current adapter is not NVML compatible"), alreadyLoggedHandleInvalidated);
+
+        nvmlPstates_t pState;
+        auto result = adapter->GetNvmlPerformanceState(&pState);
+
+        pPstatesInfo->numPstates = 1;                                       /* TODO: Get number of pstates total */
+        pPstatesInfo->numClocks = 2;                                        /* GPU and Mem clocks */
+        pPstatesInfo->numBaseVoltages = 0;                                  /* nvml does not get voltages */
+        pPstatesInfo->pstates[0].pstateId = static_cast<NV_GPU_PERF_PSTATE_ID>(result);
```
jp7677, line 323: This is not result but one of these: https://docs.nvidia.com/gameworks/content/gameworkslibrary/coresdk/nvapi/group__gpupstate.html#ga8f33f5167a96a722b885a705520aa25b

Filling the NVAPI `pStatesInfo` structure from the NVML `pState`structure doesn´t work. The former want to have all performance states, the latter only contains the current one.

SveSop: The thing is that as you mention above, there does not seem to be any easy way to build that structure from information from nvml, thus "faking" only having 1 numPstate seemed reasonable - Basically forcing only having Pstate=0 all the time.

Ofc. casting from `result` is fruitless.. will work on that, cos that is obviously the wrong thing to do. 

So, making a "proper" table that are supposed to be in nvapi for this function, i think is not going to work very well. Doing what i did with the table (might just aswell put `pPstatesInfo->pstates[0].pstateId = 0`  (or `NVAPI_GPU_PERF_PSTATE_P0`) is not going to make much difference, cos i do not think it will be "proper" anyway.

jp7677: Using `NVAPI_GPU_PERF_PSTATE_UNDEFINED` instead of `NVAPI_GPU_PERF_PSTATE_P0` might be the most sensible thing here, but it would be useless for applications since they cannot map the current state on it to get the frequencies etc. We could also always return `NVAPI_GPU_PERF_PSTATE_UNDEFINED`as current pstate but then it would be a lottery if you see the correct frequenties or frequencies from other pstates than the current.
I think there is no good way to solve this using nvml.

SveSop:
I force pushed this with only NvAPI_GetCurrentPstate, although the PR text is somewhat wrong now :)

tests/nvapi_system.cpp
```diff
@@ -203,6 +205,13 @@ TEST_CASE("Sysinfo methods succeed against local system", "[system]") {
         result == NVAPI_OK && frequencies.domain[NVAPI_GPU_PUBLIC_CLOCK_VIDEO].bIsPresent
             ? std::cout << std::dec << frequencies.domain[NVAPI_GPU_PUBLIC_CLOCK_VIDEO].frequency / 1000 << "MHz" << std::endl
             : std::cout << "N/A" << std::endl;
+
+        NV_GPU_PERF_PSTATE_ID currentpstate;
```
jp7677, line 209: Lets move this before before printing the different clocks, since they are a result of the current p-state.  
Please also change the order above.

src/nvapi_gpu.cpp
```diff
@@ -260,6 +260,43 @@ extern "C" {
         }
     }
 
+    NvAPI_Status __cdecl NvAPI_GPU_GetCurrentPstate(NvPhysicalGpuHandle hPhysicalGpu, NV_GPU_PERF_PSTATE_ID *pCurrentPstate) {
```
jp7677, line 263: Lets move this above `NvAPI_GPU_GetAllClockFrequencies`, thus one method down.

src/nvapi_interface.cpp
```diff
@@ -57,6 +57,7 @@ extern "C" {
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetThermalSettings)
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetVbiosVersionString)
         INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetAllClockFrequencies)
+        INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetCurrentPstate)
```
jp7677, line 60: One up please above  `INSERT_AND_RETURN_WHEN_EQUALS(NvAPI_GPU_GetAllClockFrequencies)`

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -237,4 +237,8 @@ namespace dxvk {
     nvmlReturn_t NvapiAdapter::GetNvmlDeviceClockInfo(nvmlClockType_t type, unsigned int* clock) const {
         return m_nvml.DeviceGetClockInfo(m_nvmlDevice, type, clock);
     }
+
+    nvmlReturn_t NvapiAdapter::GetNvmlPerformanceState(nvmlPstates_t *pState) const {
```
jp7677, line 241: One up please above `NvapiAdapter::GetNvmlDeviceClockInfo`

src/sysinfo/nvapi_adapter.h
```diff
@@ -31,6 +31,7 @@ namespace dxvk {
         [[nodiscard]] nvmlReturn_t GetNvmlDeviceUtilizationRates(nvmlUtilization_t* utilization) const;
         [[nodiscard]] nvmlReturn_t GetNvmlDeviceVbiosVersion(char* version, unsigned int length) const;
         [[nodiscard]] nvmlReturn_t GetNvmlDeviceClockInfo(nvmlClockType_t type, unsigned int* clock) const;
+        [[nodiscard]] nvmlReturn_t GetNvmlPerformanceState(nvmlPstates_t* pState) const;
```
jp7677, line 34: One up please.

src/sysinfo/nvml.cpp
```diff
@@ -22,6 +22,7 @@ namespace dxvk {
         m_nvmlDeviceGetUtilizationRates = GetProcAddress<PFN_nvmlDeviceGetUtilizationRates>("nvmlDeviceGetUtilizationRates");
         m_nvmlDeviceGetVbiosVersion = GetProcAddress<PFN_nvmlDeviceGetVbiosVersion>("nvmlDeviceGetVbiosVersion");
         m_nvmlDeviceGetClockInfo = GetProcAddress<PFN_nvmlDeviceGetClockInfo>("nvmlDeviceGetClockInfo");
+        m_nvmlDeviceGetPerformanceState = GetProcAddress<PFN_nvmlDeviceGetPerformanceState>("nvmlDeviceGetPerformanceState");
```
jp7677, line 25: One up.

src/sysinfo/nvml.cpp
```diff
@@ -81,6 +82,10 @@ namespace dxvk {
         return m_nvmlDeviceGetClockInfo(device, type, clock);
     }
 
+    nvmlReturn_t Nvml::DeviceGetPerformanceState(nvmlDevice_t device, nvmlPstates_t *pState) const {
```
jp7677, line 85: One up.

src/sysinfo/nvml.h
```diff
@@ -16,6 +16,7 @@ namespace dxvk {
         [[nodiscard]] virtual nvmlReturn_t DeviceGetUtilizationRates(nvmlDevice_t device, nvmlUtilization_t *utilization) const;
         [[nodiscard]] virtual nvmlReturn_t DeviceGetVbiosVersion(nvmlDevice_t device, char *version, unsigned int length) const;
         [[nodiscard]] virtual nvmlReturn_t DeviceGetClockInfo(nvmlDevice_t device, nvmlClockType_t type, unsigned int *clock) const;
+        [[nodiscard]] virtual nvmlReturn_t DeviceGetPerformanceState(nvmlDevice_t device, nvmlPstates_t *pState) const;
```
jp7677, line 19: One up.

src/sysinfo/nvml.h
```diff
@@ -26,6 +27,7 @@ namespace dxvk {
         typedef decltype(&nvmlDeviceGetUtilizationRates) PFN_nvmlDeviceGetUtilizationRates;
         typedef decltype(&nvmlDeviceGetVbiosVersion) PFN_nvmlDeviceGetVbiosVersion;
         typedef decltype(&nvmlDeviceGetClockInfo) PFN_nvmlDeviceGetClockInfo;
+        typedef decltype(&nvmlDeviceGetPerformanceState) PFN_nvmlDeviceGetPerformanceState;
```
jp7677, line 30: One up.

src/sysinfo/nvml.h
```diff
@@ -36,6 +38,7 @@ namespace dxvk {
         PFN_nvmlDeviceGetUtilizationRates m_nvmlDeviceGetUtilizationRates{};
         PFN_nvmlDeviceGetVbiosVersion m_nvmlDeviceGetVbiosVersion{};
         PFN_nvmlDeviceGetClockInfo m_nvmlDeviceGetClockInfo{};
+        PFN_nvmlDeviceGetPerformanceState m_nvmlDeviceGetPerformanceState{};
```
jp7677, line 41: One up.

tests/nvapi_system.cpp
```diff
@@ -22,6 +22,7 @@ typedef decltype(&NvAPI_GPU_GetVbiosVersionString) PFN_NvAPI_GPU_GetVbiosVersion
 typedef decltype(&NvAPI_GPU_GetDynamicPstatesInfoEx) PFN_NvAPI_GPU_GetDynamicPstatesInfoEx;
 typedef decltype(&NvAPI_GPU_GetThermalSettings) PFN_NvAPI_GPU_GetThermalSettings;
 typedef decltype(&NvAPI_GPU_GetAllClockFrequencies) PFN_NvAPI_GPU_GetAllClockFrequencies;
+typedef decltype(&NvAPI_GPU_GetCurrentPstate) PFN_NvAPI_GPU_GetCurrentPstate;
```
jp7677, line 25: One up.

tests/nvapi_system.cpp
```diff
@@ -96,6 +97,7 @@ TEST_CASE("Sysinfo methods succeed against local system", "[system]") {
     auto nvAPI_GPU_GetDynamicPstatesInfoEx = GetNvAPIProcAddress<PFN_NvAPI_GPU_GetDynamicPstatesInfoEx>(nvAPI_QueryInterface, "NvAPI_GPU_GetDynamicPstatesInfoEx");
     auto nvAPI_GPU_GetThermalSettings = GetNvAPIProcAddress<PFN_NvAPI_GPU_GetThermalSettings>(nvAPI_QueryInterface, "NvAPI_GPU_GetThermalSettings");
     auto nvAPI_GPU_GetAllClockFrequencies = GetNvAPIProcAddress<PFN_NvAPI_GPU_GetAllClockFrequencies>(nvAPI_QueryInterface, "NvAPI_GPU_GetAllClockFrequencies");
+    auto nvAPI_GPU_GetCurrentPstate = GetNvAPIProcAddress<PFN_NvAPI_GPU_GetCurrentPstate>(nvAPI_QueryInterface, "NvAPI_GPU_GetCurrentPstate");
```
jp7677, line 100: One up.

jp7677 (changes requested):
Looks good, just a nit about method ordering.
Thanks for extending the system tests, could you please also add a normal unit test like this one: https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_sysinfo.cpp#L688

jp7677 (approved):

PR #38, opened 2021-09-20
nvapi-d3d12: add caching for ID3D12GraphicsCommandListExt

rochaudhari:

### DIFF
--- a/src/d3d12/nvapi_d3d12_device.cpp
+++ b/src/d3d12/nvapi_d3d12_device.cpp
@@ -36,8 +36,8 @@ namespace dxvk {
     }
 
     bool NvapiD3d12Device::LaunchCubinShader(ID3D12GraphicsCommandList* commandList, NVDX_ObjectHandle pShader, NvU32 blockX, NvU32 blockY, NvU32 blockZ, const void* params, NvU32 paramSize) {
-        Com<ID3D12GraphicsCommandListExt> commandListExt;
-        if (FAILED(commandList->QueryInterface(IID_PPV_ARGS(&commandListExt))))
+        auto commandListExt = GetCommandListExt(commandList);
+        if(commandListExt == nullptr)
             return false;
 
         return commandListExt->LaunchCubinShader(reinterpret_cast<D3D12_CUBIN_DATA_HANDLE*>(pShader), blockX, blockY, blockZ, params, paramSize) >= 0;
@@ -79,5 +79,19 @@ namespace dxvk {
             return nullptr;
 
         return deviceExt;
+    }
+
+    Com<ID3D12GraphicsCommandListExt> NvapiD3d12Device::GetCommandListExt(ID3D12GraphicsCommandList* commandList) {
+        std::scoped_lock lock(m_CommandListMutex);
+        auto it = m_CommandListMap.find(commandList);
+        if (it!= m_CommandListMap.end())
+            return it->second;
+
+        Com<ID3D12GraphicsCommandListExt> commandListExt;
+        if (FAILED(commandList->QueryInterface(IID_PPV_ARGS(&commandListExt))))
+            return nullptr;
+
+        m_CommandListMap.emplace(commandList, commandListExt.ptr());
+        return commandListExt;
     }
 }
--- a/src/d3d12/nvapi_d3d12_device.h
+++ b/src/d3d12/nvapi_d3d12_device.h
@@ -19,8 +19,12 @@ namespace dxvk {
 
     private:
         inline static std::unordered_map<ID3D12Device*, ID3D12DeviceExt*> m_cubinDeviceMap;
+        inline static std::unordered_map<ID3D12GraphicsCommandList*, ID3D12GraphicsCommandListExt*> m_CommandListMap;
+
+        inline static std::mutex m_CommandListMutex;
 
         [[nodiscard]] static Com<ID3D12DeviceExt> GetCubinDevice(ID3D12Device* device);
         [[nodiscard]] static Com<ID3D12DeviceExt> GetDeviceExt(ID3D12Device* device, D3D12_VK_EXTENSION extension);
+        [[nodiscard]] static Com<ID3D12GraphicsCommandListExt> GetCommandListExt(ID3D12GraphicsCommandList* commandList);
     };
 }


src/d3d12/nvapi_d3d12_device.cpp
```diff
@@ -80,4 +80,17 @@ namespace dxvk {
 
         return deviceExt;
     }
+
+    Com<ID3D12GraphicsCommandListExt> NvapiD3d12Device::GetCommandListExt(ID3D12GraphicsCommandList* commandList) {
+        auto it = m_CommandListMap.find(commandList);
```
K0bin, line 85: Shouldn't this be protected by a mutex or something like that?

rochaudhari: Good point! Currently all NvAPIs are assuming they are called in thread-safe way, since all other APIs are implemented this way.

K0bin: I think it's technically necessary for the D3D11 NvAPI functions too because you could call it from different threads with different DeferredContexts. Access to the context itself isn't expected to be thread safe, so I think the global map is the only problematic part.

jp7677: Unless I'm missing something, that method looks now different, there is no longer a map for CommandLists, see https://github.com/jp7677/dxvk-nvapi/blob/master/src/d3d12/nvapi_d3d12_device.cpp#L38

For other places where the cubinDevice is looked up in the map (and also for similar maps in nvapi D3D11), from my understanding, worst thing that might happen in a concurrent scenario is that `emplace` is called twice for the same device, which should be ok.

K0bin: > worst thing that might happen in a concurrent scenario is that emplace is called twice for the same device, which should be ok.

Is that okay though? Would rehashing for example not be a problem in that case?

Even if it's okay for devices, it needs to be handled for CommandLists and for DeviceContexts on the D3D11 side.

rochaudhari: I was searching for the documentation in NvAPI for multi-threading, and could not find, so it looks like we need to add locks here. @K0bin thanks for the catch!
@jp7677 I would create separate PR for locks for others.

jp7677: > > worst thing that might happen in a concurrent scenario is that emplace is called twice for the same commandList,
 
> No, games create multiple command lists and record those on different threads. They could certainly call into NvAPI from those worker threads simultanously with different command lists. (As usual, same thing on the D3D11 side just with DeferredContexts)


From my understanding / how it was intended: The map here just acts like a cache, to prevent lots of `QueryInterface` calls. Having concurrent calls with different `commandList` would  just add multiple entries to the map and not interfere with each other. Having concurrent calls with the same `commandList` might result in a cache-miss, thus for both calls we would do the QueryInterface calls for the same `commandList` and then try to insert the same device into the map with `emplace`. From my understanding how emplace works, even when it is called twice for the same key, both calls should succeed (with the key just added once effectively.). Did I missed something?

jp7677: PS: Just to be sure, I really appreciate the feedback, changes are indeed very real that I had missed something, though I want to understand the issue.

K0bin: I don't think it's guaranteed that concurrent emplace with different keys is safe. It either needs to do rehashing or add the element to a per-bucket list. Neither of which should be thread safe.

Joshua-Ashton: This needs a mutex.

src/d3d12/nvapi_d3d12_device.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <mutex>
```
jp7677, line 3: `<mutex>` is already included in `nvapi_private.h`, no need to include it again.

src/d3d12/nvapi_d3d12_device.cpp
jp7677: @rochaudhari thanks for the updated PR, alright, convinced we need a mutex for `emplace` :). Not sure if this works elegantly, but could you try to place the lock for the `emplace` call only and not for everything in this method? (My experience is a bit limited in that area in c++.)

K0bin: > but could you try to place the lock for the emplace call only and not for everything in this method? (My experience is a bit limited in that area in c++.)

That could lead to reads while you're writing which would also break.

jp7677: Break in the sense of crash or just no result for find()?

jp7677: (Having the lock in the hot path might defeat the whole cashing thing, that’s why I’m asking)

K0bin: I think it's undefined behavior, so crashes are possible. You could use a Read-Write lock (concurrent reads are supported).

FWIW, Direct3D9, 10 and 11 have an option to use an internal mutex for every single context function and DXVK just uses a spin lock for that.

jp7677: Alright, after having a short chat on discord, seems a shared_ mutex is the way to like described here https://stackoverflow.com/questions/38061067/data-race-with-stdunordered-map-despite-locking-insertions-with-mutex

rochaudhari: Thanks for the link. I agree that read write lock is more appropriate however, I also fond this: https://stackoverflow.com/questions/50972345/when-is-stdshared-timed-mutex-slower-than-stdmutex-and-when-not-to-use-it

Which also states that: "When the read operation is only a lookup in an in-memory data structure, most likely a simple lock will outperform the read/writer solution." 
Since this locks also comes with little more cost. 
What do you think?

jp7677: Thanks, this topic really isn't the easiest one ;) Following your links, a simple mutex is indeed better.
Since I prefer simple solutions, lets do just this (actually you already did) until we face any issues in real life.

jp7677: PS: `nvapi.cpp` we use `scoped_lock`, we might wanna use that (mostly for consistency). From what I read, there is no penalty for performance and we (currently) don't need the extra capabilities of `unique_lock`

rochaudhari: It was a good learning exercise :)

jp7677:
Merged, thanks a lot, also for the discussion. Could you please create a PR for similar places too?

rochaudhari:
@jp7677 Shall I create single PR for d3d11 and d3d12 ?

PR #39, opened 2021-09-21
nvapi-d3d12, nvapi-d3d11: add locks for handling caching maps

rochaudhari:


jp7677:
Thanks for the PR. Yeah, it is fine to have one PR for both D3D11 and D3D12. If you want though, It would be nicer to have two commits in this PR, one for D3D11 and one for D3D12.

jp7677:
Thanks for splitting the commit, merged!

PR #40, opened 2021-09-25
Reverse logging logic

tannisroot:
With this change, logging will only occur if `DXVK_NVAPI_LOG_LEVEL` is set to `full`.
The benefit of this is that if some function is used *a lot*, it won't cause performance issues by default.


jp7677:
Thanks a lot, pushed as https://github.com/jp7677/dxvk-nvapi/commit/4d498bfd1c591058d8c0d4d60558dfc5f2d685ea with some modifications. Most important difference is that I went for `info`  instead of `full` to use the same severity terms like DXVK. There are no other implemented levels though.

PR #41, opened 2021-09-26
nvapi: Use __func__ macro for function names.

vityafx:
Improves logging a bit by removing hard-coded strings for function
names.

Uses __func__ due to its being portable and accessible in all the
compilers after 90s, while __FUNCTION__ isn't portable and is a gnu
extension, and we know we build the project with more or less modern
compilers. The __PRETTY_FUNCTION__ isn't used as we only need to have a
function name printed out.

### DIFF
--- a/src/nvapi.cpp
+++ b/src/nvapi.cpp
@@ -15,7 +15,7 @@ extern "C" {
     using namespace dxvk;
 
     NvAPI_Status __cdecl NvAPI_EnumLogicalGPUs(NvLogicalGpuHandle nvGPUHandle[NVAPI_MAX_LOGICAL_GPUS], NvU32 *pGpuCount) {
-        constexpr auto n = "NvAPI_EnumLogicalGPUs";
+        constexpr auto n = __func__;
 
         if (nvapiAdapterRegistry == nullptr)
             return ApiNotInitialized(n);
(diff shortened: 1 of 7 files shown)


vityafx:
I am not sure which one of these `build: `, `meta: ` and others to use here as it touches the most of the project. Please tell what the policy is, I'll adjust and update the PR.

jp7677:
Oh thanks a lot, really cool that you did the actual work 👍 
I would prefer to wait until the D3D11 DLSS PR is merged, since then there are couple of additional places where we need to do this. Ok with that?
Looking through your changes, no need to introduce `n` for very simple methods, it’s fine to use `__func__` directly.

jp7677:
Regarding the commit prefix, I would go for just `nvapi:`.

vityafx:
> I would prefer to wait until the D3D11 DLSS PR is merged, since then there are couple of additional places where we need to do this. Ok with that?
Sure, no problem.

> Looking through your changes, no need to introduce `n` for very simple methods, it’s fine to use `__func__` directly.
Yes, so I thought this was sort of a convention to have it there defined that way. I can remove it if it is used only once, of course.

> Regarding the commit prefix, I would go for just `nvapi:`.
Yes, okay.

vityafx:
Please notify me when I can/should rebase so that we could actually merge it.

jp7677:
@vityafx The DLSS PR is merged, feel to rebase and once you are ready I can review and merge. Thanks!

vityafx:
That should be it. 

I have also found some style issues like "if(` instead of `if (`, what if we use some clang tools for formatting? I am a huge fan of Rust and can't live without rustfmt and clippy, so just asking :-)

jp7677:
Merged, thanks a lot.

jp7677:
About styling, yeah, I'm trying to be consistent but sometime still misses it :)
Feel free to open a PR for places where you see weird formatting. 
I haven't yet looked into linting cpp code, but yes would be cool to have that, also as part of CI.

PR #56, opened 2021-11-16
nvapi-drs: Stub some methods related to driver settings

jp7677:
Hopefully this is not needed anywhere.


jp7677:
Lets keep the branch just in case but no need for now for this PR.

PR #65, opened 2022-01-05
DAZ-Studio

jp7677:


src/nvapi_gpu.cpp
```diff
@@ -95,6 +95,34 @@ extern "C" {
...
+            return ApiNotInitialized(n);
+
+        if (hPhysicalGpu == nullptr)
+            return InvalidArgument(n);
+
+        *pBusSlotId = 0; // just assume zero...
+
+        return Ok(n);
+    }
+
```
SveSop, line 111: Maybe you could use nvml: nvmlDeviceGetPciInfo_v3(nvmlDevice_t device, nvmlPciInfo_t *pci); and you should get the id from: pci.device ?

jp7677: From my understanding, slot ID and bus ID are unfortunately not the same.

SveSop: Ok. It would have been too easy for nVidia to not use ambiguity in their header language i guess.. :roll_eyes:

jp7677: Nah, please refrain bashing NVIDIAs choices in nvapi, at least here in this project :)
Nvapi has a history of I guess 25 years with probably lots of different developers and lots of different functionalities while still being backwards compatible all the way back. I can tell you from own experiences that it is impossible to do this without having places where history shows, especially with something evolving that fast as GPU architecture and GPU use cases.

SveSop: It is probably not a huge issue, since most usage of dxvk-nvapi would be from a GPU connected to pci-e anyway... I mean, would anyone realistically use anything less than a GTX2xx series card now? And if so - what would the REAL problems be with reporting an AGP card as pcie?

jp7677: Yeah, let’s wait first if something of this PR is needed at all, considering the first real issue is/was the adapter LUID in cuda. If possible, I would prefer not to add implementations for those two GPU_ methods. The other two methods are probably ok to add, if those make difference.

jp7677:
The method `NvAPI_GPU_GetCurrentPCIEDownstreamWidth` is also used by DAZ-Studio, but not having this implemented seems not to have any visible effects. We could implement this using NVML (https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1g6c1ca57225bd4e6a43916352044340c7), but no need currently.

PR #66, opened 2022-01-15
nvapi: Reflex support through LatencyFleX

ishitatsuyuki:
This has been tested for a while on Radeon 5700 XT, but I haven't received any user reports about NV.

Edit: apparently multiple users have been using Nvidia to run this fine.

Some implementation questions:
- Does lfx.h/cpp belongs to the sysinfo folder?
- Do we need a mock for Lfx? If so, how should we create it? Put it inside the factory?


jp7677:
Thanks a lot for the PR and the questions regarding how to wire this all together. This is what come to my mind after briefly looking at this PR:

- Move  `lfx.h` and `lfx.cpp` into a new folder `src/d3d/`. LatencyFlex is not related to sysinfo things, but to D3D, I think it deserves its own place.
-  Creation of `lfx` (`std::make_unique<Lfx>()`) should go into the `ResourceFactory`, similar to `nvml` and `vulkan`. That way we keep a single place for handling all external dependencies. This makes also makes the wiring easier for the tests. The `ResourceFactory` is then no longer just about `sysinfo` dependencies, so we should move it one level up to `src`.
- Not if the following is to much overhead: I think we might need one new class to wire the nvapi methods together with the lfx wrapper, similar what the `adapterRegistry` is for `sysinfo` and where the (currently limited) actual work might happen. May be we can call this `D3DInstance` (pretty bad name, but can't find something better now). 

About tests, yeah, I prefer to have tests for the different return codes (with and without lfx) and validations that we actually call lfx, similar like the tests for nvml related methods. The tests for `NvAPI_D3D12_GetGraphicsCapabilities returns OK` also have a similar setup.

Let me know your thoughts.

ishitatsuyuki:
I think all of these suggestions look good, and that clears a lot of my confusions about code organization. Thanks! I'll try to implement this in the next few days.

ishitatsuyuki:
Thanks for the patience, I have addressed the review comments and added some tests.

ishitatsuyuki:
CLion auto refactor went rogue in a few places, removed that and another unrelated change.

tests/nvapi_d3d.cpp
```diff
@@ -56,3 +56,60 @@ TEST_CASE("D3D methods succeed", "[.d3d]") {
...
+        SetupResourceFactory(std::move(dxgiFactory), std::move(vulkan), std::move(nvml), std::move(lfx));
+        REQUIRE(NvAPI_Initialize() == NVAPI_OK);
+
+        NV_SET_SLEEP_MODE_PARAMS_V1 params {};
+        params.version = NV_SET_SLEEP_MODE_PARAMS_VER1;
+        params.bLowLatencyMode = true;
+        REQUIRE(NvAPI_D3D_SetSleepMode(&unknown, &params) != NVAPI_OK);
+    }
+
+    SECTION("GetSleepStatus (V1) returns OK") {
```
jp7677, line 80: You can create a section for all tests with `lfx` being available, that way you only have to  `ALLOW_CALL(*lfx, IsAvailable()).RETURN(true);` only once at the beginning of that section.

tests/nvapi_d3d.cpp
```diff
@@ -56,3 +56,60 @@ TEST_CASE("D3D methods succeed", "[.d3d]") {
...
+        REQUIRE(NvAPI_Initialize() == NVAPI_OK);
+
+        NV_GET_SLEEP_STATUS_PARAMS_V1 params {};
+        params.version = NV_GET_SLEEP_STATUS_PARAMS_VER1;
+        REQUIRE(NvAPI_D3D_GetSleepStatus(&unknown, &params) == NVAPI_OK);
+    }
+
+    SECTION("SetSleepMode returns OK") {
+        ALLOW_CALL(*lfx, IsAvailable()).RETURN(true); // NOLINT(bugprone-use-after-move)
+        ALLOW_CALL(*lfx, SetTargetFrameTime(UINT64_C(0)));
```
jp7677, line 92: Since the goal is to call into `lfx` you should not just `ALLOW_CALL`, but `REQUIRE_CALL` the call to `SetTargetFrameTime`.

tests/nvapi_d3d.cpp
```diff
@@ -56,3 +56,60 @@ TEST_CASE("D3D methods succeed", "[.d3d]") {
...
+
+        NV_SET_SLEEP_MODE_PARAMS_V1 params {};
+        params.version = NV_SET_SLEEP_MODE_PARAMS_VER1;
+        params.bLowLatencyMode = true;
+        REQUIRE(NvAPI_D3D_SetSleepMode(&unknown, &params) == NVAPI_OK);
+    }
+
+    SECTION("Sleep calls LFX bridge function") {
+        ALLOW_CALL(*lfx, IsAvailable()).RETURN(true); // NOLINT(bugprone-use-after-move)
+        ALLOW_CALL(*lfx, SetTargetFrameTime(UINT64_C(0)));
```
jp7677, line 104: This should also be `REQUIRE_CALL`.

tests/nvapi_d3d.cpp
jp7677, line 80: There is no V2, so you can omit the V1.

tests/nvapi_d3d.cpp
```diff
@@ -56,3 +56,60 @@ TEST_CASE("D3D methods succeed", "[.d3d]") {
...
+    }
+
+    SECTION("GetSleepStatus (V1) returns OK") {
+        ALLOW_CALL(*lfx, IsAvailable()).RETURN(true); // NOLINT(bugprone-use-after-move)
+        SetupResourceFactory(std::move(dxgiFactory), std::move(vulkan), std::move(nvml), std::move(lfx));
+        REQUIRE(NvAPI_Initialize() == NVAPI_OK);
+
+        NV_GET_SLEEP_STATUS_PARAMS_V1 params {};
+        params.version = NV_GET_SLEEP_STATUS_PARAMS_VER1;
+        REQUIRE(NvAPI_D3D_GetSleepStatus(&unknown, &params) == NVAPI_OK);
```
jp7677, line 87: As far as I understood NVAPI, one should always use the latest version of a struct, thus:
```c++
NV_GET_SLEEP_STATUS_PARAMS params {};
params.version = NV_GET_SLEEP_STATUS_PARAMS_VER;
REQUIRE(NvAPI_D3D_GetSleepStatus(&unknown, &params) == NVAPI_OK);
```

tests/nvapi_d3d.cpp
```diff
@@ -56,3 +56,60 @@ TEST_CASE("D3D methods succeed", "[.d3d]") {
...
+        SetupResourceFactory(std::move(dxgiFactory), std::move(vulkan), std::move(nvml), std::move(lfx));
+        REQUIRE(NvAPI_Initialize() == NVAPI_OK);
+
+        NV_SET_SLEEP_MODE_PARAMS_V1 params {};
+        params.version = NV_SET_SLEEP_MODE_PARAMS_VER1;
+        params.bLowLatencyMode = true;
+        REQUIRE(NvAPI_D3D_SetSleepMode(&unknown, &params) == NVAPI_OK);
+    }
+
+    SECTION("Sleep calls LFX bridge function") {
```
jp7677, line 102: Could you please improve the description of this test?

src/d3d/nvapi_d3d_instance.cpp
```diff
@@ -0,0 +1,37 @@
+#include "nvapi_d3d_instance.h"
+
+namespace dxvk {
+    NvapiD3dInstance::NvapiD3dInstance(ResourceFactory &resourceFactory): m_resourceFactory(resourceFactory) {
+
+    }
+
+    NvapiD3dInstance::~NvapiD3dInstance() {
```
jp7677, line 8: Please use the following:
```c++
NvapiD3dInstance::~NvapiD3dInstance() = default;
```

src/d3d/nvapi_d3d_instance.cpp
```diff
@@ -0,0 +1,37 @@
+#include "nvapi_d3d_instance.h"
+
+namespace dxvk {
+    NvapiD3dInstance::NvapiD3dInstance(ResourceFactory &resourceFactory): m_resourceFactory(resourceFactory) {
```
jp7677, line 4: Please use the following format for consistency:
```c++
NvapiD3dInstance::NvapiD3dInstance(ResourceFactory &resourceFactory)
    : m_resourceFactory(resourceFactory) {}
```

jp7677 (changes requested):
Thanks a lot for improving this PR. I've added some review comments, mostly just smaller things. Overall this looks already really good.

src/d3d/lfx.cpp
```diff
@@ -0,0 +1,42 @@
+#include "lfx.h"
+
+#include <windows.h>
+#include "../util/util_string.h"
+#include "../util/util_log.h"
+
+namespace dxvk {
+    Lfx::Lfx() {
+        const auto lfxModuleName = "latencyflex_wine.dll";
```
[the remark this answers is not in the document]
ishitatsuyuki: I agree that having a unified dll name helps. But since this involves a breaking change, I would like to have `latencyflex_wine.dll` as a fallback when we get this released (it can be removed once enough time has passed for the users to update). Does it sound good?

I will update the LatencyFleX side with `latencyflex_layer.dll` becoming the primary bridge, and `latencyflex_wine.dll` as a pure (no native code) export forwarder. This involves a few build system changes, which I'll try to test out in the next few days.

src/nvapi_d3d.cpp
```diff
@@ -87,4 +88,47 @@ extern "C" {
...
+        }
+        auto f = str::format(n, " (", pSetSleepModeParams->bLowLatencyMode ? "Enable" : "Disable", ")");
+        if (pSetSleepModeParams->bLowLatencyMode && !nvapiD3dInstance->IsReflexAvailable()) {
+            log::write(str::format(n, ": tried to enable Reflex but Wine LFX module is not available!"));
+        }
+        nvapiD3dInstance->SetReflexEnabled(pSetSleepModeParams->bLowLatencyMode);
+        if (pSetSleepModeParams->bLowLatencyMode) {
+            log::write(str::format(n, ": interval ", pSetSleepModeParams->minimumIntervalUs, " us"));
+            constexpr uint64_t kNanoInMicro = 1000;
+            nvapiD3dInstance->SetTargetFrameTime(pSetSleepModeParams->minimumIntervalUs * kNanoInMicro);
```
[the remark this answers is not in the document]
ishitatsuyuki: I kept the interval logging here. Let me know if that should be moved or removed as well.

tests/nvapi_d3d.cpp
ishitatsuyuki: See if this one is better.

ishitatsuyuki:
Thanks for another iteration of review comments, the code now feels a bit more cleaner. I have pushed v2 to address the review comments, plus a rebase to also apply the refactor into newly added tests.

I'll push v3 when I have addressed https://github.com/jp7677/dxvk-nvapi/pull/66#discussion_r790311468 (and any extra review comments if you have).

tests/nvapi_tests.cpp
```diff
@@ -14,8 +14,8 @@
 #include "../src/nvapi_mosaic.cpp"
 
 #include "nvapi_sysinfo_mocks.cpp"
-#include "nvapi_sysinfo_util.cpp"
```
jp7677, line 17: Is this change needed? Same with the change two lines below?

tests/nvapi_d3d.cpp
```diff
@@ -56,3 +56,61 @@ TEST_CASE("D3D methods succeed", "[.d3d]") {
...
+    auto e = ConfigureDefaultTestEnvironment(*dxgiFactory, *vulkan, *nvml, *lfx, adapter, output);
+
+    SECTION("Enabling Reflex does not return OK when LFX is unavailable") {
+        SetupResourceFactory(std::move(dxgiFactory), std::move(vulkan), std::move(nvml), std::move(lfx));
+        REQUIRE(NvAPI_Initialize() == NVAPI_OK);
+
+        NV_SET_SLEEP_MODE_PARAMS params {};
+        params.version = NV_SET_SLEEP_MODE_PARAMS_VER;
+        params.bLowLatencyMode = true;
+        REQUIRE(NvAPI_D3D_SetSleepMode(&unknown, &params) != NVAPI_OK);
```
jp7677, line 77: Could you please do this for all three methods? The require can be changed to `== NVAPI_NO_IMPLEMENTATION`.

tests/nvapi_d3d.cpp
```diff
@@ -56,3 +56,61 @@ TEST_CASE("D3D methods succeed", "[.d3d]") {
...
+    auto dxgiFactory = std::make_unique<DXGIFactory1Mock>();
+    auto vulkan = std::make_unique<VulkanMock>();
+    auto nvml = std::make_unique<NvmlMock>();
+    auto lfx = std::make_unique<LfxMock>();
+    DXGIDxvkAdapterMock adapter;
+    DXGIOutputMock output;
+
+    auto e = ConfigureDefaultTestEnvironment(*dxgiFactory, *vulkan, *nvml, *lfx, adapter, output);
+
+    SECTION("Enabling Reflex does not return OK when LFX is unavailable") {
```
jp7677, line 70: What about: `LatencyFlex depending methods return no-implementation when LatencyFlex is not available`?

jp7677 (changes requested):

tests/nvapi_d3d.cpp
jp7677: Yes, thanks

src/nvapi_d3d.cpp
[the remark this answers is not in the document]
jp7677: What about add that info the OK log statement, similar like you did with Enable/Disable?
That said, how often is this method called? Since DXVK-NVAPI unfortunately only knows about info logging, I like to keep the logging limited.

src/d3d/lfx.cpp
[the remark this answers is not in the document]
jp7677: Yes, that sounds like a sensible approach, I'm fine with looking for `latencyflex_layer.dll` and retrying with `latencyflex_wine.dll` for a limited time. Please also add a comment (incl. an issue or commit in your repo as a reference?) why the fallback is needed.

jp7677:
> Thanks for another iteration of review comments, the code now feels a bit more cleaner. I have pushed v2 to address the review comments, plus a rebase to also apply the refactor into newly added tests.
> 
> I'll push v3 when I have addressed [#66 (comment)](https://github.com/jp7677/dxvk-nvapi/pull/66#discussion_r790311468) (and any extra review comments if you have).

Hi @ishitatsuyuki thanks again for your changes. Please see my answers to your questions and some other smaller comments. I hope your are not yet tired about all the nits and tests, we are really close ;)

ishitatsuyuki:
Just a status update: I'm still working through refactoring the binding DLLs but I'm a bit short on time due to exams. Can't promise when I can get back to this, but I'll try to not delay this too much.

tests/nvapi_tests.cpp
ishitatsuyuki: I think it deserves a rename since it now handles Lfx as well which is not a sysinfo component. I can revert and keep the filename if desired.

ishitatsuyuki (commented):
v3 changes are now pushed. A LFX update with new DLL naming will be published soon after I sort out some documentation changes.

src/d3d/lfx.cpp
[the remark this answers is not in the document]
ishitatsuyuki: I added a comment (`// Try fallback entrypoints. These were used by an older version of LatencyFleX.`) but it might deserve more explanation or links for better tracking. Not sure if an issue or a commit link is the best, let me address that in a later revision...

tests/nvapi_tests.cpp
jp7677: May be I'm missing it, but as far as I'm seeing it, there are no renames in the test files. The other renames should be covered with the existing includes in this file.

src/d3d/lfx.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <windows.h>
+#include "../util/util_string.h"
+#include "../util/util_log.h"
+
+namespace dxvk {
+    Lfx::Lfx() {
+        const auto lfxModuleName = "latencyflex_layer.dll";
+        const auto lfxModuleNameFallback = "latencyflex_wine.dll";
+        m_lfxModule = ::LoadLibraryA(lfxModuleName);
+        if (m_lfxModule) {
```
jp7677, line 12: I think the fallback can be more elegant. something in the line of:
```c++
    const auto lfxModuleName = "latencyflex_layer.dll";
    const auto lfxModuleNameFallback = "latencyflex_wine.dll";
    m_lfxModule = ::LoadLibraryA(lfxModuleName);
    if (m_lfxModule == nullptr && ::GetLastError() == ERROR_MOD_NOT_FOUND) 
        m_lfxModule = ::LoadLibraryA(lfxFallbackModuleName);
    }
    if (m_nvmlModule == nullptr) {
        auto lastError = ::GetLastError();
        if (lastError != ERROR_MOD_NOT_FOUND) // Ignore library not found
            log::write(str::format("Loading ", nvmlModuleName, " failed with error code: ", lastError));

        return;
    }

    m_lfx_WaitAndBeginFrame = reinterpret_cast<PFN_lfx_WaitAndBeginFrame>(GetProcAddress(m_lfxModule, "lfx_WaitAndBeginFrame"));
    m_lfx_SetTargetFrameTime = reinterpret_cast<PFN_lfx_SetTargetFrameTime>(GetProcAddress(m_lfxModule, "lfx_SetTargetFrameTime"));
```

jp7677 (changes requested):

tests/nvapi_tests.cpp
jp7677: 4Ah, scratch that. I'm seeing the rename now, missed it because I had flagged the file as already seen. All good!

src/d3d/lfx.cpp
jp7677: Alternatively using a conditional operator  like 
```c++
m_lfx_WaitAndBeginFrame = reinterpret_cast<PFN_lfx_WaitAndBeginFrame>(
    GetProcAddress(m_lfxModule, !fallback ? "lfx_WaitAndBeginFrame" : "winelfx_WaitAndBeginFrame"));
```

src/d3d/lfx.cpp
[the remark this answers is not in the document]
ishitatsuyuki: I added a link to the commit too.

ishitatsuyuki:
Pushed v4, reorganizing the fallback logic. I also added back the cast of `GetProcAddress` result to `void*` since the compiler gives a false positive warning otherwise.

jp7677 (approved):

jp7677:
All good now and merged, thanks a lot for the contribution and for your patience!

Issue #76, opened 2022-03-24
God of War crash


K0bin:
God of War recently received an update that causes it to crash when NVAPI is enabled.

`PROTON_ENABLE_NVAPI=1`=> crash
`PROTON_ENABLE_NVAPI=0`=> works

There's a couple of unknown functions mentioned in the log but they don't seem to be part of the open source NVAPI headers as far as I can tell.

[dxvk-nvapi.log](https://github.com/jp7677/dxvk-nvapi/files/8337831/dxvk-nvapi.log)

jp7677:
Thanks for the report and for providing the logs. Though unfortunately there is nothing really obvious in it.

Do you see the same behavior with lets say DXVK-NVAPI v0.5.0  (https://github.com/jp7677/dxvk-nvapi/releases/tag/v0.5)? There are some DRS (driver settings) calls in that logs which I added (stubbed) recently , may be I screwed up there.

Just to be sure, the crash happens inside GoW and nvapi is not part of the stack trace?

K0bin:
It only crashes with the NVAPI dll shipped by Proton Experimental Bleeding Edge. If I remove that DLL with either your 1.5.3 release build or my own build of master, it works.

jp7677:
Nice find. Is only beading edge affected or does normal Proton or Proton Experimental have the same issue?

K0bin:
Normal Proton is fine. I haven't tried regular experimental.

jp7677:
Do you have the chance to try regular experimental? Experimental and bleading-edge contain the same DXVK-NVAPI source, I guess having a differences (or not) between those two versions makes it easier to pinpoint the actual issue.

K0bin:
Experimental works.

jp7677:
@K0bin @ivyl Proton beading-edge should be good again with DXVK-NVAPI from master. Thanks a lot for finding and helping to resolve this issue!

K0bin:
Nice, great work everyone. :)

ivyl:
[This has just landed in the bleeding-edge](https://github.com/ValveSoftware/Proton/commits/experimental-bleeding-edge-7.0-12609-20220328-p974ec7-w8fb04d-debdaf9-v2e8fb2). I think that manual bumps for dxvk-nvapi are good enough, but let me know if you disagree and would like to have automated updates like we have for vkd3d-proton and dxvk.

jp7677:
[...]

PR #78, opened 2022-03-25
nvapi: Copy into NVAPI-Shortstring without writing out of bounds

jp7677:
DXVK_NVAPI_VERSION comes from the VCS tag, we should not asume that it is always short. The same applies for adapter and display names.

Fixes https://github.com/jp7677/dxvk-nvapi/issues/76

### DIFF
--- a/src/util/util_string.cpp
+++ b/src/util/util_string.cpp
@@ -35,4 +35,9 @@ namespace dxvk::str {
         auto w = std::wstring(reinterpret_cast<wchar_t*>(nvus));
         return {w.begin(), w.end()};
     }
+
+    void tonvss(NvAPI_ShortString nvss, std::string str) {
+        str.resize(NVAPI_SHORT_STRING_MAX);
+        strcpy(nvss, str.c_str());
+    }
 }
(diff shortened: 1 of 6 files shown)


src/util/util_string.cpp
```diff
@@ -35,4 +35,9 @@ namespace dxvk::str {
         auto w = std::wstring(reinterpret_cast<wchar_t*>(nvus));
         return {w.begin(), w.end()};
     }
+
+    void tonvss(NvAPI_ShortString nvss, std::string str) {
+        str.resize(NVAPI_SHORT_STRING_MAX);
```
ivyl, line 40: shouldn't this be `NVAPI_SHORT_STRING_MAX-1` so we have space for the terminating NUL?

```c++
failcraft ivyl ~ % cat test.cpp
#include <iostream>
#include <string>

int main()
{
	std::string str("1234567890");
	str.resize(2);
	std::cout << str.c_str() << std::endl;
	return 0;
}
failcraft ivyl ~ % g++ test.cpp -o test && ./test
12
```

ivyl:
Thanks for taking over and creating a fix that's more in line with the style of your project. I went ahead and closed my PR.

Turns out that the `-1` is indeed necessary. I've just tested the version without it and God of War still crashes, most likely not liking not terminated strings. After adding it the game works :-)

jp7677:
Thanks a lot for taking time for a retest, I've pushed the ` - 1`, should be good now.

PR #87, opened 2022-06-06
includes: Add missing standard library includes

Saancreed:
Fixes build errors on GCC 12.

By the way, we should also fix this:

```
[1/38] Generating version.h with a custom command
fatal: unsafe repository ('/github/workspace' is owned by someone else)
To add an exception for this directory, call:
	git config --global --add safe.directory /github/workspace
```

…, otherwise builds from GH Actions could report unexpected version to callers.

### DIFF
--- a/src/nvapi_private.h
+++ b/src/nvapi_private.h
@@ -12,6 +12,10 @@
 #include <fstream>
 #include <ctime>
 #include <mutex>
+#include <memory>
+#include <unordered_map>
+#include <utility>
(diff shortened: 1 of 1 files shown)


jp7677:
Thanks a lot! Merged and fixed the git permissions before that in https://github.com/jp7677/dxvk-nvapi/commit/3052a78a4d3e3c798286d759a7f3a8c03b3bd691 . All should be good again.

PR #89, opened 2022-06-06
NVML 515

Saancreed:
An attempt to reimplement a few NVAPI functions using better NVML counterparts from newer API versions. Not tested yet because the only Windows machine I have at hand is a very small VM I set up on my laptop and doing GPU passthrough is annoying so I wanted to gather some feedback and ensure you're willing to accept a PR like this before actually getting my hands dirty :sweat_smile:

Some notes:
* I still need to fix/update tests
* we no longer require all NVML function to be present in the loaded library
* old implementations are kept as a fallback in case new functions are not available
* `nvmlDeviceGetDynamicPstatesInfo` and `nvmlDeviceGetThermalSettings` seem to be almost direct "ports" of `NvAPI_GPU_GetDynamicPstatesInfoEx` and `NvAPI_GPU_GetThermalSettings` respectively, but apparently they are not exactly the same (pending some testing):
  * `nvmlDeviceGetDynamicPstatesInfo` has `flags` which are "reserved for future use", while `NvAPI_GPU_GetDynamicPstatesInfoEx` has one already documented flag
  * thermal controller enums have known values ordered differently and NVML knows about a few more than NVAPI
  * it's undocumented/unknown whether `nvmlDeviceGetThermalSettings` accepts `NVAPI_THERMAL_TARGET_ALL` for `sensorIndex` to retrieve info for all sensors like `NvAPI_GPU_GetThermalSettings` does
  * returned temperature values are unsigned, making `nvmlDeviceGetThermalSettings` closer to V1 version of `NvAPI_GPU_GetThermalSettings`, fortunately we should be able to safely cast values from expected range to signed integers for V2 and just never report negative values this way
* there are a few more low hanging fruits which can be implemented using new functions:
  * `NvAPI_GPU_GetCurrentPCIEDownstreamWidth` using `nvmlDeviceGetCurrPcieLinkWidth` (wasn't this previously stubbed out?)
  * `NvAPI_GPU_GetIRQ` using `nvmlDeviceGetIrqNum`
  * `NvAPI_GPU_GetGpuCoreCount` using `nvmlDeviceGetNumGpuCores`

In the meantime, I'm working on the [next version of `wine-nvml`](https://github.com/Saancreed/wine-nvml/pull/6) that uses Wine's unixlib syscall interface and automatically generates code from NVIDIA's API header, which in theory should make adding support for new functions as they get released much easier in the future.

### DIFF
--- a/src/nvapi_gpu.cpp
+++ b/src/nvapi_gpu.cpp
@@ -115,6 +115,32 @@ extern "C" {
         if (!nvapiAdapterRegistry->IsAdapter(adapter))
             return ExpectedPhysicalGpuHandle(n);
 
+        if (adapter->HasNvmlDevice()) {
+            nvmlBusType_t busType;
+            auto result = adapter->GetNvmlDeviceBusType(&busType);
+            if (result == NVML_SUCCESS) {
+                switch (busType) {
+                    case NVML_BUS_TYPE_PCI:
+                        *pBusType = NVAPI_GPU_BUS_TYPE_PCI;
+                        break;
+                    case NVML_BUS_TYPE_PCIE:
+                        *pBusType = NVAPI_GPU_BUS_TYPE_PCI_EXPRESS;
+                        break;
+                    case NVML_BUS_TYPE_FPCI:
+                        *pBusType = NVAPI_GPU_BUS_TYPE_FPCI;
+                        break;
+                    case NVML_BUS_TYPE_AGP:
+                        *pBusType = NVAPI_GPU_BUS_TYPE_AGP;
+                        break;
+                    default:
+                        *pBusType = NVAPI_GPU_BUS_TYPE_UNDEFINED;
+                        break;
+                }
+
+                return Ok(n);
+            }
+        }
+
         if (adapter->GetArchitectureId() >= NV_GPU_ARCHITECTURE_GM200)
             *pBusType = NVAPI_GPU_BUS_TYPE_PCI_EXPRESS; // Assume PCIe on Maxwell like generation and newer
         else
--- a/src/nvapi_gpu.cpp
+++ b/src/nvapi_gpu.cpp
@@ -520,31 +520,42 @@ extern "C" {
         if (!adapter->HasNvmlDevice())
             return HandleInvalidated(str::format(n, ": NVML available but current adapter is not NVML compatible"), alreadyLoggedHandleInvalidated);
 
+        unsigned sensors;
+        // NVML claims that defaultMinTemp is unsigned int, but it can return values like uint32_t(4294967256)
+        // which would be equal to -40 if interpreted as signed, while NvAPI simply returns NvS32(-40) for the same device
+        signed int defaultMinTemp;
         nvmlGpuThermalSettings_t thermalSettings;
         auto result = adapter->GetNvmlDeviceThermalSettings(sensorIndex, &thermalSettings);
         switch (result) {
             case NVML_SUCCESS:
+                // both NvAPI and NVML fill $(count) sensors when sensorIndex == 15,
+                // 1 sensor when 0 ≤ sensorIndex < $(count) and 0 sensors otherwise
+                sensors = sensorIndex == NVAPI_THERMAL_TARGET_ALL
+                    ? thermalSettings.count
+                    : sensorIndex < thermalSettings.count;
                 switch (pThermalSettings->version) {
                     case NV_GPU_THERMAL_SETTINGS_VER_1: {
                         auto pThermalSettingsV1 = reinterpret_cast<NV_GPU_THERMAL_SETTINGS_V1*>(pThermalSettings);
                         pThermalSettingsV1->count = thermalSettings.count;
-                        for (auto i = 0U; i < thermalSettings.count; i++) {
+                        for (auto i = 0U; i < sensors; i++) {
                             pThermalSettingsV1->sensor[i].controller = MapThermalController(thermalSettings.sensor[i].controller);
                             pThermalSettingsV1->sensor[i].target = MapThermalTarget(thermalSettings.sensor[i].target);
                             pThermalSettingsV1->sensor[i].currentTemp = thermalSettings.sensor[i].currentTemp;
                             pThermalSettingsV1->sensor[i].defaultMaxTemp = thermalSettings.sensor[i].defaultMaxTemp;
-                            pThermalSettingsV1->sensor[i].defaultMinTemp = thermalSettings.sensor[i].defaultMinTemp;
+                            memcpy(&defaultMinTemp, &thermalSettings.sensor[i].defaultMinTemp, sizeof(defaultMinTemp));
+                            pThermalSettingsV1->sensor[i].defaultMinTemp = static_cast<NvU32>(std::max(defaultMinTemp, 0));
                         }
                         break;
                     }
                     case NV_GPU_THERMAL_SETTINGS_VER_2:
                         pThermalSettings->count = thermalSettings.count;
-                        for (auto i = 0U; i < thermalSettings.count; i++) {
+                        for (auto i = 0U; i < sensors; i++) {
                             pThermalSettings->sensor[i].controller = MapThermalController(thermalSettings.sensor[i].controller);
                             pThermalSettings->sensor[i].target = MapThermalTarget(thermalSettings.sensor[i].target);
                             pThermalSettings->sensor[i].currentTemp = static_cast<NvS32>(thermalSettings.sensor[i].currentTemp);
                             pThermalSettings->sensor[i].defaultMaxTemp = static_cast<NvS32>(thermalSettings.sensor[i].defaultMaxTemp);
-                            pThermalSettings->sensor[i].defaultMinTemp = static_cast<NvS32>(thermalSettings.sensor[i].defaultMinTemp);
+                            memcpy(&defaultMinTemp, &thermalSettings.sensor[i].defaultMinTemp, sizeof(defaultMinTemp));
+                            pThermalSettings->sensor[i].defaultMinTemp = defaultMinTemp;
                         }
                         break;
                     default:
--- a/src/nvapi_gpu.cpp
+++ b/src/nvapi_gpu.cpp
@@ -370,8 +370,39 @@ extern "C" {
         if (!adapter->HasNvmlDevice())
             return HandleInvalidated(str::format(n, ": NVML available but current adapter is not NVML compatible"), alreadyLoggedHandleInvalidated);
 
+        nvmlGpuDynamicPstatesInfo_t gpuDynamicPstatesInfo;
+        auto result = adapter->GetNvmlDeviceDynamicPstatesInfo(&gpuDynamicPstatesInfo);
+        switch (result) {
+            case NVML_SUCCESS:
+                // nvmlGpuDynamicPstatesInfo_t also has `flags` but they are reserved for future use
+                pDynamicPstatesInfoEx->flags = 0;
+                // order of the first four utilization domains in NVML matches NVAPI
+                // to consider: should be avoid blindly copying domains 4-7 in case they don't match once introduced?
+                static_assert(NVML_MAX_GPU_UTILIZATIONS == NVAPI_MAX_GPU_UTILIZATIONS);
+                for (auto i = 0U; i < NVAPI_MAX_GPU_UTILIZATIONS; i++) {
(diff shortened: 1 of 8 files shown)


jp7677:
Hi @Saancreed, thanks a lot for already opening this draft PR and really cool that you are working again on the nvml side here and in your repo.
I’ve only looked briefly but indeed, seeing much closer nvml counterparts to nvapi methods removes a lot of guesswork and simplifies things. I’m even in favor of removing the fallback paths to the old implementation. I guess most people using dxvk-nvapi i.c.w. wine-nvml are using latest driver version anyway and if not, nothing substantially breaks.
I’m certainly interested in going forward with this PR? How would you like to proceed?

PS: I haven’t looked, but would be cool if the new headers allows us later to add an implementation to https://github.com/jp7677/dxvk-nvapi/blob/master/src/nvapi_gpu.cpp#L619 since that method is used in UE4, though I have never investigated what UE4 does once this call succeeds.

SveSop:
> PS: I haven’t looked, but would be cool if the new headers allows us later to add an implementation to https://github.com/jp7677/dxvk-nvapi/blob/master/src/nvapi_gpu.cpp#L619 since that method is used in UE4, though I have never investigated what UE4 does once this call succeeds. Edit: Seems that UE4 uses this only to detect overclocking for adding this fact to the logs.

I am not really aware of anything breaking without this function. Nifty to have when benchmarking or whatnot for those windows tools that uses that (Unigine Valley/Heaven comes to mind).

Saancreed:
Oh dear, having tests crash on me with no debug symbols is _not_ a fun experience. But they should be working now. Some more notes:

* `GetDynamicPstatesInfo` and `GetBusType` seem to be as reasonable as I initially expected so they remain unchanged from the initial push.
* I wouldn't drop old implementations _yet_, not only because users might be using older NV driver but also they might be using older `wine-nvml`, like the one shipped by Lutris. We can afford to let them stay for a while, I think.
* This needs a quite recent build of `wine-nvml`. You can build one from the source or get artifacts from some recent Actions run [here](https://github.com/Saancreed/wine-nvml/actions?query=branch%3Amaster+is%3Asuccess) but the latter require glibc 2.35 or newer. Mind the new installation instructions, it seems that copying stuff directly into Wine tree is now required (but maybe it could be solved by `WINEDLLPATH` enviroment variable? not sure about this)

The only functional change is adapting `GetThermalSettings` for two quirks I initially missed:
1. NVML claims `defaultMinTemp` is `unsigned`, which doesn't appear to be the case so I reinterpret it as `signed` to get a meaningful value.
2. NvAPI and NVML both (should) return total number of sensors in `.count` but `sensor` array is filled based on `sensorIndex`:
    * for 15, `count` members are filled (or so I presume because my own laptop has only 1 sensor total)
    * for `sensorIndex < count`, one member is filled
    * for any other value, the array is untouched

Anyway, I think it's ready for review now so I'm marking it as such.

src/nvapi_gpu.cpp
```diff
@@ -372,6 +434,63 @@ extern "C" {
         }
     }
 
+    static NV_THERMAL_TARGET MapThermalTarget(nvmlThermalTarget_t target) {
```
jp7677, line 437: Thanks for start using static_assert. Based on your code I did the following for another place https://github.com/jp7677/dxvk-nvapi/blob/master/src/sysinfo/vulkan.cpp#L67
I suggest to move this mapping to a new method `Nvml::ToNvThermalTarget.`

src/nvapi_gpu.cpp
```diff
@@ -372,6 +434,63 @@ extern "C" {
...
+        static_assert(static_cast<int>(NVML_THERMAL_TARGET_BOARD) == static_cast<int>(NVAPI_THERMAL_TARGET_BOARD));
+        static_assert(static_cast<int>(NVML_THERMAL_TARGET_VCD_BOARD) == static_cast<int>(NVAPI_THERMAL_TARGET_VCD_BOARD));
+        static_assert(static_cast<int>(NVML_THERMAL_TARGET_VCD_INLET) == static_cast<int>(NVAPI_THERMAL_TARGET_VCD_INLET));
+        static_assert(static_cast<int>(NVML_THERMAL_TARGET_VCD_OUTLET) == static_cast<int>(NVAPI_THERMAL_TARGET_VCD_OUTLET));
+        static_assert(static_cast<int>(NVML_THERMAL_TARGET_ALL) == static_cast<int>(NVAPI_THERMAL_TARGET_ALL));
+        static_assert(static_cast<int>(NVML_THERMAL_TARGET_UNKNOWN) == static_cast<int>(NVAPI_THERMAL_TARGET_UNKNOWN));
+        return static_cast<NV_THERMAL_TARGET>(target);
+    }
+
+    static NV_THERMAL_CONTROLLER MapThermalController(nvmlThermalController_t controller) {
```
jp7677, line 451: Same here, lets move this to `Nvml::ToNvThermalController`.

src/nvapi_gpu.cpp
```diff
@@ -115,6 +115,32 @@ extern "C" {
         if (!nvapiAdapterRegistry->IsAdapter(adapter))
             return ExpectedPhysicalGpuHandle(n);
 
+        if (adapter->HasNvmlDevice()) {
+            nvmlBusType_t busType;
+            auto result = adapter->GetNvmlDeviceBusType(&busType);
+            if (result == NVML_SUCCESS) {
+                switch (busType) {
```
jp7677, line 122: I suggest to move the mapping to `Nvml::ToNvGpuBusType`.

src/sysinfo/nvml.cpp
```diff
@@ -14,27 +14,28 @@ namespace dxvk {
...
-        m_nvmlShutdown = GetProcAddress<PFN_nvmlShutdown>("nvmlShutdown");
-        m_nvmlErrorString = GetProcAddress<PFN_nvmlErrorString>("nvmlErrorString");
-        m_nvmlDeviceGetHandleByPciBusId_v2 = GetProcAddress<PFN_nvmlDeviceGetHandleByPciBusId_v2>("nvmlDeviceGetHandleByPciBusId_v2");
-        m_nvmlDeviceGetPciInfo_v3 = GetProcAddress<PFN_nvmlDeviceGetPciInfo_v3>("nvmlDeviceGetPciInfo_v3");
-        m_nvmlDeviceGetTemperature = GetProcAddress<PFN_nvmlDeviceGetTemperature>("nvmlDeviceGetTemperature");
-        m_nvmlDeviceGetUtilizationRates = GetProcAddress<PFN_nvmlDeviceGetUtilizationRates>("nvmlDeviceGetUtilizationRates");
-        m_nvmlDeviceGetVbiosVersion = GetProcAddress<PFN_nvmlDeviceGetVbiosVersion>("nvmlDeviceGetVbiosVersion");
-        m_nvmlDeviceGetPerformanceState = GetProcAddress<PFN_nvmlDeviceGetPerformanceState>("nvmlDeviceGetPerformanceState");
-        m_nvmlDeviceGetClockInfo = GetProcAddress<PFN_nvmlDeviceGetClockInfo>("nvmlDeviceGetClockInfo");
+#define GETPROCADDR(x) m_##x = GetProcAddress<PFN_##x>(#x)
```
jp7677, line 17: I'm torn about this macro, it does indeed remove boilerplate code, but not sure if a macro wrapping a template is a wise thing.

tests/nvapi_d3d.cpp
```diff
@@ -44,7 +44,7 @@ TEST_CASE("D3D methods succeed", "[.d3d]") {
 
         SECTION("GetCurrentSLIState (V2) returns OK") {
             NV_GET_CURRENT_SLI_STATE_V2 state;
-            state.version = NV_GET_CURRENT_SLI_STATE_VER1;
+            state.version = NV_GET_CURRENT_SLI_STATE_VER2;
```
jp7677, line 47: Thanks for finding this!

tests/nvapi_sysinfo.cpp
```diff
@@ -781,9 +781,72 @@ TEST_CASE("Sysinfo methods succeed", "[.sysinfo]") {
             REQUIRE(strcmp(revision, version) == 0);
         }
 
-        SECTION("GetDynamicPstatesInfoEx returns OK") {
+        SECTION("GetBusType returns OK") {
+            ALLOW_CALL(*nvml, DeviceGetBusType(_, _)) // NOLINT(bugprone-use-after-move)
+                .LR_SIDE_EFFECT(*_2 = NVML_BUS_TYPE_FPCI) // something more exotic than default PCIE
```
jp7677, line 786: Using the GENERATE macro from catch we can quite simple test the complete mapping, see e.g. https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_sysinfo.cpp#L573

src/nvapi_gpu.cpp
```diff
@@ -342,8 +371,39 @@ extern "C" {
...
             return HandleInvalidated(str::format(n, ": NVML available but current adapter is not NVML compatible"), alreadyLoggedHandleInvalidated);
 
+        nvmlGpuDynamicPstatesInfo_t gpuDynamicPstatesInfo;
+        auto result = adapter->GetNvmlDeviceDynamicPstatesInfo(&gpuDynamicPstatesInfo);
+        switch (result) {
+            case NVML_SUCCESS:
+                // nvmlGpuDynamicPstatesInfo_t also has `flags` but they are reserved for future use
+                pDynamicPstatesInfoEx->flags = 0;
+                // order of the first four utilization domains in NVML matches NVAPI
+                // to consider: should be avoid blindly copying domains 4-7 in case they don't match once introduced?
```
jp7677, line 381: Good question, I tend to say blindly copying is fine since chances are very high that nvapi will match nvml. But yeah, it's a small gamble.

tests/nvapi_sysinfo.cpp
```diff
@@ -813,8 +876,74 @@ TEST_CASE("Sysinfo methods succeed", "[.sysinfo]") {
...
+
+            SECTION("GetThermalSettings with current struct version returns not incompatible-struct-version") {
+                // This test should fail when a header update provides a newer not yet implemented struct version
+                NV_GPU_THERMAL_SETTINGS settings;
+                settings.version = NV_GPU_THERMAL_SETTINGS_VER;
+                REQUIRE(NvAPI_GPU_GetThermalSettings(handle, NVAPI_THERMAL_TARGET_ALL, &settings) != NVAPI_INCOMPATIBLE_STRUCT_VERSION);
+            }
+        }
+
+        SECTION("GetThermalSettings succeeds when DeviceGetTemperature is available") {
```
jp7677, line 943: Since this is the fallback, naming it like `GetThermalSettings succeeds when DeviceGetThermalSettings is not available` fits better.

jp7677 (changes requested):
I finished the first round, looks really good! 
Just in case you haven't already, could you also run `clang-format` with the rules in the project?

tests/nvapi_sysinfo.cpp
Saancreed: I'd actually prefer to be more verbose here and have something like `GetThermalSettings succeeds when DeviceGetThermalSettings is not available but DeviceGetTemperature is` so we don't claim that it's meant to succeed when local version of NVML supports neither function.

src/sysinfo/nvml.cpp
Saancreed: I think it's justified because it makes you literally unable to query wrong method (by passing wrong name by accident) and it's not hidden away in some other header file so it's easy to verify it really does what you expect it to do.

But I can revert this change if you'd like.

tests/nvapi_sysinfo.cpp
Saancreed: Initially I merely wanted to test that the result of `nvmlDeviceGetBusType` is respected, not that every possible value is mapped correctly, but I see your point so I've added cases for remaining bus types as suggested.

tests/nvapi_sysinfo.cpp
jp7677: Yeah that's even better.

src/sysinfo/nvml.cpp
jp7677: Yeah, I'm seeing the benefit. Lets do it the other way around, could you please apply the same magic to https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_system.cpp#L102 and the `GetInstanceProcAddress` usages in  https://github.com/jp7677/dxvk-nvapi/blob/master/src/sysinfo/vulkan.cpp (also moving them back into the constructor). That way we do similar scoped magic everywhere when obtaining function pointers.

jp7677: As discussed, lets do this in a separate PR.

jp7677 (approved):

PR #90, opened 2022-06-25
Macro magic

Saancreed:
As discussed earlier. Vulkan (and LFX) stuff remains untouched because there's not much to be gained there.

A minor change to `GETPROCADDR` allowed me to also get rid of `template <typename T> T GetProcAddress(…)`.

### DIFF
--- a/src/sysinfo/nvml.cpp
+++ b/src/sysinfo/nvml.cpp
@@ -14,7 +14,7 @@ namespace dxvk {
             return;
         }
 
-#define GETPROCADDR(x) m_##x = GetProcAddress<PFN_##x>(#x)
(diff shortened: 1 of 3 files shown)


jp7677:
Thanks for the follow up! Tbh, I would prefer to keep the template in `Nvml`, otherwise LGTM!

jp7677 (changes requested):
See https://github.com/jp7677/dxvk-nvapi/pull/90#issuecomment-1166308785

Saancreed:
Done :slightly_smiling_face:

jp7677 (approved):

jp7677:
Pushed to master, thanks!

PR #91, opened 2022-06-25
NVML: Implement `NvAPI_GPU_GetGpuCoreCount`

Saancreed:
At least Immortals Fenyx Rising and Assassin's Creed Valhalla call this function, though I have no idea what for. In case of the former, I can't tell any difference (but hey, at least the game still works!) and in case of the latter, I can't check it because Valhalla seems to be utterly broken on NV r515 for unrelated reasons :upside_down_face:


jp7677 (changes requested):
Could you please add a calls to `GetGpuCoreCount` here as well https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_sysinfo.cpp#L1127 for testing the behavior without `nvml` and variations of that?
Otherwise all good!

Saancreed:
Like this?

jp7677 (approved):

jp7677:
Yes! :)

Issue #92, opened 2022-06-25
Setup clang-format check for CI


jp7677:
This might do the trick https://mesonbuild.com/Code-formatting.html

PR #93, opened 2022-06-26
Add linting to CI

jp7677:
~TODO: Need to investigate if the meson version bump works for Proton~

This addresses https://github.com/jp7677/dxvk-nvapi/issues/92


meson.build
```diff
@@ -5,7 +5,7 @@ project('dxvk-nvapi',
     'warning_level=2'
   ],
   version : 'v0.0',
-  meson_version : '>= 0.49.0')
+  meson_version : '>= 0.60.0')
```
jp7677, line 8: Note to self, version 0.58.0 is sufficient.

PR #97, opened 2022-09-10
nvapi-gpu: Remove temperature signedness workaround

Saancreed:
NVML header from release 515.65.01 has the signedness of `int`s in `nvmlGpuThermalSettings_t` fixed, so we can remove our hackish reinterpret cast now.

### DIFF
--- a/inc/nvml.h
+++ b/inc/nvml.h
@@ -625,9 +625,9 @@ typedef struct
     struct
     {
         nvmlThermalController_t controller;
-        unsigned int defaultMinTemp;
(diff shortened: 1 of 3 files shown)


jp7677:
Nice, thanks a lot for the clean up and for keeping an eye on this.
Would you mind to put the header update into a separate commit? Otherwise, LGTM!

Saancreed:
Done. I thought that a separate commit with just header update would be unbuildable but it doesn't appear to be the case.

jp7677:
Thanks!

Issue #47, opened 2021-10-03
Enabling NVAPI breaks Monster Hunter World


Saancreed:
MHW (which is a DLSS 1.1.13 title) crashes shortly after launch with DXVK-NVAPI enabled. Setting `WINEDLLOVERRIDES='nvngx='` allows it to run again.

Proton log file (created with `env DXVK_NVAPI_LOG_LEVEL=info PROTON_LOG=1 PROTON_ENABLE_NVAPI=1 prime-run %command%`): [steam-582010.log](https://github.com/jp7677/dxvk-nvapi/files/7274273/steam-582010.log)

One line of interest is probably this:
```
011c:err:d3d12_device_vkd3d_ext_CreateCubinComputeShaderWithName: Failed to create cubin function module, vr -3.
```

Tested on GTX 1660 Ti Mobile (driver version 470.62.02), using Arch Linux, current Proton Experimental (but with Soldier runtime disabled) and DXVK-NVAPI from ca1d79c4692a13e9669e6a42dc136dd92c1197d3 because I wasn't sure if the issue is related.

jp7677:
@rochaudhari Do you have an idea whats going on here?

liam-middlebrook:
Can you grab a log with https://github.com/jp7677/dxvk-nvapi/pull/49 merged in please?

rochaudhari:
Sorry, I was on vacation and back today.
@Saancreed please provide the log mentioned by Liam.

Saancreed:
@liam-middlebrook @rochaudhari [steam-582010.log](https://github.com/jp7677/dxvk-nvapi/files/7288240/steam-582010.log)

jp7677:
Hi @Saancreed, may I ask you to try something?
The game asks for driver version, GPU architecture and adapter Luid before trying to start DLSS. 

Could you try to override the driver version to something older (can be set with `DXVK_NVAPI_DRIVER_VERSION`)? May be that convinces the game to not initialize DLSS? If not, what happens if you modify the source of dxvk-nvapi to return an architecture older than Turing (https://github.com/jp7677/dxvk-nvapi/blob/master/src/nvapi_gpu.cpp#L145) or by return Error when it requests the Luid (https://github.com/jp7677/dxvk-nvapi/blob/master/src/nvapi_gpu.cpp#L118)?

[...]

PR #98, opened 2022-10-31
Update to r520 NvAPI headers

liam-middlebrook:
Some procedural updates for the new NvAPI release which includes support for NVIDIA Ada GPUs. Additionally this PR adds a workaround for a bug affecting DLSS versions 2.0 through 2.3.


src/nvapi_gpu.cpp
```diff
@@ -3,6 +3,10 @@
 #include "util/util_statuscode.h"
 #include "util/util_string.h"
 
+#if defined(__GNUC__)
+#define _ReturnAddress() __builtin_return_address(0);
```
jp7677, line 7: I think we can move this define to `nvapi_private.h`.

src/nvapi_gpu.cpp
```diff
@@ -310,11 +386,22 @@ extern "C" {
 
         auto architectureId = adapter->GetArchitectureId();
 
+        if (architectureId > NV_GPU_ARCHITECTURE_GA100) {
```
jp7677, line 389: Thinking about this, I would prefer to move those quirks further in the background. What about moving the `is*Needed` method to `env` and change the signature so that it looks like
```c++
if (needsAmpereSpoofing(architectureId, returnAddress))
    architectureId = NV_GPU_ARCHITECTURE_GA100;
```
(The check for Ada should move into that method.  I'm also fine to declare  `returnAddress` at the top of this method, even if it is unused for other architecture)

Once we have this, I can continue adding a quirk for https://github.com/jp7677/dxvk-nvapi/issues/47 (likely `needsPascalSpoofing(architectureId)` since we can just spoof a non DLSS architecture if the executable name matches that game.

src/nvapi_gpu.cpp
```diff
@@ -289,6 +293,78 @@ extern "C" {
         return Ok(n);
     }
 
+    // Function to check for WAR to DLSS Bug 3634851 if an affected DLSS DLL is
+    // detected this function will return true; false will be returned for all
+    // non-affected DLLs.
+    static bool isDLSSBug3634851WARNeeded(void* pReturnAddress) {
```
jp7677, line 299: See below about the suggestion to move to `env`.
The bug ID is unfortunately private, so I would prefer a more descriptive name. That said please leave the ID (and/or a private link to it) in source code comments.

src/nvapi_gpu.cpp
```diff
@@ -310,11 +386,22 @@ extern "C" {
...
+            if (isDLSSBug3634851WARNeeded(returnAddress))
+                architectureId = NV_GPU_ARCHITECTURE_GA100;
+        }
+
         // Assume the implementation ID from the architecture ID. No simple way
         // to do a more fine-grained query at this time. Would need wine-nvml
         // usage.
         NV_GPU_ARCH_IMPLEMENTATION_ID implementationId;
         switch (architectureId) {
+            case NV_GPU_ARCHITECTURE_AD100:
```
jp7677, line 402: Cou;d you try to add a unit test for this case to https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_sysinfo.cpp#L641
(You need to add another test case to `args` and also add  `deviceId` as field to the `args` structure and configure idProps here https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_sysinfo.cpp#L663)

jp7677 (commented):
Hi @liam-middlebrook  thanks for the PR, see the review for some points. As a general note, I would prefer auto when possible and I'm fine to remove `{}` when not needed.

jp7677:
@liam-middlebrook Looking at the updated headers, is it wise to also add `NvAPI_GPU_GetLogicalGpuInfo` in the near future since `NvAPI_GPU_GetAdapterIdFromPhysicalGpu` got deprecated?

src/nvapi_gpu.cpp
liam-middlebrook: Sure

src/nvapi_gpu.cpp
jp7677: I forgot this earlier, this method also needs to know about the new generation: https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_system.cpp#L33

jp7677: Here is another place I forgot about earlier https://github.com/jp7677/dxvk-nvapi/blob/master/src/nvapi_d3d12.cpp#L149 
From what I could find, it should be `sm_89` for Ada.

jp7677:
Superseded by https://github.com/jp7677/dxvk-nvapi/pull/101

PR #101, opened 2022-11-26
Update to r525 NvAPI headers (2)

jp7677:
This PR contains the commits from https://github.com/jp7677/dxvk-nvapi/pull/98 but based on the 525 headers. Thus add support for Ada and add a spoofing work around for certain DLSS versions.

Also move the spoofing check further to the background and extend tests.


jp7677:
TODO:
- Adding a log message when spoofing another architecture is a good idea.
- Since we are on it, add Pascal spoofing for https://github.com/jp7677/dxvk-nvapi/issues/47

PR #109, opened 2023-02-13
nvapi-d3d11: Add D3D11_CreateDevice and D3D11_CreateDeviceAndSwapChain

jp7677:


Saancreed (approved):
I'm assuming we don't need any sophisticate error handling and can always return generic error to the application if our own call fails.

jp7677:
Yeah, i also think we can be lazy and just refer to DXVK logs :)

PR #119, opened 2023-04-12
Add a few OpacityMicromap-related stubs

Saancreed:
Should fix https://github.com/HansKristian-Work/vkd3d-proton/issues/1507 on Nvidia GPUs which is actually caused by dxvk-nvapi not implementing those 3 functions.

I'm not sure if that's the correct place for a helper like `ConvertBuildRaytracingAccelerationStructureInputs` but dumping it into `NvapiD3d12Device` doesn't feel right either.


src/nvapi_d3d12.cpp
```diff
@@ -240,4 +240,146 @@ extern "C" {
...
+            // assume that OMM via VK_EXT_opacity_micromap is not supported, allow only standard flags to be passed
+            .Flags = static_cast<D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS>(nvDesc->flags & 0x3f),
+            .NumDescs = nvDesc->numDescs,
+            .DescsLayout = nvDesc->descsLayout,
+            .InstanceDescs = nvDesc->instanceDescs,
+        };
+
+        if (d3dDesc->Type == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL && d3dDesc->DescsLayout == D3D12_ELEMENTS_LAYOUT_ARRAY) {
+            if (nvDesc->geometryDescStrideInBytes != sizeof(NVAPI_D3D12_RAYTRACING_GEOMETRY_DESC_EX))
+                return NVAPI_INCOMPATIBLE_STRUCT_VERSION;
```
jp7677, line 286: I guess `NVAPI_INVALID_ARGUMENT` fits better. `NVAPI_INCOMPATIBLE_STRUCT_VERSION` should I guess be reserved for struct  version mismatches.
(This also removes one `case` from the two switch statements below)

src/nvapi_d3d12.cpp
```diff
@@ -240,4 +240,146 @@ extern "C" {
...
+
+        if (pDevice == nullptr || pData == nullptr)
+            return InvalidPointer(n);
+
+        switch (type) {
+            case NVAPI_D3D12_RAYTRACING_CAPS_TYPE_THREAD_REORDERING:
+                if (dataSize != sizeof(NVAPI_D3D12_RAYTRACING_THREAD_REORDERING_CAPS))
+                    return InvalidArgument(n);
+
+                // let's hope that NvAPI_D3D12_IsNvShaderExtnOpCodeSupported returning false is enough to discourage games from attempting to use SER
```
jp7677, line 256: Could you please write out SER, since that abbreviation is not that common?

jp7677 (changes requested):
small nits ..

jp7677:
> I'm not sure if that's the correct place for a helper like `ConvertBuildRaytracingAccelerationStructureInputs` but dumping it into `NvapiD3d12Device` doesn't feel right either.

We could introduce a new `util_` for that method, but no strong feelings for that or leave it here like you did.  `NvapiD3d12Device` indeed doesn't feel like the right place.

src/nvapi_d3d12.cpp
Saancreed: If there's only one type of error reported, I could also return a `bool` instead and replace the entire `switch` with a simpler `if`. Should I do that or would you like me to keep `NvAPI_Status` there?

jp7677: Good point, lets go for `bool`, we already do that in other places e.g.  `NvapiD3d11Device`

Saancreed:
Feel free to move that helper around as you see fit, especially if you decide to test it separately from the functions calling it.

Also, nits should be resolved now.

jp7677 (approved):

PR #125, opened 2023-05-19
Add support for Opacity Micromaps

### LINKED ISSUE #123: Support Nvidia new OMM sample..
Hi,
recently VKD3D fixed a bug (https://github.com/HansKristian-Work/vkd3d-proton/issues/1521) for running new OMM SDK sample under DX12 mode..
in DX12 mode demo uses NVAPI D3D12 OMM methods..
now demo fails with "[FAIL]: NvAPI_D3D12_SetCreatePipelineStateOptions"..
searching binary seems can fail also with:
```
[FAIL]: NvAPI_D3D12_SetCreatePipelineStateOptions
[FAIL]: NvAPI_D3D12_GetRaytracingOpacityMicromapArrayPrebuildInfo
[FAIL]: NvAPI_D3D12_GetRaytracingAccelerationStructurePrebuildInfoEx
[FAIL]: NvAPI_D3D12_BuildRaytracingOpacityMicromapArray
[FAIL]: NvAPI_D3D12_BuildRaytracingAccelerationStructureEx
```
so ideally some of this calls should map to use VK_EXT_opacity_micromap either in VKD3D or DXVK-nvapi but should be enough if some stubs are added avoiding failing these calls even if render is not correct..

Saancreed:
A rough proof-of-concept that gets at least _some_ parts of OMMs working.

image

Needs to be combined with https://github.com/Saancreed/vkd3d-proton/tree/nvapi-omm and therefore will remain a draft for the foreseeable future. Used interfaces should not be considered stable.

Still, I'm posting this for some visibility, maybe one day we will actually find any game that makes use of this that doesn't just call `SetCreatePipelineStateOptions(NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS_ENABLE_OMM_SUPPORT)` only to never actually build any micromap like Cyberpunk 2077 v1.62 seems to do.

Closes #123


adamdmoss (approved):
Looks good to me, FWIW!  I guess I'd want to actually try it out to be really confident though. :laughing:

adamdmoss:
> maybe one day we will actually find any game that makes use of this that doesn't just call SetCreatePipelineStateOptions(NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS_ENABLE_OMM_SUPPORT) only to never actually build any micromap like Cyberpunk 2077 v1.62

I'm not exploiting privileged information but the OMM usage situation may be better with more recent Cyberpunk 2077 versions.

Saancreed:
> I'm not exploiting privileged information but the OMM usage situation may be better with more recent Cyberpunk 2077 versions.

Yup, I've noticed Cyberpunk 2077 actually using OMMs since around Phantom Liberty update. I have verified that this code does work with the game (as in, it successfully uses micromaps in its pipeline and does not explode), or at least it did a few months ago, but unfortunately I was unable to notice any performance gain in its built-in benchmark on my GeForce RTX 4080 Mobile.

Saancreed:
Superseded by #371.

PR #133, opened 2023-07-08
build: Switch symbols to DWARF 4

jp7677:
This mirrors https://github.com/doitsujin/dxvk/commit/2ef41bdbf6841c2887e7af4b7c3b26235d3e7efa and also makes the meson setup again more like DXVK.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,15 +7,25 @@ project('dxvk-nvapi',
   version : 'v0.0',
   meson_version : '>= 0.58.0')
 
-dxvk_compiler = meson.get_compiler('cpp')
-add_project_arguments(dxvk_compiler.get_supported_arguments('-Wno-unused-parameter'), language : 'cpp')
+cpp = meson.get_compiler('cpp')
+
+compiler_args = ['-Wno-unused-parameter']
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -7,15 +7,26 @@ project('dxvk-nvapi',
...
-  add_project_arguments(dxvk_compiler.get_supported_arguments('-Wa,-mbig-obj'), language : 'cpp')
+  compiler_args += [
+    '-gstrict-dwarf',
+    '-gdwarf-4',
+    '-Wa,-mbig-obj'
+  ]
 endif
 
-add_project_link_arguments(dxvk_compiler.get_supported_link_arguments('-Wl,--file-alignment=4096'), language: 'cpp')
+link_args = [
```
jp7677, line 21: Those are the same arguments DXVK uses.

Saancreed (approved):
Since https://github.com/doitsujin/dxvk/commit/2ef41bdbf6841c2887e7af4b7c3b26235d3e7efa DXVK also no longer uses `-gstrict-dwarf`, maybe we could remove it as well?

PR #134, opened 2023-07-11
Add resource details to fine .dll file

SveSop:
There is not really support for adding other than numbers for the version string, so using VCS_TAG seems out of the question. Used the "release" tag, but this will for now need manual update for each release.

GPU Caps Viewer still shows "unknown" for the file version of the .dll, so this is mostly just for some cosmetic "information" - similar to what DXVK does for d3d11.dll and such. "For completeness" so to speak?
nvapi_dll

### DIFF
--- a/build-win32.txt
+++ b/build-win32.txt
@@ -2,6 +2,7 @@
 cpp = 'i686-w64-mingw32-g++'
 strip = 'i686-w64-mingw32-strip'
 ar = 'i686-w64-mingw32-ar'
+windres = 'i686-w64-mingw32-windres'
 
 [built-in options]
 cpp_args=['-msse', '-msse2']
--- a/build-win64.txt
+++ b/build-win64.txt
@@ -2,6 +2,7 @@
 cpp = 'x86_64-w64-mingw32-g++'
 strip = 'x86_64-w64-mingw32-strip'
 ar = 'x86_64-w64-mingw32-ar'
+windres = 'x86_64-w64-mingw32-windres'
 
 [built-in options]
 cpp_link_args = ['-static', '-static-libgcc', '-static-libstdc++']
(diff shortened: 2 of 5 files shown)


jp7677:
Thanks for the PR. I'm actually quite torn on this one. While I don't mind the extra bits for having proper file information, maintaining versions numbers manually sounds very tedious. That said, somehow extracting the version from tags is also risky since e.g. Proton uses a very different tag format.

(on a side note, the filename should differ between 32bit and 64bit builds)

Saancreed:
> (on a side note, the filename should differ between 32bit and 64bit builds)

Actually, on native Windows it does not:

```
$ exiftool -OriginalFileName /windows/Windows/System32/nvapi64.dll 
Original File Name              : nvapi.dll
```

jp7677:
> > (on a side note, the filename should differ between 32bit and 64bit builds)
> 
> Actually, on native Windows it does not:
> 
> ```
> $ exiftool -OriginalFileName /windows/Windows/System32/nvapi64.dll 
> Original File Name              : nvapi.dll
> ```

Thanks, well, things are never as easy as they should be :)

SveSop:
The initial thought of doing this was because i thought that GPU Caps Viewer that now shows "unknown version" for the dll would pick this up somehow, but it does not seem as it is satisfied.. If that is because of wine not really have any functions for this or whatnot i dunno, or if there is some missing setting in the "manifest" or whatever, i dunno. I can fiddle a bit with it, but it is mostly just for cosmetics i guess. 

Ill keep it around on my experimental branch, and i tend to agree that it would be a bit more ideal if the version actually was some "dirty git +" tag when building, but the limitations is "4 numbers with . between them", so no letters and such there, and well.. needing manual update.

SveSop:
Not that i have done extensive testing for this, but afaik i do not really know of any apps or anything that uses the .dll information to pull version information yet, and the upkeep on updating this every time new nvapi or driverversion trickles along is too tedious until we can show a real purpose.

So.. Closing this PR for now. Can always fiddle with it if something happens that needs this for anything.

PR #137, opened 2023-08-06
Use __wine_dbg_output for console output

jp7677:
This mirrors what DXVK and VKD3D-Proton does for printing to console (https://github.com/doitsujin/dxvk/commit/05fb634f9102f18e6a2c9e84939d0fb0a3618671) and also fixes lost console output on Wine 8.12.

### DIFF
--- a/src/util/util_log.cpp
+++ b/src/util/util_log.cpp
@@ -1,8 +1,27 @@
 #include "util_log.h"
 #include "util_env.h"
+#include "util_string.h"
+
+using PFN_wineDbgOutput = int(__cdecl*)(const char*);
+
+static PFN_wineDbgOutput wineDbgOutput = nullptr;
 
 namespace dxvk::log {
+    void print(const std::string& message) {
+        auto line = message + '\n';
+        if (wineDbgOutput)
+            wineDbgOutput(line.c_str());
+        else
+            std::cerr << line;
+    }
+
     void initialize(std::ofstream& filestream, bool& skipAllLogging) {
+#ifdef _WIN32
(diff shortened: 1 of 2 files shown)


SveSop:
Seems to be working fine with this change - console logging is back 👍

src/util/util_log.cpp
```diff
@@ -1,8 +1,26 @@
 #include "util_log.h"
 #include "util_env.h"
+#include "util_string.h"
+
+using PFN_wineDbgOutput = int(__cdecl*)(const char*);
+
+static PFN_wineDbgOutput wineDbgOutput = nullptr;
 
 namespace dxvk::log {
+    void cout(const std::string& message) {
```
Saancreed, line 10: I think it's kinda weird to name this after console/standard out when in both cases the message ends in console/standard error (or so I believe at least).

src/util/util_log.cpp
```diff
@@ -1,8 +1,26 @@
...
+        if (wineDbgOutput)
+            wineDbgOutput((message + '\n').c_str());
```
Saancreed, line 12: I don't know C++ lifetime rules well enough so I'm not sure if that's necessary but I'd assign `message + '\n'` to a variable just to make sure it doesn't stop existing once we call `c_str()` on it and lose it from the scope.

src/util/util_log.cpp
jp7677: Yeah, I've struggled with the name.. renamed to `print` now..

Saancreed (approved):

PR #139, opened 2023-08-15
Implement D3D12 CuBIN Dynamic Shared Memory Support

liam-middlebrook:
dxvk-nvapi follow-up for https://github.com/HansKristian-Work/vkd3d-proton/pull/1640

Implements `NvAPI_D3D12_CreateCubinComputeShaderEx` allowing for dynamic shared memory to be used in CuBINs. Some refactoring of the `NvapiD3D12Device` implementation and its handling of CuBINs and `ID3D12GraphicsCommandListExt`s is done to facilitate this.


liam-middlebrook:
ACK on the test failures. Will resolve and add new tests for this. I had misunderstood how things were laid out, but I think I get it now.

liam-middlebrook:
Let me know if I should squash the `ID3D12GraphicsCommandListExt1` implementation + test fixes into a single commit. I kept them separate for reviewability, but I know that some prefer to have every single commit pass all tests.

src/d3d12/nvapi_d3d12_device.cpp
```diff
@@ -105,17 +107,36 @@ namespace dxvk {
         return deviceExt;
     }
 
-    Com<ID3D12GraphicsCommandListExt> NvapiD3d12Device::GetCommandListExt(ID3D12GraphicsCommandList* commandList) {
+    d3d12_graphics_command_list_ext_iface* NvapiD3d12Device::GetCommandListExt(ID3D12GraphicsCommandList* commandList, uint32_t* interfaceVersion) {
```
Saancreed, line 151: I'm a bit conflicted about what would be less awkward here, this or returning `std::optional<std::pair<d3d12_graphics_command_list_ext_iface, uint32_t>>` containing both the interface and its version. I'll leave this one to Jens.

src/d3d12/nvapi_d3d12_device.cpp
```diff
@@ -59,11 +86,27 @@ namespace dxvk {
...
+            auto it = m_cubinSmemMap.find(pShader);
+            if (it != m_cubinSmemMap.end())
+                smem = it->second;
+            else
+                log::write("Failed to find CuBIN in m_cubinSmemMap, defaulting to 0");
+        }
+
+        if (interfaceVersion >= 1)
+            return SUCCEEDED(commandListExt->LaunchCubinShaderEx(reinterpret_cast<D3D12_CUBIN_DATA_HANDLE*>(pShader), blockX, blockY, blockZ, smem, params, paramSize, nullptr, 0));
+        else
```
Saancreed, line 108: I think finding a non-zero smem size in the map when interface version is `< 1` also warrants for a log message warning about this.

src/d3d12/nvapi_d3d12_device.cpp
```diff
@@ -105,17 +148,45 @@ namespace dxvk {
...
+                iface = (d3d12_graphics_command_list_ext_iface*)commandListExt;
+                break;
+            }
+
+            log::write("commandList has no ID3D12GraphicsCommandListExt compatible interface!");
             return nullptr;
+        } while (0);
+
+        // Release the reference that was taken by our successful QueryInterface() call
+        iface->Release();
```
Saancreed, line 186: I am by no means a COM expert but this looks like it would drop the refcount back to 0 and free the object before we have a chance to use it?

src/d3d12/nvapi_d3d12_device.cpp
```diff
@@ -105,17 +148,45 @@ namespace dxvk {
...
+        }
+
+        d3d12_graphics_command_list_ext_iface* iface = nullptr;
+
+        // Grab the highest interface possible
+        do {
+            ID3D12GraphicsCommandListExt1* commandListExt1 = nullptr;
+            commandList->QueryInterface(IID_PPV_ARGS(&commandListExt1));
+
+            if (commandListExt1) {
```
Saancreed, line 166: By convention:

```suggestion
            if (SUCCEEDED(commandList->QueryInterface(IID_PPV_ARGS(&commandListExt1)))) {
```

Same below.

Saancreed (commented):
I have some nitpicks and suggestions myself but I'm going to defer the final judgement to Jens here.

tests/nvapi_d3d12.cpp
```diff
@@ -382,6 +392,33 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
...
+        REQUIRE_CALL(commandList, LaunchCubinShaderEx(shaderHandle, blockX, blockY, blockZ, smemSize, params, paramSize, rawParam, rawParamCount))
+            .RETURN(S_OK)
+            .TIMES(1);
+
+        REQUIRE(NvAPI_D3D12_CreateCubinComputeShaderEx(static_cast<ID3D12Device*>(&device), cubinData, cubinSize, blockX, blockY, blockZ, smemSize, shaderName, reinterpret_cast<NVDX_ObjectHandle*>(handle)) == NVAPI_OK);
+        REQUIRE(NvAPI_D3D12_LaunchCubinShader(static_cast<ID3D12GraphicsCommandList*>(&commandList), reinterpret_cast<NVDX_ObjectHandle>(shaderHandle), blockX, blockY, blockZ, params, paramSize) == NVAPI_OK);
+        REQUIRE(deviceRefCount == 0);
+        REQUIRE(commandListRefCount == 0);
+    }
+
```
jp7677, line 421: Thanks a lot for the test, I suggest to add another test that validates the behavior when `ID3D12GraphicsCommandListExt1` is not supported, something ilke:

```c++
    SECTION("LaunchCubinShader without ID3D12GraphicsCommandListExt1 returns OK") {
        // Use and configure new commandList to avoid cache hits
        D3D12Vkd3dGraphicsCommandListMock commandList2;
        ALLOW_CALL(commandList2, QueryInterface(__uuidof(ID3D12GraphicsCommandList1), _))
            .LR_SIDE_EFFECT(*_2 = static_cast<ID3D12GraphicsCommandList1*>(&commandList2))
            .LR_SIDE_EFFECT(commandListRefCount++)
            .RETURN(S_OK);
        ALLOW_CALL(commandList2, QueryInterface(ID3D12GraphicsCommandListExt::guid, _))
            .LR_SIDE_EFFECT(*_2 = static_cast<ID3D12GraphicsCommandListExt*>(&commandList2))
            .LR_SIDE_EFFECT(commandListRefCount++)
            .RETURN(S_OK);
        ALLOW_CALL(commandList2, QueryInterface(ID3D12GraphicsCommandListExt1::guid, _))
            .RETURN(E_NOINTERFACE);
        ALLOW_CALL(commandList2, AddRef())
            .LR_SIDE_EFFECT(commandListRefCount++)
            .RETURN(commandListRefCount);
        ALLOW_CALL(commandList2, Release())
            .LR_SIDE_EFFECT(commandListRefCount--)
            .RETURN(commandListRefCount);

        NVDX_ObjectHandle pShader = nullptr;
        auto blockX = 1U;
        auto blockY = 2U;
        auto blockZ = 3U;
        const void* params = nullptr;
        auto paramSize = 4U;
        REQUIRE_CALL(commandList2, LaunchCubinShader(reinterpret_cast<D3D12_CUBIN_DATA_HANDLE*>(pShader), blockX, blockY, blockZ, params, paramSize))
            .RETURN(S_OK)
            .TIMES(1);

        REQUIRE(NvAPI_D3D12_LaunchCubinShader(static_cast<ID3D12GraphicsCommandList*>(&commandList2), pShader, blockX, blockY, blockZ, params, paramSize) == NVAPI_OK);
        REQUIRE(deviceRefCount == 0);
        REQUIRE(commandListRefCount == 0);
    }

```

jp7677 (changes requested):

jp7677:
Hi @liam-middlebrook thanks for the PR. See the suggestions for (hopefully) improving readability/testability and the question about the reference count.
No need to squash implementation and tests together.

src/d3d12/nvapi_d3d12_device.cpp
```diff
@@ -31,14 +32,40 @@ namespace dxvk {
...
+            return false;
+
+        bool status = SUCCEEDED(deviceExt->CreateCubinComputeShaderWithName(cubinData, cubinSize, blockX, blockY, blockZ, shaderName, reinterpret_cast<D3D12_CUBIN_DATA_HANDLE**>(pShader)));
+
+        if (status == true) {
+            std::scoped_lock lock(m_CubinSmemMutex);
+            m_cubinSmemMap.emplace(*pShader, smemSize);
+        }
+
+        return status;
```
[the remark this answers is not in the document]
liam-middlebrook: That raises the scope that the `std::scoped_lock` exists in, which means that we'd be holding the lock when calling into vkd3d-proton. This means that if multiple threads call `NvAPI_D3D12_CreateCubinShader*()` at the same time, they'll execute serially rather than in parallel.

src/d3d12/nvapi_d3d12_device.h
```diff
@@ -6,6 +6,7 @@
 
 namespace dxvk {
 
+    typedef ID3D12GraphicsCommandListExt1 d3d12_graphics_command_list_ext_iface;
```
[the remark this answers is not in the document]
liam-middlebrook: I had done this to reduce the number of lines we'd need to change if another interface version was written in the future. I'll remove it from my changes since it's not really necessary.

tests/nvapi_d3d12.cpp
liam-middlebrook: That's a good point, thanks!. I'll add that in with my next batch of updates.

Once I started to grok the mantra behind these tests they made more sense to me :)

src/d3d12/nvapi_d3d12_device.cpp
liam-middlebrook: Agreed, will add in next update.

src/d3d12/nvapi_d3d12_device.cpp
liam-middlebrook: I varied from your suggestion a bit and decided to separate out building the `NvapiD3d12Device::CommandListExtWithVersion` from the emplace, to improve debuggability.

src/d3d12/nvapi_d3d12_device.cpp
liam-middlebrook: From my local experiments with the test app (and a real app) that's not the case.

Consider the code before any changes. When `LaunchCubinShader` is called we first try and lookup the `commandList` that was passed in to see if we've seen it before. If we haven't we use `QueryInterface` to populate  `Com<ID3D12GraphicsCommandListExr>` object with it's type.

At this point, the object's refcount raises from 0 to 1.

Next we extract the raw pointer from the `Com<ID3D12GraphicsCommandListExt>` object. `Com<>::ptr()` does *not* alter the refcount. This extract is done so that we can add that extracted pointer to `m_CommandListMap`.

After this we return from the function. Because `Com<ID3D12GraphicsCommandListExt>` was stack declared, it "pops" and its destructor is called. The destructor decrements the reference count for the object from 1 to 0.

Subsequent calls to `GetCommandListExt()` simply return the raw pointer value and ignore refcounting completely.

This is kind of necessary, because if we held onto a reference the whole time we'd hit issues where an application couldn't properly release it's command lists to free them because we held a reference to them.

src/d3d12/nvapi_d3d12_device.cpp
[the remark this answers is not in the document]
Saancreed: Doesn't the `scoped_lock` in above example protect only the statements after its construction until the function returns? The interface call should be outside its scope and hence executed without the lock being held because it's constructed only after the interface calls returns so the only code protected by it is the `emplace` call and the evaluation of returned `true` constant.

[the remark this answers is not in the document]
Saancreed: A sample to illustrate the point:

```cpp
#include <format>
#include <iostream>
#include <mutex>
#include <pthread.h>
#include <thread>

void proc()
{
    static std::mutex m;
    auto tid = pthread_self();

    std::cout << std::format("Hello from thread {}!\n", tid); // not protected, this could be the interface call

    // { not needed
    std::scoped_lock<std::mutex> lock{m};

    std::this_thread::sleep_for(std::chrono::seconds(2)); // protected, this could be the emplace
    std::cout << std::format("Goodbye from thread {}!\n", tid);
    // } not needed
}

int main()
{
    std::thread a{proc};
    std::thread b{proc};
    std::thread c{proc};
    std::thread d{proc};

    a.join();
    b.join();
    c.join();
    d.join();
}
```

src/d3d12/nvapi_d3d12_device.cpp
jp7677: ah yeah, looking again at the old code (and at other cache maps), we always neutralize the ref counting from our side when putting things in the cache maps. Strict;y speaking in terms of COM, we shouldn't do that and release when we are sure that those pointers aren't used anymore (thus by e.g. storing `Com<>` in the maps instead of the raw pointers). But that would mean that we keep all those instances alive until someone calls `NvAPI_Uninitialize`. Being afraid that we keep things unnecessary alive, like Liam also stated, we went in the past for the currently implemented approach.
Currently we thus rely on good behavior of consumers and (at least to my knowledge), it hasn't exploded.

tests/nvapi_d3d12.cpp
jp7677: Cool ;)
(also for my todo list: extract the setup of the mocks at some point in the future to never hit cache issues)

src/d3d12/nvapi_d3d12_device.cpp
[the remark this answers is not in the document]
liam-middlebrook: C vs C++ semantics, whoops :) Fixed in latest push

jp7677 (approved):

jp7677:
Thanks a lot for the PR and the short response times!

PR #140, opened 2023-08-17
nvapi-disp: Implement HDR support

Joshua-Ashton:
Implements global HDR support using the new DXVK global HDR interfaces.


Joshua-Ashton:
Draft while DXVK side gets merged.

src/dxvk/dxvk_interfaces.h
```diff
@@ -175,7 +175,19 @@ ID3D11VkExtContext1 : public ID3D11VkExtContext {
         uint32_t numWriteResources) = 0;
 };
 
+MIDL_INTERFACE("2a289dbd-2d0a-4a51-89f7-f2adce465cd6")
+IDXGIVkInteropFactory1 : public IDXGIVkInteropFactory {
```
Saancreed, line 179: Looks like this is missing `static const GUID guid;` field?

src/nvapi_disp.cpp
```diff
@@ -19,18 +19,11 @@ extern "C" {
...
-        // Doing so would only make sense when also fully implementing NvAPI_Disp_HdrColorControl to set HDR properties,
-        // but implementing that function correctly is not possible since DXVK/DX sets HDR properties on a
-        // SwapChain and not on a display/output.
-        // Reference (a.o.): https://www.asawicki.info/news_1703_programming_hdr_monitor_support_in_direct3d
         switch (pHdrCapabilities->version) {
             case NV_HDR_CAPABILITIES_VER1: {
                 auto pHdrCapabilitiesV1 = reinterpret_cast<NV_HDR_CAPABILITIES_V1*>(pHdrCapabilities);
                 *pHdrCapabilitiesV1 = {};
-                pHdrCapabilitiesV1->isST2084EotfSupported = false; // output->GetColorData().HasST2084Support;
+                pHdrCapabilitiesV1->isST2084EotfSupported = output->GetColorData().HasST2084Support;
```
Saancreed, line 26: It might be a good idea to extract `output->GetColorData()` to a variable because it's heavily used throughout this entire function.

src/nvapi_disp.cpp
```diff
@@ -96,15 +89,85 @@ extern "C" {
...
+                    pHDRColorDataV1->mastering_display_data.displayPrimary_x2 = metadata.BluePrimary[0];
+                    pHDRColorDataV1->mastering_display_data.displayPrimary_y2 = metadata.BluePrimary[1];
+                    pHDRColorDataV1->mastering_display_data.displayWhitePoint_x = metadata.WhitePoint[0];
+                    pHDRColorDataV1->mastering_display_data.displayWhitePoint_y = metadata.WhitePoint[1];
+                    pHDRColorDataV1->mastering_display_data.max_display_mastering_luminance = metadata.MaxMasteringLuminance;
+                    pHDRColorDataV1->mastering_display_data.min_display_mastering_luminance = metadata.MinMasteringLuminance;
+                    pHDRColorDataV1->mastering_display_data.max_content_light_level = metadata.MaxContentLightLevel;
+                    pHDRColorDataV1->mastering_display_data.max_frame_average_light_level = metadata.MaxFrameAverageLightLevel;
+                } else {
+                    pHDRColorDataV1->hdrMode = output->GetColorData().HasST2084Support ? NV_HDR_MODE_UHDA : NV_HDR_MODE_OFF;
```
Saancreed, line 118: Also here, variable for `output->GetColorData()`.

src/nvapi_disp.cpp
```diff
@@ -96,15 +89,85 @@ extern "C" {
...
+                switch (output->GetColorData().BitsPerColor) {
+                    case 6:   pHDRColorDataV2->hdrBpc = NV_BPC_6; break;
+                    case 8:   pHDRColorDataV2->hdrBpc = NV_BPC_8; break;
+                    default:
+                    case 10:  pHDRColorDataV2->hdrBpc = NV_BPC_10; break;
+                    case 12:  pHDRColorDataV2->hdrBpc = NV_BPC_12; break;
+                    case 16:  pHDRColorDataV2->hdrBpc = NV_BPC_16; break;
+                }
+            } else {
+                // Nothing to set. It's all random useless garbage that you would NEVER trust an app to set.
```
Saancreed, line 166: :sweat_smile:

src/sysinfo/nvapi_adapter_registry.h
```diff
@@ -30,8 +31,12 @@ namespace dxvk {
         [[nodiscard]] NvapiOutput* FindPrimaryOutput() const;
         [[nodiscard]] bool IsOutput(NvapiOutput* handle) const;
 
+        IDXGIVkInteropFactory1 *GetInteropFactory() const { return m_dxgiVkInterop.ptr(); }
```
Saancreed, line 34: `[[nodiscard]]` probably to match other methods.

src/sysinfo/nvapi_adapter_registry.h
```diff
@@ -30,8 +31,12 @@ namespace dxvk {
...
+
       private:
         ResourceFactory& m_resourceFactory;
+        Com<IDXGIFactory1> m_dxgiFactory;
+        Com<IDXGIVkInteropFactory1> m_dxgiVkInterop;
```
Saancreed, line 39: IIRC there were previously some concerns about keeping DXGI factories and/or Vulkan instances alive for the entire lifetime of NVAPI but I don't see an easy way around this now.

@jp7677 I hope this is fine.

src/nvapi_disp.cpp
```diff
@@ -96,15 +89,85 @@ extern "C" {
...
+                    .GreenPrimary = { pHDRColorDataV1->mastering_display_data.displayPrimary_x1, pHDRColorDataV1->mastering_display_data.displayPrimary_y1 },
+                    .BluePrimary = { pHDRColorDataV1->mastering_display_data.displayPrimary_x2, pHDRColorDataV1->mastering_display_data.displayPrimary_y2 },
+                    .WhitePoint = { pHDRColorDataV1->mastering_display_data.displayWhitePoint_x, pHDRColorDataV1->mastering_display_data.displayWhitePoint_y },
+                    .MaxMasteringLuminance = pHDRColorDataV1->mastering_display_data.max_display_mastering_luminance,
+                    .MinMasteringLuminance = pHDRColorDataV1->mastering_display_data.min_display_mastering_luminance,
+                    .MaxContentLightLevel = pHDRColorDataV1->mastering_display_data.max_content_light_level,
+                    .MaxFrameAverageLightLevel = pHDRColorDataV1->mastering_display_data.max_frame_average_light_level,
+                };
+                if (FAILED(interop->SetGlobalHDRState(colorspace, &metadata)))
```
Saancreed, line 145: How global should this state be, exactly? Since this NVAPI function operates only on one display, shouldn't this interface extend something like `IDXGIOutput` instead? Or does setting HDR state on one display makes it retrievable from all displays even on Windows with native NVAPI?

src/dxvk/dxvk_interfaces.h
```diff
@@ -175,7 +175,19 @@ ID3D11VkExtContext1 : public ID3D11VkExtContext {
...
+          DXGI_COLOR_SPACE_TYPE   *pOutColorSpace,
+          DXGI_HDR_METADATA_HDR10 *ppOutMetadata) = 0;
+
+  virtual HRESULT STDMETHODCALLTYPE SetGlobalHDRState(
+          DXGI_COLOR_SPACE_TYPE    ColorSpace,
+    const DXGI_HDR_METADATA_HDR10 *pMetadata) = 0;
+};
+
 DXVK_DEFINE_GUID(IDXGIVkInteropFactory)
+DXVK_DEFINE_GUID(IDXGIVkInteropFactory1)
```
Saancreed, line 190: You should probably also initialize it in `dxvk_interfaces.cpp`.

Saancreed (commented):
Needs a `clang-format` pass and probably some changes to tests.

src/nvapi_disp.cpp
Joshua-Ashton: Swapchains are not necessarily linked to outputs.

We treat it as a global override for all of the app's swapchains to be treated as HDR, etc.

This is obviously different to the behaviour on Windows where it forces a single display into a weird forced HDR mode, but it ends up being better overall.

Saancreed: That makes sense. Let's just hope that we won't confuse any app by implementing it this way.

src/nvapi_disp.cpp
```diff
@@ -96,15 +89,85 @@ extern "C" {
...
+        } else {
+            return IncompatibleStructVersion(n);
         }
 
-        return NoImplementation(str::format(n, " (0x", std::hex, displayId, ")")); // See comment in NvAPI_Disp_GetHdrCapabilities
+        if (pHdrColorData->version == NV_HDR_COLOR_DATA_VER2) {
+            auto pHDRColorDataV2 = reinterpret_cast<NV_HDR_COLOR_DATA_V2*>(pHdrColorData);
+            if (pHDRColorDataV2->cmd == NV_HDR_CMD_GET) {
+                pHDRColorDataV2->hdrColorFormat = NV_COLOR_FORMAT_RGB;
+                pHDRColorDataV2->hdrDynamicRange = NV_DYNAMIC_RANGE_VESA; // No idea what this is meant to mean.
```
EndlesslyFlowering, line 156: - `NV_DYNAMIC_RANGE_VESA` is RGB full range
- `NV_DYNAMIC_RANGE_CEA` is RGB/YCbCr limited range
- `NV_DYNAMIC_RANGE_AUTO` automatically chooses something (probably reads some EDID/CTA-861 entries)

<https://github.com/Maassoft/ColorControl/blob/2c5b08af721362d28433e35397b0401a894a5d18/ColorControl/Services/NVIDIA/NvService.cs#L527>

> [Dynamic range: VESA (Full RGB) or CEA (Limited RGB/YUV)](https://github.com/Maassoft/ColorControl)

EndlesslyFlowering: VESA probably means that it is "IT" content which means it doesn't use timings from the CTA-861 block but timings from the normal EDID.
While CEA means that it is "CE" content which uses CTA definied timings.

EDID is more associated with RGB full because it is for PC monitors while CTA/CEA is for TVs which are more likely to expect limited range.

No idea what Nvidia was thinking with naming it like this...

src/sysinfo/nvapi_adapter_registry.h
jp7677: Yes, keeping the dxgi factory alive is imho also the way to go. Creating a new dxgi factory on every hdr-color-control call would be an alternative, but certainly not a good one since DXVK does some initialization work in the dxgi factory constructor.

Once this PR has landed, we can probably revisit/refactor the earlier life time changes to make things simpler again.

src/nvapi_disp.cpp
```diff
@@ -96,15 +89,85 @@ extern "C" {
...
+            return NoImplementation(str::format(n, " (0x", std::hex, displayId, ")"));
+
+        // V2 is based on V1.
+        if (pHdrColorData->version == NV_HDR_COLOR_DATA_VER1 || pHdrColorData->version == NV_HDR_COLOR_DATA_VER2) {
+            auto pHDRColorDataV1 = reinterpret_cast<NV_HDR_COLOR_DATA_V1*>(pHdrColorData);
+            if (pHDRColorDataV1->cmd == NV_HDR_CMD_GET) {
+                DXGI_COLOR_SPACE_TYPE colorspace;
+                DXGI_HDR_METADATA_HDR10 metadata;
+                if (interop->GetGlobalHDRState(&colorspace, &metadata) == S_OK) {
+                    pHDRColorDataV1->hdrMode = colorspace != DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P709 ? NV_HDR_MODE_UHDA : NV_HDR_MODE_OFF;
```
Kaldaien, line 103: This looks wrong. I've modified dozens of games that use NvAPI for HDR, not a single one has ever used `NV_HDR_MODE_UHDA`. They all use `NV_HDR_MODE_UHDA_PASSTHROUGH`.

`NV_HDR_MODE_UHDA` would be scRGB if even a single game ever used it. `NV_HDR_MODE_UHDA_PASSTHROUGH` is HDR10. Despite NVIDIA's docs saying it's "experimental", it is in fact the only format any shipping game uses aside from Mass Effect: Andromeda (uses `NV_HDR_MODE_DOLBY_VISION` as an alternative).

Joshua-Ashton: This is the "Get" mode, but yeah. My thinking here was that if the global HDR toggle in Windows is on, it'd be scRGB here. But maybe it should be UHDA_PASSTHRU. I will check native behaviour.

src/sysinfo/nvapi_adapter_registry.h
Joshua-Ashton: dxvk nvapi creates the factory so that wouldn't work, we don't use a factory passed in from the app.

src/sysinfo/nvapi_output.h
```diff
@@ -14,6 +14,9 @@ namespace dxvk {
 
         struct ColorData {
             bool HasST2084Support;
+
+            uint32_t BitsPerColor;
```
jp7677, line 18: I would prefer to declare `BitPerColor` here as `NV_BPC` and move the conversion from `NvAPI_Disp_HdrColorControl()` to `NvapiOutput::Initialize()`.

src/nvapi_gpu.cpp
```diff
@@ -25,7 +25,7 @@ extern "C" {
             return Ok(n);
         }
 
-        if (*pDisplayIdCount != count) {
+        if (*pDisplayIdCount < count) {
```
jp7677, line 28: Thanks for finding this!

src/sysinfo/nvapi_adapter_registry.cpp
```diff
@@ -40,6 +40,11 @@ namespace dxvk {
                 delete nvapiAdapter;
         }
 
+        m_dxgiFactory = dxgiFactory;
```
jp7677, line 43: Might as well replace `dxgiFactory` with `m_dxgiFactory` everywhere here. When `Initialize` fails (`return false`), the current instance of `NvapiAdapterRegistry` will be deleted anyway (happens in `NvAPI_Initialize`).

jp7677 (changes requested):
Thanks a lot for the PR, very cool and this fits nicely. Some nits from my side and see also the other comments.  Could you also please use clang-format?

With regards to the tests, they need indeed some adjustments to succeed. It's not rocket science but the setup and the mocking framework are not that straight forward. If your time is limited, I'm happy to take over for the tests once the runtime implementation is done.

Joshua-Ashton:
I would definitely appreciate if you took the tests, it looks like it is all hardcoded to your monitor.

Joshua-Ashton:
Should be good to review again now.

jp7677 (approved):

jp7677:
@Joshua-Ashton merged manually into master with tests on top starting with https://github.com/jp7677/dxvk-nvapi/commit/6bcafd81317eb3175f5ed8f2aa16d58d5fdc33b5.

PR #142, opened 2023-09-11
Some DMM-related fixes and improvements

Saancreed:
We should keep games built against older NVAPI working, so instead of failing when encountering unexpected `geometryDescStrideInBytes` (like, when using NVAPI headers from before 535 in which DMMs were added causing the union to grow) this will try to access array members by manually calculating their address (and this will now keep working without rebuilding or updating headers if another union member is added in the future).

Adding tests this led me to discover a bug where we assume that after `NvAPI_D3D12_GetRaytracingAccelerationStructurePrebuildInfoEx` / `NvAPI_D3D12_BuildRaytracingAccelerationStructureEx` returns, we can still read values from pointer passed in `pGeometryDescs` which is not the case (because it points to data of a `std::vector` living on the stack of called function). To avoid this issue, the content of this array is now copied to another local `std::vector` as a side effect.

### DIFF
--- a/tests/nvapi_d3d12.cpp
+++ b/tests/nvapi_d3d12.cpp
@@ -5,6 +5,20 @@
 
 using namespace trompeloeil;
 
+// previous version of this structure before the union gained a new dmmTriangles member in R535
+typedef struct _NVAPI_D3D12_RAYTRACING_GEOMETRY_DESC_EX_R520 {
+    NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_EX type; //!< The type of geometry stored in the union of this structure.
+    D3D12_RAYTRACING_GEOMETRY_FLAGS flags;        //!< Flags affecting how this geometry is processed by the raytracing pipeline.
+    union {
+        D3D12_RAYTRACING_GEOMETRY_TRIANGLES_DESC triangles;              //!< Describes triangle geometry if \c type is #NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_TRIANGLES_EX.
+                                                                         //!< Otherwise, this parameter is unused (space repurposed in a union).
+        D3D12_RAYTRACING_GEOMETRY_AABBS_DESC aabbs;                      //!< Describes AABB geometry if \c type is #NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_PROCEDURAL_PRIMITIVE_AABBS_EX.
+                                                                         //!< Otherwise, this parameter is unused (space repurposed in a union).
+        NVAPI_D3D12_RAYTRACING_GEOMETRY_OMM_TRIANGLES_DESC ommTriangles; //!< Describes triangle geometry which may optionally use Opacity Micromaps, if \c type is #NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_OMM_TRIANGLES_EX.
+                                                                         //!< Otherwise, this parameter is unused (space repurposed in a union).
+    };
+} NVAPI_D3D12_RAYTRACING_GEOMETRY_DESC_EX_R520;
+
 TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
     D3D12Vkd3dDeviceMock device;
     D3D12Vkd3dGraphicsCommandListMock commandList;
@@ -483,8 +497,16 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
             params.pInfo = &info;
 
             D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS d3d12Desc{};
+            std::vector<D3D12_RAYTRACING_GEOMETRY_DESC> geometryDescs{};
             ALLOW_CALL(device, GetRaytracingAccelerationStructurePrebuildInfo(_, &info))
-                .LR_SIDE_EFFECT(d3d12Desc = *_1);
+                .LR_SIDE_EFFECT({
+                    d3d12Desc = *_1;
+
+                    // we know that our implementation passes pointer to data in stack-allocated vector in pGeometryDescs
+                    // so we need to copy it before the function returns and the vector is destroyed
+                    if (_1->Type == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL && _1->DescsLayout == D3D12_ELEMENTS_LAYOUT_ARRAY)
+                        geometryDescs.assign(_1->pGeometryDescs, _1->pGeometryDescs + _1->NumDescs);
+                });
 
             SECTION("GetRaytracingAccelerationStructurePrebuildInfoEx with TLAS") {
                 desc.type = D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_TOP_LEVEL;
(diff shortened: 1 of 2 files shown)


tests/nvapi_d3d12.cpp
```diff
@@ -5,6 +5,19 @@
...
+    union {
+        D3D12_RAYTRACING_GEOMETRY_TRIANGLES_DESC triangles;              //!< Describes triangle geometry if \c type is #NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_TRIANGLES_EX.
+                                                                         //!< Otherwise, this parameter is unused (space repurposed in a union).
+        D3D12_RAYTRACING_GEOMETRY_AABBS_DESC aabbs;                      //!< Describes AABB geometry if \c type is #NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_PROCEDURAL_PRIMITIVE_AABBS_EX.
+                                                                         //!< Otherwise, this parameter is unused (space repurposed in a union).
+        NVAPI_D3D12_RAYTRACING_GEOMETRY_OMM_TRIANGLES_DESC ommTriangles; //!< Describes triangle geometry which may optionally use Opacity Micromaps, if \c type is #NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_OMM_TRIANGLES_EX.
+                                                                         //!< Otherwise, this parameter is unused (space repurposed in a union).
+    };
+} NVAPI_D3D12_RAYTRACING_GEOMETRY_DESC_EX_V1;
```
jp7677, line 19: What about calling this `NVAPI_D3D12_RAYTRACING_GEOMETRY_DESC_EX_R530` and adding a comment that DMM has been added in 535 without a new struct version. 
Calling this `V1` is imho confusing since  the "usual" struct versioning isn't applied here.

tests/nvapi_d3d12.cpp
```diff
@@ -478,8 +496,10 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
             params.pInfo = &info;
 
             D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS d3d12Desc{};
+            std::vector<D3D12_RAYTRACING_GEOMETRY_DESC> geometryDescs{};
             ALLOW_CALL(device, GetRaytracingAccelerationStructurePrebuildInfo(_, &info))
-                .LR_SIDE_EFFECT(d3d12Desc = *_1);
+                .LR_SIDE_EFFECT(d3d12Desc = *_1)
+                .LR_SIDE_EFFECT(if (_1->Type == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL && _1->DescsLayout == D3D12_ELEMENTS_LAYOUT_ARRAY) geometryDescs.assign(_1->pGeometryDescs, _1->pGeometryDescs + _1->NumDescs));
```
jp7677, line 502: I would combine both calls like:
```c++
.LR_SIDE_EFFECT({
    d3d12Desc = *_1;
    if (_1->Type == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL && _1->DescsLayout == D3D12_ELEMENTS_LAYOUT_ARRAY)
       geometryDescs.assign(_1->pGeometryDescs, _1->pGeometryDescs + _1->NumDescs);
});
```
Would also be cool to have a small comment that we cannot validate `pGeometryDescs` directly from `d3d12Desc` since is has already gone out of scope after the NVAPI method call..

tests/nvapi_d3d12.cpp
```diff
@@ -500,32 +520,80 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
...
-                    REQUIRE(d3d12Desc.pGeometryDescs->Type == D3D12_RAYTRACING_GEOMETRY_TYPE_PROCEDURAL_PRIMITIVE_AABBS);
-                    REQUIRE(d3d12Desc.pGeometryDescs->AABBs.AABBCount == geometryDescEx.aabbs.AABBCount);
+                    REQUIRE(geometryDescs[0].Type == D3D12_RAYTRACING_GEOMETRY_TYPE_PROCEDURAL_PRIMITIVE_AABBS);
+                    REQUIRE(geometryDescs[0].AABBs.AABBCount == geometryDescEx[0].aabbs.AABBCount);
+                    REQUIRE(geometryDescs[1].Type == D3D12_RAYTRACING_GEOMETRY_TYPE_PROCEDURAL_PRIMITIVE_AABBS);
+                    REQUIRE(geometryDescs[1].AABBs.AABBCount == geometryDescEx[1].aabbs.AABBCount);
+                }
+            }
+
+            SECTION("GetRaytracingAccelerationStructurePrebuildInfoEx with BLAS for array of V1 structures") {
```
jp7677, line 561: See the naming comment above, I would prefer  `GetRaytracingAccelerationStructurePrebuildInfoEx with BLAS for array of R530 structures`

tests/nvapi_d3d12.cpp
```diff
@@ -581,36 +650,90 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
...
+                    REQUIRE(geometryDescs[0].Type == D3D12_RAYTRACING_GEOMETRY_TYPE_PROCEDURAL_PRIMITIVE_AABBS);
+                    REQUIRE(geometryDescs[0].AABBs.AABBCount == geometryDescEx[0].aabbs.AABBCount);
+                    REQUIRE(geometryDescs[1].Type == D3D12_RAYTRACING_GEOMETRY_TYPE_PROCEDURAL_PRIMITIVE_AABBS);
+                    REQUIRE(geometryDescs[1].AABBs.AABBCount == geometryDescEx[1].aabbs.AABBCount);
+                    REQUIRE(d3d12Desc.SourceAccelerationStructureData == desc.sourceAccelerationStructureData);
+                    REQUIRE(d3d12Desc.ScratchAccelerationStructureData == desc.scratchAccelerationStructureData);
+                }
+            }
+
+            SECTION("BuildRaytracingAccelerationStructureEx with BLAS for array of V1 structures") {
```
jp7677, line 697: See above.

tests/nvapi_d3d12.cpp
```diff
@@ -546,15 +614,16 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
...
             NVAPI_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_EX_PARAMS params{};
             params.version = NVAPI_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_EX_PARAMS_VER1;
             params.pDesc = &desc;
 
             D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_DESC d3d12Desc{};
+            std::vector<D3D12_RAYTRACING_GEOMETRY_DESC> geometryDescs{};
             ALLOW_CALL(commandList, BuildRaytracingAccelerationStructure(_, params.numPostbuildInfoDescs, params.pPostbuildInfoDescs))
-                .LR_SIDE_EFFECT(d3d12Desc = *_1);
+                .LR_SIDE_EFFECT(d3d12Desc = *_1)
+                .LR_SIDE_EFFECT(if (_1->Inputs.Type == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL && _1->Inputs.DescsLayout == D3D12_ELEMENTS_LAYOUT_ARRAY) geometryDescs.assign(_1->Inputs.pGeometryDescs, _1->Inputs.pGeometryDescs + _1->Inputs.NumDescs));
```
jp7677, line 626: See above.

jp7677 (changes requested):
Thanks for the PR, just some nits and sry for the `pGeometryDescs` bug!
I hope it wasn't to frustrating to find this...

tests/nvapi_d3d12.cpp
Saancreed: I'll check first if opacity micromaps weren't actually introduced earlier than in 530 release, if so then I'll name it after the first release it was introduced in.

By the way, is this a good place for such ad-hoc `typedef struct`? I was somewhat conflicted about where I should put it.

Saancreed: Okay, then for now I'll keep it here and when (if) we have more structures needed only in tests then we could reconsider moving them to a separate header.

jp7677 (approved):
LGTM

PR #143, opened 2023-09-12
Invalidate cache maps

jp7677:
This should especially improve test isolation.

### DIFF
--- a/src/d3d11/nvapi_d3d11_device.h
+++ b/src/d3d11/nvapi_d3d11_device.h
@@ -24,6 +24,8 @@ namespace dxvk {
         static bool CreateSamplerStateAndGetDriverHandle(ID3D11Device* pDevice, const D3D11_SAMPLER_DESC* pSamplerDesc, ID3D11SamplerState** ppSamplerState, uint32_t* pDriverHandle);
         static bool IsFatbinPTXSupported(ID3D11Device* pDevice);
 
+        static void ClearCacheMaps();
+
       private:
         inline static std::unordered_map<IUnknown*, ID3D11VkExtContext*> m_depthBoundsDeviceOrContextMap;
         inline static std::unordered_map<IUnknown*, ID3D11VkExtContext*> m_barrierControlDeviceOrContextMap;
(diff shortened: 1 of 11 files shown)


src/d3d11/nvapi_d3d11_device.h
```diff
@@ -24,6 +24,8 @@ namespace dxvk {
         static bool CreateSamplerStateAndGetDriverHandle(ID3D11Device* pDevice, const D3D11_SAMPLER_DESC* pSamplerDesc, ID3D11SamplerState** ppSamplerState, uint32_t* pDriverHandle);
         static bool IsFatbinPTXSupported(ID3D11Device* pDevice);
 
+        static void ClearCaches();
```
jp7677, line 27: nit: May be `ClearChacheMaps` sound nicer.

Saancreed: On the other hand, it also betrays some implementation details. But maybe this isn't a problem.

src/nvapi_disp.cpp
```diff
@@ -195,8 +195,8 @@ extern "C" {
                 pHDRColorDataV2->hdrDynamicRange = NV_DYNAMIC_RANGE_VESA;
                 pHDRColorDataV2->hdrBpc = data.BitsPerColor;
             } else {
+                // Ignoring some extended properties of HDRColorDataV2.
```
Saancreed, line 198: Actually, I liked the log message because we are explicitly and by design not doing what is asked of us. We could always make it print no more than once if it was too spammy.

Saancreed (approved):
Cache cleaning and tests changes are :ok_hand:, I'm just not sure how I feel about the log message. But probably it doesn't matter that much.

src/d3d12/nvapi_d3d12_device.cpp
```diff
@@ -167,4 +167,14 @@ namespace dxvk {
...
         return {};
     }
+
+    void NvapiD3d12Device::ClearCacheMaps() {
+        std::scoped_lock commandListLock(m_CommandListMutex);
+        std::scoped_lock cubinDeviceLock(m_CubinDeviceMutex);
+        std::scoped_lock cubinSmemLock(m_CubinSmemMutex);
+
+        m_cubinDeviceMap.clear();
+        m_CommandListMap.clear();
```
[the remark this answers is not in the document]
jp7677: Thanks for spotting this, renamed in a separate commit.

src/nvapi_disp.cpp
jp7677: Seeing this log statement in the logs, my thoughts were that it didn't add much and might just be confusing for users, also there is no way to "make it go away" for a user or application since it will always appear for V2 and Set (which i guess will be 90% of all usages of that method).  
But i see your point too, we can reconsider in the future.

PR #146, opened 2023-10-07
Spoof Pascal for non-NVIDIA propietary driver

jp7677:
Not that cool to do this, but seems to help to use LatencyFlex on AMD GPUs with Streamline because `NvAPI_GPU_GetArchInfo` no longer fails.

There shouldn't be any side effects since DLSS requires at least Turing.

### DIFF
--- a/src/nvapi_gpu.cpp
+++ b/src/nvapi_gpu.cpp
@@ -382,9 +382,6 @@ extern "C" {
         if (pGpuArchInfo->version != NV_GPU_ARCH_INFO_VER_1 && pGpuArchInfo->version != NV_GPU_ARCH_INFO_VER_2)
             return IncompatibleStructVersion(n);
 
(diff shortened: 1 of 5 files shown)


Saancreed (commented):
> There shouldn't be any side effects since DLSS requires at least Turing.

I'd still spoof those GPUs as Pascal for two reasons:
1. To reduce the likelyhood of some games suddenly starting to pop some annoying "This is an ancient and totally unsupported GPU!" warnings.
2. Because of https://github.com/NVIDIAGameWorks/Streamline/blob/main/source/plugins/sl.reflex/reflexEntry.cpp#L170-L182

Saancreed (approved):

FakeMichau:
Works as expected - allows the usage of latencyflex2 so latencyflex should also work.

PR #151, opened 2024-01-19
Support building with MSVC

ishitatsuyuki:
This is a series to support building dxvk-nvapi on MSVC for fun, diversity and binary size reduction.
In addition to MSVC, cross compiling from Linux with clang-cl is also supported using https://github.com/mstorsjo/msvc-wine and a cross file like https://gist.github.com/ishitatsuyuki/bfa4b4764394f94b19115b182d269f4e.

Most changes are described in the commit message. This is a list of major breaking changes for those who need to rebase:
- When declaring a new extension interface, declare the GUID in the header using `__CRT_UUID_DECL` instead of `FOO_GUID`. The `.cpp` files are deleted and the guid field no longer needs to be assigned manually.
- When writing tests, use `__uuidof` instead of `::guid` or `IID_`.
- When introducing new D3D12 mocks, methods than return a struct by value will need different mocks for MinGW and MSVC. See [tests: Make mocks compatible with MSVC COM return ABI](https://github.com/jp7677/dxvk-nvapi/commit/984c7be50699e318c34c7e4c6b7c4674939bec28) for an example.


ishitatsuyuki:
Actually, forgot one change to add dllexport to nvapi_queryinterface. One sec.

src/dxvk/dxvk_interfaces.h
```diff
@@ -188,10 +170,12 @@ IDXGIVkInteropFactory1 : public IDXGIVkInteropFactory {
...
-DXVK_DEFINE_GUID(ID3D11VkExtContext1)
+#ifndef _MSC_VER
+__CRT_UUID_DECL(IDXGIVkInteropFactory, 0x4c5e1b0d, 0xb0c8, 0x4131, 0xbf, 0xd8, 0x9b, 0x24, 0x76, 0xf7, 0xf4, 0x08);
+__CRT_UUID_DECL(IDXGIVkInteropFactory1, 0x2a289dbd, 0x2d0a, 0x4a51, 0x89, 0xf7, 0xf2, 0xad, 0xce, 0x46, 0x5c, 0xd6);
+__CRT_UUID_DECL(IDXGIVkInteropAdapter, 0x3a6d8f2c, 0xb0e8, 0x4ab4, 0xb4, 0xdc, 0x4f, 0xd2, 0x48, 0x91, 0xbf, 0xa5);
+__CRT_UUID_DECL(ID3D11VkExtDevice, 0x8a6e3c42, 0xf74c, 0x45b7, 0x82, 0x65, 0xa2, 0x31, 0xb6, 0x77, 0xca, 0x17);
+__CRT_UUID_DECL(ID3D11VkExtDevice1, 0xcfcf64ef, 0x9586, 0x46d0, 0xbc, 0xa4, 0x97, 0xcf, 0x2c, 0xa6, 0x1b, 0x06);
+__CRT_UUID_DECL(ID3D11VkExtContext, 0xfd0bca13, 0x5cb6, 0x4c3a, 0x98, 0x7e, 0x47, 0x50, 0xde, 0x2c, 0xa7, 0x91);
+__CRT_UUID_DECL(ID3D11VkExtContext1, 0x874b09b2, 0xae0b, 0x41d8, 0x84, 0x76, 0x5f, 0x3b, 0x7a, 0x0e, 0x87, 0x9d);
+#endif
```
Saancreed, line 181: Please keep the newline at the end?

src/nvapi_interface.cpp
```diff
@@ -8,7 +8,7 @@ extern "C" {
 
     static std::unordered_map<NvU32, void*> registry;
 
-    void* nvapi_QueryInterface(NvU32 id) {
+    __declspec(dllexport) void* __cdecl nvapi_QueryInterface(NvU32 id) {
```
Saancreed, line 11: Is `__cdecl` actually needed here? I believe it should be the default calling convention anyway.

Saancreed (approved):
Reasonable changes. `WIDL_EXPLICIT_AGGREGATE_RETURNS` thingy is a bit annoying but I suppose we can live with it. Just one minor request and one question, LGTM otherwise.

.github/workflows/test-build-windows.yml
```diff
@@ -0,0 +1,45 @@
+name: Test Builds on Windows
```
jp7677, line 1: Thanks a lot for add CI here! For consistency I guess we should rename the other workflow, so something like?  
`Test and package builds on Linux`
`Test builds on Windows`

I guess the tests succeed on Windows? Would be cool to add this as a step in the Windows CI as well.

jp7677 (commented):
Thanks a lot, also for fixing a few things in between. Looks really good. See a few questions.

meson.build
```diff
@@ -1,28 +1,37 @@
 project('dxvk-nvapi',
   ['cpp'],
   default_options: [
-    'cpp_std=c++17',
+    'cpp_std=c++20',
```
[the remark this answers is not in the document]
Saancreed: Hopefully this won't break Proton's CI…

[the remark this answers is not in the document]
jp7677: Yes, that's mostly why I'm asking. If there is a bump, we need to coordinate this.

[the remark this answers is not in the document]
ishitatsuyuki: It looks like MinGW GCC 9.3.0 from Soldier Proton SDK works fine compiling this branch. (FWIW, I don't think our project is buildable on Soldier anymore due to the Meson version requirement. For testing, I threw in a newer version manually.)

Works fine on Sniper SDK as well, which has MinGW GCC 10.

tests/nvapi_d3d12_mocks.h
```diff
@@ -49,7 +54,11 @@ class D3D12Vkd3dDeviceMock final : public trompeloeil::mock_interface<ID3D12Vkd3
     IMPLEMENT_MOCK1(SetStablePowerState);
     IMPLEMENT_MOCK4(CreateCommandSignature);
     IMPLEMENT_MOCK7(GetResourceTiling);
+#if defined(WIDL_EXPLICIT_AGGREGATE_RETURNS)
     MAKE_MOCK1(GetAdapterLuid, LUID*(LUID*), override);
+#else
+    MAKE_MOCK0(GetAdapterLuid, LUID(), override);
```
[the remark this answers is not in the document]
ishitatsuyuki: Nice suggestion, I forgot that the WinSDK headers don't have the overloads so the simpler macro can be used. Updated, and also mentioned it in commit message.

src/nvapi_interface.cpp
ishitatsuyuki: We have __cdecl for other NvAPI declarations, so this is more like a "why not also here" thing.

src/dxvk/dxvk_interfaces.h
ishitatsuyuki: Fixed.

meson.build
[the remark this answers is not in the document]
ivyl: Yep, we don't plan to do any updates to old Proton versions using Soldier. As @ishitatsuyuki has said the GCC version in Sniper is recent enough to not have any problems with the changes.

I gave this PR a spin inside of Proton 8's build system and it built just fine. +1 from me.

.github/workflows/test-build-windows.yml
ishitatsuyuki: > Cool to see that CI passes. Could you try to split building and testing into two steps for better readability of the CI logs? I also wouldn't mind to define `_CRT_SECURE_NO_WARNINGS` to make msvc slightly less noisy.

I did the split. I suppressed the CRT secure warnings, NvAPI deprecation warnings, and fixed the spam about GCC pragma.

> Do you really need to copy `nvapi64.dll` into the system folder? When running the test suite with `[@unit-tests]` it doesn't run the `@system` tests which rely on `nvapi64.dll`, so as far as I remember, it should be ok without it. Alternatively just extending the `PATH` would be an alternative. But no strong feelings to keep the copy action.

I removed the System32 copy for now.

> And just to be sure, CI also fails when a test fails (thus the return code isn't swallowed somewhere in powershell)?

I checked that it's not swallowed with the build below. I'm not sure if we fail fast (if build errors) but it will end up failing anyway due to missing executable.

https://github.com/ishitatsuyuki/dxvk-nvapi/actions/runs/7585231717/job/20660611025

ishitatsuyuki: FWIW we're still at 31 warnings, which is still quite a lot. But I'm not inclined to introduce more suppresses because all the warnings are kinda legit, e.g. the integer cast warnings.

jp7677: Yeah, perfectly. And I fully agree to not silence any more warnings.

jp7677 (approved):

ishitatsuyuki:
Thanks for the quick review!

PR #152, opened 2024-01-19
meta: Update usage section

jp7677:
It's long overdue to update the README usage guide.
While at it, also some smaller CI changes...

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ While originally being developed for usage with Unreal Engine 4, most notably fo
 
 ## Requirements
 
-This implementation is supposed to be used on Linux using Wine or derivatives like Proton. Usage on Windows is discouraged. Please do not replace `nvapi.dll`/`nvapi64.dll` on Windows from NVIDIA's driver package with this version. DXVK-NVAPI uses several DXVK and VKD3D-Proton extension points, thus using DXVK (D3D11 and DXGI) is a requirement. Using Wine's D3D11 or DXGI will fail. Usage of DXVK-NVAPI is not restricted to NVIDIA GPUs, though the default behavior is to skip GPUs not running the NVIDIA proprietary driver. Some entry points offer no functionality or make no sense when a different GPU vendor is detected.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -71,14 +61,18 @@ Wine does not includes DXVK-NVAPI.
 
 DXVK 1.10 and older does not support `DXVK_ENABLE_NVAPI`. Disable the `nvapiHack` in DXVK 1.10 and older with `dxgi.nvapiHack = False` set in a DXVK configuration file, see [dxvk.conf](https://github.com/doitsujin/dxvk/blob/master/dxvk.conf).
 
+### Non-NVIDIA GPUs
```
Saancreed, line 64: Perhaps we could refactor this section to begin with `### Using DXVK-NVAPI outside/without NVIDIA proprietary driver` and then add a subsection that contains additional info about non-NVIDIA GPUs. This way we can mention what would be needed to use DXVK-NVAPI with NVK (allow other drivers + optionally driver version) and then how to spoof NV with other vendors on top of that.

README.md
```diff
@@ -71,14 +61,18 @@ Wine does not includes DXVK-NVAPI.
...
+
 ## Tweaks, debugging and troubleshooting
 
 See the [DXVK-NVAPI Wiki](https://github.com/jp7677/dxvk-nvapi/wiki) for common issues and workarounds.
 
 The following environment variables tweak DXVK-NVAPI's runtime behavior:
 
 - `DXVK_NVAPI_DRIVER_VERSION` lets you override the reported driver version. Valid values are numbers between 100 and 99999. Use e.g. `DXVK_NVAPI_DRIVER_VERSION=47141` to report driver version `471.41`.
-- `DXVK_NVAPI_ALLOW_OTHER_DRIVERS`, when set to `1`, lets you use DXVK-NVAPI without running an NVIDIA card on NVIDIA's proprietary driver. Useful for using D3D11 extensions on a non-NVIDIA GPU.
+- `DXVK_NVAPI_ALLOW_OTHER_DRIVERS`, when set to `1`, lets you use DXVK-NVAPI without running an NVIDIA card on NVIDIA's proprietary driver.
```
Saancreed, line 75: Maybe mention that we will pretend all foreign devices to be Pascal GPUs (to avoid issues with DLSS) and changing that is (currently) not supported?

Saancreed (approved):
We currently have in Wine section:

> - Ensure that Wine uses the native version of `nvapi`/`nvapi64`, e.g. with `WINEDLLOVERRIDES=nvapi,nvapi64=n`.

I think this would only be needed when Wine itself is shipping `nvapi.dll` that's marked as builtin and does not have "prefer native" flag. This applies only to Wine Staging versions older than 7.22 (because anything newer has Staging nvapi patchset disabled) so we might want to specify that nowadays this is generally not needed.

Otherwise, LGTM.

jp7677:
Thanks for the useful additions @Saancreed , I've tried to incorporated  everything, please let me now if it still makes sense.

Saancreed (approved):
:+1:

PR #153, opened 2024-01-20
nvapi-gpu: Validate version before accessing struct fields

ishitatsuyuki:
The version needs to be checked before accessing any fields from the struct.

The tests for incompatible struct versions left the pOSAdapterId uninitialized, which caused flakes on MSVC. This fixes it.


jp7677 (approved):

jp7677:
Cool, thanks a lot!

PR #157, opened 2024-02-13
nvidia-gpu: Implement NvAPI_GPU_GetGPUInfo

jp7677:


src/nvapi_gpu.cpp
```diff
@@ -486,6 +486,45 @@ extern "C" {
...
+        auto architectureId = adapter->GetArchitectureId();
+
+        switch (pGpuInfo->version) {
+            case NV_GPU_INFO_VER1: {
+                auto pGpuInfoV1 = reinterpret_cast<NV_GPU_INFO_V1*>(pGpuInfo);
+                *pGpuInfoV1 = {};
+                break;
+            }
+            case NV_GPU_INFO_VER2:
+                *pGpuInfo = {};
```
Saancreed, line 514: We should probably preserve `pGpuInfo->version` as was passed on input, right? I'd just set `rayTracingCores` and `tensorCores` in this branch and (also in VER1 branch above) leave version and reserved members untouched.

jp7677: `V1` doesn't have those fields (it only has `bIsExternalGpu` and a reserved flag), but good point actually about clearing the version info. We do this at several places though, so that will be a bigger change. 

That said, even Adam used `memset` to clear a struct (https://github.com/adamnv/dxvk-nvapi/blob/c859f94b4213c5148f89a785931448d4630871e2/src/nvapi_d3d.cpp#L63) ;). So I guess in reality it would make no difference. May be something to consider for the future?

Saancreed: Oh well then, there's probably nothing that checks the returned version anyway.

Saancreed (approved):

jp7677:
Turns out this affects https://github.com/HansKristian-Work/vkd3d-proton/issues/1148 . RTX is available on NVIDIA without any spoofing when `NvAPI_GPU_GetGPUInfo` is implemented returns some numbers. The actual numbers probably doesn't matter, as long as > 0.

PR #160, opened 2024-03-01
ci: Remove gcc problem matcher

jp7677:
It's getting outdated. The original project hasn't seen much activity either :(.


Saancreed (approved):
:disappointed:

PR #167, opened 2024-03-12
sysinfo: Report 999.99 as version for  non-NVIDIA proprietary driver

jp7677:
Reporting Mesa version numbers turned out to be not very useful.

@Saancreed Any thoughts?


Saancreed (approved):
I expected users (or, well, Proton) to just always set this to some value they consider to be reasonable but this is a better fallback anyway.

PR #168, opened 2024-03-12
Initialize and report architecture/capabilities when using NVK

jp7677:
Do this unconditionally.


src/sysinfo/nvapi_adapter.cpp
```diff
@@ -200,7 +204,7 @@ namespace dxvk {
     }
 
     NV_GPU_ARCHITECTURE_ID NvapiAdapter::GetArchitectureId() const {
-        if (!this->HasNvProprietaryDriver()) {
+        if (!this->HasNvProprietaryDriver() && !this->HasNvkDriver()) {
```
Saancreed, line 207: This is one place I'm not too sure about. Except for Ada GPUs, detection based on support of various Vulkan extensions might be… not so reliable until NVK gets support for them, and there's no guarantee NVK will bother implementing NV(X) vendor extensions.

An idea for another day but maybe we could swap the extensions checked for something less specific to whatever NV proprietary driver supports on each arch? E.g. check for `VK_EXT_mesh_shader` to detect Turing-or-newer?

Saancreed (approved):
One potential issue… but probably not something we'd need to address immediately. LGTM otherwise; let's see how unprepared video games are for the coming of NVK :see_no_evil:

Sid127:
works on Tower Unite and Deep Rock Galactic, tested by dropping the dll into proton 9 beta which doesn't have any nvk specific workarounds yet.

This getting merged would also need [dxvk](https://github.com/doitsujin/dxvk/commit/0414bbe2d5d5ee0df027731a91f251312ca84cf4) and wine-proton [here](https://github.com/ValveSoftware/Proton/commit/e1ee8227c8f10dbd33ee2c43a549675ea05d59d4) and [here](https://github.com/ValveSoftware/wine/commit/caa6104097b8b14001acd7ee787c8d814a9304d9) to be reverted, to ensure none of the components interfere with each other.

src/sysinfo/nvapi_adapter.cpp
Sid127: also, correct me if I'm wrong, but should it not be an || instead of && ?

jp7677:
Reminder: log statement and readme

src/sysinfo/nvapi_adapter.cpp
jp7677: > This is one place I'm not too sure about. Except for Ada GPUs, detection based on support of various Vulkan extensions might be… not so reliable until NVK gets support for them, and there's no guarantee NVK will bother implementing NV(X) vendor extensions.
> 
> An idea for another day but maybe we could swap the extensions checked for something less specific to whatever NV proprietary driver supports on each arch? E.g. check for `VK_EXT_mesh_shader` to detect Turing-or-newer?

ah yeah, thanks, that's a very valid point. Using `VK_EXT_mesh_shader` sounds indeed good. I would also prefer to have a valid detection for the older generations. From what I remember, NVK is currently Pascal and newer so we would for now just need something to identify `Volta`. That would change ofc once NVK starts to support even older generations. I don't know if there is a limit for NVK.

@Sid127 the check should be good, note the `!` , also the test seem to confirm that it works correctly.

jp7677: PS: https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/nouveau/vulkan/nvk_physical_device.c?ref_type=heads#L230 and https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/27196

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -211,27 +228,39 @@ namespace dxvk {
...
+        // VK_NVX_image_view_handle is supported on Volta and newer on the NVIDIA proprietary driver
+        // VK_EXT_depth_clip_control's depthClipControl is supported on Volta and newer on NVK
+        if ((HasNvProprietaryDriver() && IsVkDeviceExtensionSupported(VK_NVX_IMAGE_VIEW_HANDLE_EXTENSION_NAME))
+            || (HasNvkDriver()
+                && IsVkDeviceExtensionSupported(VK_EXT_DEPTH_CLIP_CONTROL_EXTENSION_NAME)
+                && m_vkDepthClipControlFeatures.depthClipControl))
             return NV_GPU_ARCHITECTURE_GV100;
 
         // VK_NV_clip_space_w_scaling is supported on Pascal and newer
+        // TODO: We haven't found a way yet to identify Pascal on NVK
```
jp7677, line 256: Actually, I think this is `Pascal` https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/nouveau/vulkan/nvk_physical_device.c?ref_type=heads#L682 (also according to https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/28159/diffs#059d99b2edae5a4b51e79eb38f6f299f37f26831)

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -211,27 +228,39 @@ namespace dxvk {
...
+        // TODO: We haven't found a way yet to identify Ampere on NVK
         if (IsVkDeviceExtensionSupported(VK_KHR_FRAGMENT_SHADING_RATE_EXTENSION_NAME)
             && m_vkFragmentShadingRateProperties.primitiveFragmentShadingRateWithMultipleViewports)
             return NV_GPU_ARCHITECTURE_GA100;
 
-        // Variable rate shading is supported on Turing and newer
-        if (IsVkDeviceExtensionSupported(VK_NV_SHADING_RATE_IMAGE_EXTENSION_NAME))
+        // Mesh shader is supported on Turing and newer
+        // This needs https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/27196 on NVK
+        if (IsVkDeviceExtensionSupported(VK_EXT_MESH_SHADER_EXTENSION_NAME))
```
jp7677, line 244: May be `VK_KHR_fragment_shader_barycentric` is better since that is already implemented in NVK (https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/nouveau/vulkan/nvk_physical_device.c?ref_type=heads#L101)? Alternatively `VK_KHR_draw_indirect_count`?

Saancreed: `VK_KHR_fragment_shader_barycentric`, yes, we could use that as well. `VK_KHR_draw_indirect_count`, not really unless we guard this behind `HasNvkDriver` because [NV proprietary driver seems to support this on older architectures as well](https://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_KHR_draw_indirect_count&platform=all).

src/sysinfo/nvapi_adapter.cpp
Saancreed: It's Pascal indeed (MR has been merged now) and from a quick look at gpuinfo this appears to match what NV proprietary reports.

jp7677:
@Saancreed @Sid127 This should be good now (though still not able to identify Ampere on NVK). Would be cool if you could take another look/try.

src/sysinfo/nvapi_adapter.cpp
jp7677: I went for `VK_KHR_fragment_shader_barycentric`.

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -211,27 +228,38 @@ namespace dxvk {
...
 
+        // See https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/nouveau/vulkan/nvk_physical_device.c
+        // for NVK properties and features
+
         // KHR_fragment_shading_rate's
-        // primitiveFragmentShadingRateWithMultipleViewports is supported on
-        // Ampere and newer
+        // primitiveFragmentShadingRateWithMultipleViewports is supported on Ampere and newer
+        // TODO: We haven't found a way yet to identify Ampere on NVK
         if (IsVkDeviceExtensionSupported(VK_KHR_FRAGMENT_SHADING_RATE_EXTENSION_NAME)
```
Saancreed, line 237: Perhaps we should add `HasNvProprietaryDriver() &&` here for now?

Saancreed (approved):
Let's hope arch detection won't be a headache to maintain.

src/sysinfo/nvapi_adapter.cpp
jp7677: Yeah i was thinking about this too. My hopes were, that once NVK gets that extension, it will have the same architecture requirements as on the NV prop driver, so that we just have to remove the TODO at some point. But that might be indeed too optimistic.

jp7677: Done.

Sid127:
different DLSS scenarios seem to be working correctly, both on Ada and GTX Turing. Nevermind the fact that DLSS does not work, but most games tested with seem to allow the toggles for DLSS and Reflex with this, and have them inaccessible without.

Behavior with various proton env vars was also verified, and it all seemed to be working right, except for the HIDE_NVIDIA_GPU which Saancreed was made aware of and fixed already.

Games tested:
Baldurs Gate 3, Doom Eternal, Deep Rock Galactic, Tower Unite (no DLSS to begin with), Quake Champions (no DLSS, Reflex exists though)

Sid127 (approved):

jp7677:
Thanks for the confirmation @Sid127 !

PR #170, opened 2024-03-21
nvapi: Guard against null device in Reflex calls

Saancreed:
Fixes a crash in Horizon Forbidden West.


esullivan-nvidia:
Maybe it would be better to have the Reflex NVAPI entry points return NVAPI_INVALID_ARGUMENT if the device is null?

Saancreed:
> Maybe it would be better to have the Reflex NVAPI entry points return NVAPI_INVALID_ARGUMENT if the device is null?

Yeah, good point. I intend to also check how native NVAPI behaves on Windows here first but something like `NVAPI_INVALID_ARGUMENT` or `NVAPI_INVALID_POINTER` sounds likely to be the case.

jp7677 (approved):

jp7677:
Reference: https://github.com/ValveSoftware/Proton/issues/7590

PR #173, opened 2024-05-15
sysinfo: Replace deprecated Vulkan macros

jp7677:
Unfortunately we cannot just use the newer Vulkan macros since the major version for the NVIDIA driver needs 10 bits whereas the newer Vulkan macro for the major version truncates to 7 bits.

From `vulkan_core.h`:

```c++
// DEPRECATED: This define is deprecated. VK_API_VERSION_MAJOR should be used instead.
#define VK_VERSION_MAJOR(version) ((uint32_t)(version) >> 22U)

#define VK_API_VERSION_MAJOR(version) (((uint32_t)(version) >> 22U) & 0x7FU)
```
The newer `VK_API_VERSION_MAJOR` cuts after 7 bits due to introduction of the `variant` number, this conflicts with the 10 bits needed for the NVIDIA major version.

Inspired by https://github.com/doitsujin/dxvk/pull/4001

### DIFF
--- a/src/util/util_version.h
+++ b/src/util/util_version.h
@@ -1,0 +1,29 @@
+#pragma once
+
+#include "../nvapi_private.h"
+
+namespace dxvk {
+
+    // The Vulkan version number packing used by NVIDIA is slightly different
+    // from the packing specified in the Vulkan specifications:
+    // - The major version is a 10-bit integer packed into bits 31-22
+    // - The minor version is an 8-bit integer packed into bits 21-14
+    // - The patch version is an 8-bit integer packed into bits 13-6
+    // See https://registry.khronos.org/vulkan/specs/1.3/html/vkspec.html#extendingvulkan-coreversions-versionnumbers
+
+    inline uint32_t nvMakeVersion(const uint32_t major, const uint32_t minor, const uint32_t patch) {
+        return (major << 22U) | (minor << 14U) | (patch << 6U);
+    }
+
+    inline uint32_t nvVersionMajor(const uint32_t version) {
+        return (version >> 22U) & 0x3FFU;
+    }
+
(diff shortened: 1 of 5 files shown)


src/util/util_version.h
```diff
@@ -0,0 +1,29 @@
...
+
+    // The Vulkan version number packing used by NVIDIA is slightly different
+    // from the packing specified in the Vulkan specifications:
+    // - The major version is a 9 bit integer packed into bits 31-22
+    // - The minor version is an 8-bit integer packed into bits 21-14
+    // - The patch version is an 8-bit integer packed into bits 13-6
+    // See https://registry.khronos.org/vulkan/specs/1.3/html/vkspec.html#extendingvulkan-coreversions-versionnumbers
+
+    inline uint32_t nvMakeVersion(const uint32_t major, const uint32_t minor, const uint32_t patch) {
+        return ((((uint32_t)(major)) << 22U) | (((uint32_t)(minor)) << 14U) | (((uint32_t)(patch)) << 6U));
```
Saancreed, line 15: Aren't casts to `uint32_t` redundant when we have a proper function and not just a macro? Types should already match so it should be safe to just drop them.

Saancreed (approved):
Sure. I have one suggestion but otherwise LGTM.

Is this something worth writing tests for?

src/util/util_version.h
jp7677: Good point, thx.

jp7677:
> Is this something worth writing tests for?

We already somewhat cover it indirectly https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_sysinfo.cpp#L202 but yeah good idea actually

PR #176, opened 2024-05-26
Add a few trivial functions

jp7677:
I looked again at the War Thunder issue. Add the three functions here didn't changed anything wrt the DLSS crashes, but since they are fairly straight forward, we might just keep them.


src/sysinfo/nvapi_adapter.cpp
```diff
@@ -206,6 +206,11 @@ namespace dxvk {
         return m_dxgiDesc.DedicatedVideoMemory / 1024;
     }
 
+    uint32_t NvapiAdapter::GetVirtualVRamSize() const {
+        // Report VRAM size from DXVK to honor memory overrides
+        return (m_dxgiDesc.DedicatedVideoMemory + m_dxgiDesc.DedicatedSystemMemory) / 1024;
```
jp7677, line 211: In reality `DedicatedSystemMemory` will always be zero, see https://github.com/doitsujin/dxvk/blob/master/src/dxgi/dxgi_adapter.cpp#L371

Saancreed (approved):

PR #179, opened 2024-05-28
nvapi: Add argument checks and a few things found on the road

jp7677:
Also omit nullptr checks for handles.

This one is probably a bit tricky ...


jp7677:
> What about `NvAPI_D3D_SetSleepMode` and `NvAPI_D3D_GetSleepStatus`? We unsafely dereference second parameter of each function. Not sure if there's anything else in that area but at least these two functions come to mind.

Thanks, missed those two and also `NvAPI_D3D_GetLatency`. (forced) Pushed.

PR #184, opened 2024-06-12
Try to improve Reflex command queue lookups

Saancreed:
This lets us avoid having to `QueryInterface` on every call to `NvAPI_D3D12_SetAsyncFrameMarker`. The consequence of doing it like this is that now there might be multiple entries in `m_lowLatencyDeviceMap` pointing to the same `ID3DLowLatencyDevice` so in order to still correctly have a single frame ID generator per `ID3DLowLatencyDevice`, their map is now keyed on low latency devices which we have to retrieve on every call anyway.

We could further reduce the lookups if we either returned intermediate `ID3DLowLatencyDevice` from `SupportsLowLatency` and then acted on it in the following call or if we just got rid of explicit `SupportsLowLatency` and had methods like `SetFrameMarker` returned more detailed error codes (or just `NvAPI_Status` directly). But that's an idea for another day.

### DIFF
--- a/src/d3d/nvapi_d3d_low_latency_device.cpp
+++ b/src/d3d/nvapi_d3d_low_latency_device.cpp
@@ -48,6 +48,14 @@ namespace dxvk {
         return d3dLowLatencyDevice->SupportsLowLatency();
     }
 
+    bool NvapiD3dLowLatencyDevice::SupportsLowLatency(ID3D12CommandQueue* commandQueue) {
+        auto d3dLowLatencyDevice = GetLowLatencyDevice(commandQueue);
+        if (d3dLowLatencyDevice == nullptr)
+            return false;
+
+        return d3dLowLatencyDevice->SupportsLowLatency();
+    }
+
     bool NvapiD3dLowLatencyDevice::LatencySleep(IUnknown* device) {
         auto d3dLowLatencyDevice = GetLowLatencyDevice(device);
         if (d3dLowLatencyDevice == nullptr)
@@ -93,6 +101,15 @@ namespace dxvk {
             GetFrameIdGenerator(d3dLowLatencyDevice.ptr())->GetLowLatencyDeviceFrameId(frameID), markerType));
     }
 
+    bool NvapiD3dLowLatencyDevice::SetLatencyMarker(ID3D12CommandQueue* commandQueue, uint64_t frameID, uint32_t markerType) {
+        auto d3dLowLatencyDevice = GetLowLatencyDevice(commandQueue);
+        if (d3dLowLatencyDevice == nullptr)
+            return false;
+
+        return SUCCEEDED(d3dLowLatencyDevice->SetLatencyMarker(
+            GetFrameIdGenerator(d3dLowLatencyDevice.ptr())->GetLowLatencyDeviceFrameId(frameID), markerType));
+    }
+
     void NvapiD3dLowLatencyDevice::ClearCacheMaps() {
         std::scoped_lock lock(m_lowLatencyDeviceMutex, m_lowLatencyFrameIdGeneratorMutex);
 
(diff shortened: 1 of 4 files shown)


Saancreed:
Changed it a bit, now there are separate methods with overloads for `IUnknown` which is presumed to be some kind of `ID3D1[12]Device` and `ID3D12CommandQueue` where appropriate instead of a single one that blindly tries to handle both, which should limit the amount of weird DXVK and vkd3d-proton logs about unsupported `QueryInterface` calls.

src/d3d/nvapi_d3d_low_latency_device.h
```diff
@@ -26,21 +26,24 @@ namespace dxvk {
     class NvapiD3dLowLatencyDevice {
       public:
         static bool SupportsLowLatency(IUnknown* device);
+        static bool SupportsLowLatency(ID3D12CommandQueue* commandQueue);
         static bool LatencySleep(IUnknown* device);
         static bool SetLatencySleepMode(IUnknown* device, bool lowLatencyMode, bool lowLatencyBoost, uint32_t minimumIntervalUs);
         static bool GetLatencyInfo(IUnknown* device, D3D_LATENCY_RESULTS* latencyResults);
         static bool SetLatencyMarker(IUnknown* device, uint64_t frameID, uint32_t markerType);
+        static bool SetLatencyMarker(ID3D12CommandQueue* commandQueue, uint64_t frameID, uint32_t markerType);
```
Saancreed, line 34: I was wondering if this shouldn't be called `SetAsyncFrameMarker` to better match Reflex API but I feel like it would be lying about the implementation. So I'm not sure which name would be preferable here.

jp7677: I think `SetLatencyMarker` is fine since we call this thing `LowLatencyDevice` and not `ReflexDevice`.

tests/nvapi_d3d12.cpp
```diff
@@ -803,6 +804,10 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
 
         auto e = ConfigureDefaultTestEnvironment(*dxgiFactory, *vulkan, *nvml, *lfx, adapter, output);
 
+        ALLOW_CALL(commandQueue, GetDevice(__uuidof(ID3DLowLatencyDevice), _))
+            .LR_SIDE_EFFECT(*_2 = static_cast<ID3DLowLatencyDevice*>(&lowLatencyDevice))
+            .LR_SIDE_EFFECT(lowLatencyDeviceRefCount++)
```
jp7677, line 809: While at it, I think our refcounts should be zero with your changes, so we could check this at the end of this section 

```
        REQUIRE(deviceRefCount == 0);
        REQUIRE(commandListRefCount == 0);
        REQUIRE(lowLatencyDeviceRefCount == 0);
```

src/d3d/nvapi_d3d_low_latency_device.cpp
```diff
@@ -118,7 +135,27 @@ namespace dxvk {
         return d3dLowLatencyDevice;
     }
 
-    LowLatencyFrameIdGenerator* NvapiD3dLowLatencyDevice::GetFrameIdGenerator(IUnknown* device) {
+    Com<ID3DLowLatencyDevice> NvapiD3dLowLatencyDevice::GetLowLatencyDevice(ID3D12CommandQueue* commandQueue) {
```
jp7677, line 138: Some duplication with the other `GetLowLatencyDevice` and also same map, but I don't see a more elegant way either.

jp7677 (approved):
Small suggestion for the tests, otherwise LGTM! Thanks for diving into this.

tests/nvapi_d3d12.cpp
Saancreed: Like this? I'm not sure if I can just put such checks at the end of aggregate section or if they should be replicated inside each innermost section.

src/d3d/nvapi_d3d_low_latency_device.cpp
Saancreed: At worst we could extract common parts to another method and pass the way to extract `ID3DLowLatencyDevice` in another parameter as something like `std::function` but I feel like that would be greatly overengineered solution. Duplication, as long as it's limited like this, seems more readable.

tests/nvapi_d3d12.cpp
jp7677: Yes, from my understanding of Catch this should be good.

src/d3d/nvapi_d3d_low_latency_device.cpp
jp7677: Yes, let’s keep your version, it won’t get better :)

PR #186, opened 2024-06-15
util: Spoof Ada and later GPUs as Ampere for Ghost of Tsushima

Saancreed:
See https://github.com/HansKristian-Work/vkd3d-proton/issues/2008 and https://github.com/ValveSoftware/Proton/issues/7735#issuecomment-2136100714 for the background and descriptions of the issue this change aims to resolve.

### DIFF
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -117,10 +117,19 @@ namespace dxvk::env {
         }
     }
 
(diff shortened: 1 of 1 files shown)


jp7677 (approved):
LGTM! If you have another minute to spare, prefixing the commit message with `util` would be more consistent with earlier app switches.

PR #188, opened 2024-06-20
Update NVAPI headers to R555

jp7677:
.. and adjust `NvAPI_D3D12_SetAsyncFrameMarker` signature (new `presentFrameID` field).


jp7677:
While on it, added the two new trivial functions.

inc/nvapi.h
```diff
@@ -19739,6 +19856,7 @@ typedef enum
 {
     OUT_OF_BAND_RENDER  = 0,
     OUT_OF_BAND_PRESENT = 1,
+    OUT_OF_BAND_IGNORE  = 2,
 } NV_OUT_OF_BAND_CQ_TYPE;
```
jp7677, line 19860: We are casting this to https://github.com/HansKristian-Work/vkd3d-proton/blob/master/include/vkd3d_vk_includes.h#L48

inc/nvapi.h
```diff
@@ -19703,12 +19810,22 @@ typedef NV_LATENCY_MARKER_PARAMS_V1            NV_LATENCY_MARKER_PARAMS;
...
+//!
 //! Used in NvAPI_D3D12_SetAsyncFrameMarker
 //! \ingroup dx
-typedef NV_LATENCY_MARKER_PARAMS_V1         NV_ASYNC_FRAME_MARKER_PARAMS_V1;
+typedef struct _NV_ASYNC_FRAME_MARKER_PARAMS_V1
+{
+    NvU32  version;                                       //!< (IN) Structure version
+    NvU64  frameID;
+    NV_LATENCY_MARKER_TYPE markerType;
+    NvU64  presentFrameID;
```
jp7677, line 19822: `presentFrameID` is new compared to `NV_LATENCY_MARKER_PARAMS`.

src/nvapi_sys.cpp
```diff
@@ -89,4 +89,48 @@ extern "C" {
...
+        if (log::tracing())
+            log::trace(n, log::fmt::ptr(pPhysicalGPUs));
+
+        if (nvapiAdapterRegistry == nullptr)
+            return ApiNotInitialized(n);
+
+        if (pPhysicalGPUs == nullptr)
+            return InvalidArgument(n);
+
+        for (auto i = 0U; i < nvapiAdapterRegistry->GetAdapterCount(); i++) {
```
Saancreed, line 105: We should check the struct version beforehand. Would be a good idea to verify Windows behavior here.

Same in `NvAPI_SYS_GetLogicalGPUs`.

src/nvapi_sys.cpp
```diff
@@ -89,4 +89,48 @@ extern "C" {
...
+            pPhysicalGPUs->gpuHandleData[i].hPhysicalGpu = reinterpret_cast<NvPhysicalGpuHandle>(nvapiAdapterRegistry->GetAdapter(i));
+            pPhysicalGPUs->gpuHandleData[i].adapterType = NV_ADAPTER_TYPE_UNKNOWN;
```
Saancreed, line 107: I think it would be safer to report `NV_ADAPTER_TYPE_WDDM` instead, afaik it's the model you expect to see when you want to render some stuff and present it to the screen, so it should be the best match for GPUs that aren't in data centers or some compute-only clusters. Or call NVML to check but be aware that `wine-nvml` will just always return WDDM :stuck_out_tongue:

Same in `NvAPI_SYS_GetLogicalGPUs`.

tests/nvapi_sysinfo_topo.cpp
```diff
@@ -347,4 +347,34 @@ TEST_CASE("Topology methods succeed", "[.sysinfo-topo]") {
             REQUIRE(NvAPI_GPU_GetConnectedDisplayIds(gpuHandles[0U], gpu0DisplayIds, &gpu0DisplayIdCount, 0) != NVAPI_INCOMPATIBLE_STRUCT_VERSION);
         }
     }
+
+    SECTION("GetPhysicalGPUs returns OK") {
+        NvPhysicalGpuHandle handles[NVAPI_MAX_LOGICAL_GPUS]{};
```
Saancreed, line 352: `NvPhysicalGpuHandle` but `NVAPI_MAX_LOGICAL_GPUS`?

tests/nvapi_sysinfo_topo.cpp
```diff
@@ -347,4 +347,34 @@ TEST_CASE("Topology methods succeed", "[.sysinfo-topo]") {
...
+        NvU32 count;
+        REQUIRE(NvAPI_EnumPhysicalGPUs(handles, &count) == NVAPI_OK);
+
+        NV_PHYSICAL_GPUS gpus{};
+        REQUIRE(NvAPI_SYS_GetPhysicalGPUs(&gpus) == NVAPI_OK);
+        REQUIRE(gpus.gpuHandleCount > 0);
+        REQUIRE(gpus.gpuHandleCount == count);
+        for (auto i = 0U; i <= count; i++) {
```
Saancreed, line 360: Shouldn't we stop before `count`? Or are you checking one-past-end entry if it's zeroed by choice?

jp7677: No, just a bit tired actually ;)

src/nvapi_sys.cpp
jp7677: Thanks, forget about the versioning, I've added the usual ceremony.

src/nvapi_sys.cpp
jp7677: Makes sense.

Saancreed (approved):

inc/nvapi.h
jp7677: Hi @esullivan-nvidia , do you know what needs to be done here when `OUT_OF_BAND_IGNORE` has been provided as argument in `NvAPI_D3D12_NotifyOutOfBandCommandQueue`?

Saancreed:
By the way, I just noticed that R555 open source headers now include shader extension opcodes. We could import these files as well and [`fromCode`](https://github.com/jp7677/dxvk-nvapi/blob/master/src/util/util_op_code.h) at least.

src/nvapi_private.h
```diff
@@ -49,3 +49,7 @@
 #elif defined(_MSC_VER)
 #include <intrin.h>
 #endif
+
+// clang-format off
+#define MAP_ENUM_VALUE(v) {v, #v}
```
jp7677, line 54: There is probably a better name..

Saancreed: I think it's fine.

However, `// clang-format off` above looks suspicious. Does autoformatting break this horribly somehow?

jp7677: I added this because of https://github.com/jp7677/dxvk-nvapi/actions/runs/9650777826/job/26617157512
Unfortunately I didn’t hit this locally, so I took this shortcut.

Saancreed: Ah, weird. We might be suffering from infamous `clang-format` version mismatch, although suppressing it for one `#define` is not a big deal.

inc/nvapi.h
jp7677: Handling this with an early bail out for now https://github.com/jp7677/dxvk-nvapi/pull/188/commits/5c277564589bf62c50b9ac4da2bc4c7d370d8c2a

src/nvapi_d3d12.cpp
```diff
@@ -474,33 +475,41 @@ extern "C" {
         if (nvapiD3dInstance->IsUsingLfx() || !NvapiD3dLowLatencyDevice::SupportsLowLatency(pCommandQueue))
             return NoImplementation(n);
 
+        if (cqType == OUT_OF_BAND_IGNORE) {
+            if (!alreadyLoggedNotSupported)
+                log::info("NvAPI_D3D12_NotifyOutOfBandCommandQueue is called with OUT_OF_BAND_IGNORE, uncertain how to handle this type.");
+
+            // This would to an error anyway https://github.com/HansKristian-Work/vkd3d-proton/blob/master/libs/vkd3d/command_queue_vkd3d_ext.c#L67
+            return NotSupported(n, alreadyLoggedNotSupported);
```
Saancreed, line 483: While logging this explicitly makes sense, I'd just let vkd3d-proton fail for us. This way we aren't blocking anything and should vkd3d-proton add support for this in the future, the message would become slightly misleading but not problematic.

Also, this would cause a funny result of us failing the call for `OUT_OF_BAND_IGNORE` but not for any other yet-undefined value, while dxvk-nvapi built against R550 headers would pass the value as-is and possibly work correctly with some future version of vkd3d-proton.

```suggestion
            if (!std::exchange(alreadyLoggedNotSupported, true))
                log::info("NvAPI_D3D12_NotifyOutOfBandCommandQueue called with OUT_OF_BAND_IGNORE.");
```

Saancreed (approved):
:ok_hand: but see the suggestion.

src/nvapi_d3d12.cpp
jp7677: My reasoning was to bail out fast with a clearer response code than just "Error". I'm quite certain that either me or your would catch up fast enough with vkd3d-proton if something changes there.  But I'm realizing that this change might be a small hurdle when somebody _is_ developing and testing the needed changes for vkd3d-proton. Will go for your suggestion.

inc/nvapi.h
esullivan-nvidia: Sorry, I just saw this in my inbox. I think the early bail out for now is the best approach. We don't have any concept of this out of band queue type in LL2 so there isn't much we can do anyways. I will get in touch if we need to make any additional changes.

jp7677: Thanks for your confirmation!

PR #193, opened 2024-08-02
Switch NVAPI headers to a git submodule and bump them to r560

Saancreed:
Headers published in https://github.com/NVIDIA/nvapi are now MIT-licensed so let's use them. The SAL sorcery makes it a bit tricky, however. There's an include tree that looks like this:

```
#include "nvapi.h"
|-- #include "nvapi_lite_salstart.h"
|   |-- (definitions for `__success`, `NVAPI_INTERFACE` and other stuff)
|
|-- #include "nvapi_lite_common.h"
|   |-- #include "nvapi_lite_salstart.h"
|   |-- (common types)
|   |-- #include "nvapi_lite_salend.h"
|
|-- #include "nvapi_lite_sli.h"
|   |-- #include "nvapi_lite_salstart.h"
|   |-- #include "nvapi_lite_common.h"
|   |   |-- #include "nvapi_lite_salstart.h"
|   |   |-- (common types)
|   |   |-- #include "nvapi_lite_salend.h"
|   |
|   |-- (SLI stuff here)
|   |-- #include "nvapi_lite_salend.h"
|
|-- #include "nvapi_lite_surround.h"
|   |-- #include "nvapi_lite_salstart.h"
|   |-- #include "nvapi_lite_common.h"
|   |   |-- #include "nvapi_lite_salstart.h"
|   |   |-- (common types)
|   |   |-- #include "nvapi_lite_salend.h"
|   |
|   |-- (surround stuff here)
|   |-- #include "nvapi_lite_salend.h"
|
|-- #include "nvapi_lite_stereo.h"
|   |-- #include "nvapi_lite_salstart.h"
|   |-- #include "nvapi_lite_common.h"
|   |   |-- #include "nvapi_lite_salstart.h"
|   |   |-- (common types)
|   |   |-- #include "nvapi_lite_salend.h"
|   |
|   |-- (stereo stuff here)
|   |-- #include "nvapi_lite_salend.h"
|
|-- #include "nvapi_lite_d3dext.h"
|   |-- #include "nvapi_lite_salstart.h"
|   |-- #include "nvapi_lite_common.h"
|   |   |-- #include "nvapi_lite_salstart.h"
|   |   |-- (common types)
|   |   |-- #include "nvapi_lite_salend.h"
|   |
|   |-- (D3D stuff here)
|   |-- #include "nvapi_lite_salend.h"
|
|-- (important stuff like `NVAPI_INTERFACE NvAPI_Initialize();`)
|-- #include "nvapi_lite_salend.h"
```

Whenever `nvapi_lite_salstart.h` is `#include`d, important macros like `__in`, `__out` and `__success` are `#define`d and whenever `nvapi_lite_salend.h` is `#include`d, they are `#undef`ined unless `__NVAPI_EMPTY_SAL` was defined earlier. But if we do nothing then `nvapi_lite_salend.h` included from `nvapi_lite_d3dext.h` will `#undef` then before `nvapi.h` itself can use them, so the only place where we `#include "nvapi.h"` now uses the following ceremony:

```cpp
#define __NVAPI_EMPTY_SAL
#include "nvapi.h"
#undef __NVAPI_EMPTY_SAL
#include "nvapi_lite_salend.h"
```

This ensures that important symbols won't get `#undef`ined before we get to the core `nvapi.h` and instead we do the cleanup ourselves manually (which we have to do because leaving defines such as `__out` will mess up STL headers).

There might be some less hackish solution to it but right now I don't see one so let's hope this works.


jp7677 (approved):
Very nice, thanks!

PR #195, opened 2024-08-08
util: Log the struct version when returning IncompatibleStructVersion

Saancreed:


src/util/util_statuscode.h
```diff
@@ -83,8 +83,8 @@ namespace dxvk {
         return NVAPI_EXPECTED_LOGICAL_GPU_HANDLE;
     }
 
-    inline NvAPI_Status IncompatibleStructVersion(const std::string& logMessage) {
-        log::info(str::format("<-", logMessage, ": Incompatible struct version"));
+    inline NvAPI_Status IncompatibleStructVersion(const std::string& logMessage, NvU32 version) {
+        log::info(str::format("<-", logMessage, ": Incompatible struct version: ", version));
```
jp7677, line 87: Just a suggestion for consistency with how we trace function argument:
```c++
log::info(str::format("<-", logMessage, ": Incompatible struct version (", version, ")"));
```

No strong preference though, I'm also good with your version.

Saancreed: You have a point, I'll change.

jp7677 (approved):

PR #198, opened 2024-08-14
nvapi-gpu: Add NvAPI_GPU_GetMemoryInfo/Ex

### LINKED ISSUE #197: Add NvAPI_GPU_GetMemoryInfo
It turns out this is needed to solve https://github.com/doitsujin/dxvk/issues/4198
See https://github.com/SveSop/dxvk-nvapi/commit/f2c58cee33bb0c7fdc2becc44f49c1c7da342883 for an implementation. 

Reminder for later ...

jp7677:
TODO: 

- [x] Look at the currently unknown fields  
- [x] Tests

Closes #197


jp7677:
cc @SveSop since you are the original author.

Saancreed:
Isn't available memory pretty much the same thing as `Budget` that's returned from [IDXGIAdapter3::QueryVideoMemoryInfo](https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_4/nf-dxgi1_4-idxgiadapter3-queryvideomemoryinfo)?

src/nvapi_gpu.cpp
```diff
@@ -382,11 +382,102 @@ extern "C" {
...
+                auto pMemoryInfoV3 = reinterpret_cast<NV_DISPLAY_DRIVER_MEMORY_INFO_V3*>(pMemoryInfo);
+                pMemoryInfoV3->dedicatedVideoMemory = adapter->GetVRamSize();
+                pMemoryInfoV3->availableDedicatedVideoMemory = 0; // Not found in DXVK
+                pMemoryInfoV3->systemVideoMemory = adapter->GetDedicatedSystemRamSize();
+                pMemoryInfoV3->sharedSystemMemory = adapter->GetSharedSystemRamSize();
+                pMemoryInfoV3->curAvailableDedicatedVideoMemory = 0; // Not found in DXVK
+                pMemoryInfoV3->dedicatedVideoMemoryEvictionsSize = 0;
+                pMemoryInfoV3->dedicatedVideoMemoryEvictionCount = 0;
+                break;
+            }
```
Saancreed, line 435: I wonder, can we just use the given pointer as-it without reinterpret-casting it? I think it should be possible because the offset of each member appears to be the same in every struct version in which it's present. That should allow us to make this a bit more DRY, if you don't mind some classic switch fallthrough:

```suggestion
            case NV_DISPLAY_DRIVER_MEMORY_INFO_VER_3: 
                pMemoryInfo->dedicatedVideoMemoryEvictionsSize = 0;
                pMemoryInfo->dedicatedVideoMemoryEvictionCount = 0;
                [[fallthrough]];
            case NV_DISPLAY_DRIVER_MEMORY_INFO_VER_2:
                pMemoryInfo->curAvailableDedicatedVideoMemory = …;
                [[fallthrough]];
            case NV_DISPLAY_DRIVER_MEMORY_INFO_VER_1:
                pMemoryInfo->dedicatedVideoMemory = adapter->GetVRamSize();
                pMemoryInfo->availableDedicatedVideoMemory = …;
                pMemoryInfo->systemVideoMemory = adapter->GetDedicatedSystemRamSize();                
                pMemoryInfo->sharedSystemMemory = adapter->GetSharedSystemRamSize();
                break;
```

jp7677: yes, this looks cool, we should probably do this for other places too.

jp7677:
> Isn't available memory pretty much the same thing as `Budget` that's returned from [IDXGIAdapter3::QueryVideoMemoryInfo](https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_4/nf-dxgi1_4-idxgiadapter3-queryvideomemoryinfo)?

Yes, see last WIP commit, I'm also thinking/playing in that direction, with thanks to @SveSop for his hint this morning.

src/sysinfo/nvapi_adapter.h
```diff
@@ -48,17 +58,19 @@ namespace dxvk {
       private:
         Vulkan& m_vulkan;
         Nvml& m_nvml;
+        Com<IDXGIAdapter3> m_dxgiAdapter;
```
jp7677, line 61: I guess no way around to keep an adapter alive. I hope this won't give us tear down issues in titles.

src/nvapi_gpu.cpp
```diff
@@ -382,11 +382,106 @@ extern "C" {
...
+                pMemoryInfoV1->sharedSystemMemory = adapter->GetVideoMemory().SharedSystemMemory;
+                break;
+            }
+            case NV_DISPLAY_DRIVER_MEMORY_INFO_VER_2: {
+                auto pMemoryInfoV2 = reinterpret_cast<NV_DISPLAY_DRIVER_MEMORY_INFO_V2*>(pMemoryInfo);
+                pMemoryInfoV2->dedicatedVideoMemory = adapter->GetVideoMemory().DedicatedVideoMemory;
+                pMemoryInfoV2->availableDedicatedVideoMemory = adapter->GetVideoMemory().Budget;
+                pMemoryInfoV2->systemVideoMemory = adapter->GetVideoMemory().DedicatedSystemMemory;
+                pMemoryInfoV2->sharedSystemMemory = adapter->GetVideoMemory().SharedSystemMemory;
+                pMemoryInfoV2->curAvailableDedicatedVideoMemory = adapter->GetVideoMemory().Budget - adapter->GetVideoMemory().CurrentUsage;
```
jp7677, line 424: Not sure about this one. The others seem to roughly match.

src/sysinfo/nvapi_adapter.h
```diff
@@ -25,10 +35,10 @@ namespace dxvk {
         [[nodiscard]] uint32_t GetPciBusId() const;
         [[nodiscard]] uint32_t GetPciDeviceId() const;
         [[nodiscard]] uint32_t GetBoardId() const;
-        [[nodiscard]] uint32_t GetVRamSize() const;
-        [[nodiscard]] uint32_t GetVirtualVRamSize() const;
         [[nodiscard]] std::optional<LUID> GetLuid() const;
         [[nodiscard]] NV_GPU_ARCHITECTURE_ID GetArchitectureId() const;
+        void UpdateVideoMemory();
```
Saancreed, line 40: I somewhat don't like this, having to remember to call it before checking the values is extra mental load _and_ it makes it unable to query this if we ever got a const reference to an adapter. Can we instead persist only values that should be immutable for given adapter and query budget and current usage via some `[[nodiscard]] struct VideoMemoryInfo GetCurrentVideoMemoryInfo() const;`?

src/sysinfo/nvapi_adapter.h
```diff
@@ -10,10 +10,20 @@ namespace dxvk {
...
 
-        bool Initialize(Com<IDXGIAdapter1>& dxgiAdapter, uint32_t index, std::vector<NvapiOutput*>& outputs);
+        struct VideoMemory {
+            uint64_t DedicatedVideoMemory;
+            uint64_t DedicatedSystemMemory;
+            uint64_t SharedSystemMemory;
+            uint64_t Budget;
+            uint64_t CurrentUsage;
+            uint64_t AvailableForReservation;
+            uint64_t CurrentReservation;
```
Saancreed, line 23: Do we need reservation values anywhere? If not, I'd ignore them completely on our side because DXVK doesn't exactly report anything useful in there.

src/sysinfo/nvapi_adapter.h
Saancreed: Unless adapters are singletons (which I don't think they are) it should be fine. But if it proves to be a problem we can always try switching to manual usage of `VK_EXT_memory_budget` on our own.

src/nvapi_gpu.cpp
Saancreed: Do I understand correctly that:

* `dedicatedVideoMemory` is simply the total size of device-local heap and should probably never change,
* `availableDedicatedVideoMemory` is how much of above is left for the application to allocate, so `dedicatedVideoMemory` reduced by other application's allocations and areas reserved for private use by the driver, if any (and current application allocating memory shouldn't affect this),
* `curAvailableDedicatedVideoMemory` is how much of above is left for the application to allocate after we include allocations it already made, so `availableDedicatedVideoMemory` reduced by application's own allocated memory size?

SveSop:
@Saancreed `DedicatedVideoMemory` is total vram - eg. 8192MB or whatnot
`availableDedicatedVideoMemory` seemed to be `DedicatedvideoMemory` - some pre-allocation that did not seem to change on windows no matter what load/app was running. It was around 2%, so 300'ish MB on my 16GB card, and 200'ish MB on my 8GB card.
`curAvailableDedicatedVideoMemory` is the actual remaining.. so if you have some app using 2GB of video memory on a 8GB card, it would be 6GB. Would possibly like to verify this somehow if it would be `DedicatedVideoMemory` - usage, or `availableDedicatedVideoMemory` - usage...

src/sysinfo/nvapi_adapter.h
jp7677: yeah makes sense. Changed, though not yet completely happy with the naming.

src/nvapi_gpu.cpp
jp7677: This is indeed more or less how I understood the terminology, based on the nvapi headers and @SveSop 's numbers:

```c++
    NvU64   dedicatedVideoMemory;              //!< Size(in bytes) of the physical framebuffer. Refers to the dedicated video memory on discrete GPUs.
                                               //!  It is more performant for GPU operations than the reserved systemVideoMemory.
    NvU64   availableDedicatedVideoMemory;     //!< Size(in bytes) of the available physical framebuffer for allocating video memory surfaces.
    NvU64   systemVideoMemory;                 //!< Size(in bytes) of system memory the driver allocates at load time. It is a substitute for dedicated video memory.
                                               //!< Typically used with integrated GPUs that do not have dedicated video memory.
    NvU64   sharedSystemMemory;                //!< Size(in bytes) of shared system memory that driver is allowed to commit for surfaces across all allocations.
                                               //!< On discrete GPUs, it is used to utilize system memory for various operations. It does not need to be reserved during boot.
                                               //!< It may be used by both GPU and CPU, and has an �on-demand� type of usage.
    NvU64   curAvailableDedicatedVideoMemory;  //!< Size(in bytes) of the current available physical framebuffer for allocating video memory surfaces. 
```

SveSop:
Me thinks the calculation of `availableDedicatedVideoMemory` is not overly correct, as that goes down when video memory is used when testing with wine... On Windows it does not.

Then again, `dedicatedVideoMemory` is also calculated wrong, as it seems to overshoot a couple of hundred MB's

On windows i get on my 2070 8GB card: 8192MB, but using dxvk-nvapi i get 8438MB.

Might be some method that DXVK calculates this as it might be adding some extra "heap" stuff-whatever? (Safeguards or whatnot)

I have a small test proggy i have compiled that runs some nvapi tests that i attach.
[nvapi_test.zip](https://github.com/user-attachments/files/16630395/nvapi_test.zip)

Saancreed:
Thanks for the explanation @SveSop. Interestingly, for me nvapi system tests print

```
    GPU 0
    GPU name:                   NVIDIA GeForce RTX 4080 Laptop GPU
    GPU type:                   2 (Discrete)
    Device ID:                  0x27e010de
    Subsystem ID:               0x21ed1043
    Bus:Slot ID:                PCI:01:00
    Board ID:                   0x100
    Dedicated video memory:     12282MB
    Available dedicated v.m.:   11898MB
    System video memory:        0MB
    Shared system memory:       47937MB
    Cur. avail. dedicated v.m.: 11898MB
```
so dedicated video memory matches exactly. This also matches how Vulkan memory heaps are reported, which afaik is what DXVK sums up when reporting memory budget info:
```
memoryHeaps: count = 2
        memoryHeaps[0]:
                size   = 12878610432 (0x2ffa00000) (11.99 GiB)
                budget = 12476612608 (0x2e7aa0000) (11.62 GiB)
                usage  = 0 (0x00000000) (0.00 B)
                flags: count = 1
                        MEMORY_HEAP_DEVICE_LOCAL_BIT
        memoryHeaps[1]:
                size   = 50266269696 (0xbb41a6800) (46.81 GiB)
                budget = 50266269696 (0xbb41a6800) (46.81 GiB)
                usage  = 0 (0x00000000) (0.00 B)
                flags:
                        None
```

As another data point, `nvidia-smi` reports `6MiB /  12282MiB` when no processes are running and NVML's `nvmlDeviceGetMemoryInfo_v2` reports:

```
Total: 12878610432, in MiB: 12282
Reserved: 396034048, in MiB: 377.688
Free: 12476612608, in MiB: 11898.6
Used: 5963776, in MiB: 5.6875
```

So yeah, there is ~378 MiB of VRAM that's reserved for the driver and GPU firmware which we should probably subtract from `dedicatedVideoMemory` when calculating `availableDedicatedVideoMemory` but I'm not sure if we can get that number outside of NVML.

jp7677:
I'm reaching the same conclusion. What we have now is:

```C++
pMemoryInfoV1->dedicatedVideoMemory = vidMemInfo.DedicatedVideoMemory;
pMemoryInfoV1->systemVideoMemory = vidMemInfo.DedicatedSystemMemory;
pMemoryInfoV1->sharedSystemMemory = vidMemInfo.SharedSystemMemory;
pMemoryInfoV1->availableDedicatedVideoMemory = curVidMemInfo.Budget;
pMemoryInfoV1->curAvailableDedicatedVideoMemory = curVidMemInfo.Budget - curVidMemInfo.CurrentUsage;
```

but what we actually want to have is:

```C++
...
pMemoryInfoV1->availableDedicatedVideoMemory = vidMemInfo.DedicatedVideoMemory - vidMemInfo.Reseverd;
...
```

But since we don't have `vidMemInfo.Reseverd` without NVML I guess 

```C++
...
pMemoryInfoV1->availableDedicatedVideoMemory = vidMemInfo.DedicatedVideoMemory;
...
```

is indeed the best we can do (that is, without NVML). That way we have a stable number, but we are also overshooting. 

Just blindly subtracting 2% from `DedicatedVideoMemory` crossed my mind, but not sure if that is a good idea, probably not.

jp7677:
I suggest to create two follow up issues for picking up after this PR:
- Refactor all versioned struct handling like suggested in https://github.com/jp7677/dxvk-nvapi/pull/198#discussion_r1718709030
- Obtain reserved memory from NVML

jp7677:
PS: Still not completely sure if it should be 

```c++
pMemoryInfoV1->curAvailableDedicatedVideoMemory = curVidMemInfo.Budget - curVidMemInfo.CurrentUsage;
```

or 

```c++
pMemoryInfoV1->curAvailableDedicatedVideoMemory = curVidMemInfo.Budget;
```

So, does `curAvailableDedicatedVideoMemory` includes (like Budget) the apps own allocations or not.

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -130,9 +130,14 @@ namespace dxvk {
 
             nvmlDevice_t nvmlDevice{};
             auto result = m_nvml.DeviceGetHandleByPciBusId_v2(pciId, &nvmlDevice);
-            if (result == NVML_SUCCESS)
+            if (result == NVML_SUCCESS) {
                 m_nvmlDevice = nvmlDevice;
-            else
+
+                nvmlMemory_v2_t memory{};
```
Saancreed, line 136: You should set the `version` field to `nvmlMemory_v2` here, otherwise the call below will most likely fail.

Saancreed (commented):
One comment and we should really check if we should subtract usage from budget because I would guess we shouldn't but otherwise LGTM.

src/sysinfo/nvapi_adapter.cpp
jp7677: I'm setting the version here https://github.com/jp7677/dxvk-nvapi/blob/fc0b9a7ab8e9763359ad8d38a6fd4cbbd24d4e3b/src/sysinfo/nvml.cpp#L81

jp7677:
> we should subtract usage from budget because I would guess we shouldn't but otherwise LGTM.

Yes, sleeping a night over it, I also think that we should not. This would also better fit terminology since `availableDedicatedVideoMemory` also does not includes application usage.

src/sysinfo/nvapi_adapter.cpp
```diff
@@ -120,9 +130,15 @@ namespace dxvk {
...
+                memory.version = 2;
```
Saancreed, line 137: Are you sure the value should be just `2` and not `nvmlMemory_v2`? I made such a mistake while testing and it caused the call to return `NVML_FUNCTION_NOT_FOUND` for me :/

jp7677: You are right ofc, fixed and also validated the NVML path:

```    
Dedicated video memory:     24576MB
Available dedicated v.m.:   24131MB
System video memory:        0MB
Shared system memory:       48119MB
Cur. avail. dedicated v.m.: 23787MB
```

The documentation tricked me ;) https://docs.nvidia.com/deploy/nvml-api/structnvmlMemory__v2__t.html#structnvmlMemory__v2__t states  `Structure format version (must be 2).`

Saancreed (approved):

PR #200, opened 2024-08-18
d3d12: Report other capabilities

jp7677:
For Turing+ only.

I'm tempted to report those two also for Turing+.
Since there are no `NvAPI_D3D12` equivalents to `NvAPI_D3D11_RSSetExclusiveScissorRects` or  `NvAPI_D3D11_RSSetViewportsPixelShadingRates` I guess those capabilities refer to D3D12 core functionalities, but should do some research into it first. 
Reporting those probably makes no difference in practice.

### DIFF
--- a/src/nvapi_d3d12.cpp
+++ b/src/nvapi_d3d12.cpp
@@ -192,8 +192,10 @@ extern "C" {
 
         pGraphicsCaps->majorSMVersion = 0;
         pGraphicsCaps->minorSMVersion = 0;
-        // all Vulkan drivers are expected to support ZBC clear without padding
+        // All Vulkan drivers are expected to support ZBC clear without padding
         pGraphicsCaps->bFastUAVClearSupported = true;
+        pGraphicsCaps->bExclusiveScissorRectsSupported = false;
+        pGraphicsCaps->bVariablePixelRateShadingSupported = false;
 
         NvapiAdapter* adapter = nullptr;
         auto luid = NvapiD3d12Device::GetLuid(pDevice);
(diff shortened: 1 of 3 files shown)


Saancreed:
`bVariablePixelRateShadingSupported` is simple enough I think, afaict this refers to feature that's accessible using [`ID3D12GraphicsCommandList5::RSSetShadingRate`](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist5-rssetshadingrate). Though I'd recommend checking if `VK_KHR_fragment_shading_rate` extension has been enabled on underlying `VkDevice`.

`bExclusiveScissorRectsSupported` looks a bit tougher though. It sounds like something that is exposed by `VK_NV_scissor_exclusive` extension in Vulkan but vkd3d-proton doesn't use it and I don't see a way to access a feature like this neither in D3D12 core API nor in NVAPI D3D12 extensions. Maybe it would be safer to keep reporting `false` here.

jp7677:
> `bVariablePixelRateShadingSupported` is simple enough I think, afaict this refers to feature that's accessible using [`ID3D12GraphicsCommandList5::RSSetShadingRate`](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist5-rssetshadingrate). Though I'd recommend checking if `VK_KHR_fragment_shading_rate` extension has been enabled on underlying `VkDevice`.

Good suggestion. Did that, which should result in the same.
 
> `bExclusiveScissorRectsSupported` looks a bit tougher though. It sounds like something that is exposed by `VK_NV_scissor_exclusive` extension in Vulkan but vkd3d-proton doesn't use it and I don't see a way to access a feature like this neither in D3D12 core API nor in NVAPI D3D12 extensions. Maybe it would be safer to keep reporting `false` here.

I couldn't find anything related to d3d12 either, this hints in your direction https://developer.nvidia.com/vulkan-turing though. Returning false seems sensible.

src/nvapi_d3d12.cpp
```diff
@@ -203,6 +207,8 @@ extern "C" {
         if (adapter == nullptr || (!adapter->HasNvProprietaryDriver() && !adapter->HasNvkDriver()))
             return Ok(str::format(n, " (sm_0)"));
 
+        pGraphicsCaps->bVariablePixelRateShadingSupported = adapter->IsVariablePixelRateShadingSupported();
```
Saancreed, line 210: In the incredibly unlikely scenario that we are running this code alongside some _ancient_ version of vkd3d-proton (or usage of the ext is disabled via `VKD3D_DISABLE_EXTENSIONS`), the safer choice would be to check which extensions were actually enabled on the logical device instead. That would look something like

```cpp
Com<ID3D12DXVKInteropDevice> device;
if (SUCCEEDED(pDevice->QueryInterface(IID_PPV_ARGS(&device)))) {
    std::vector<const char*> extensions;
    UINT count;
    if (SUCCEEDED(device->GetDeviceExtensions(&count, nullptr))) {
        extensions.resize(count);
        if (SUCCEEDED(device->GetDeviceExtensions(&count, extensions.data())))
            pGraphicsCaps->bVariablePixelRateShadingSupported = std::any_of(extensions.begin(), extensions.end(), [](auto ext) { return ext == std::string_view{VK_KHR_FRAGMENT_SHADING_RATE_EXTENSION_NAME}; });
    }
}
```

but it's fairly disgusting and probably overcomplicated for such a rare corner-case, so if you are fine with this small risk then let's just leave a comment that would explain potential mismatch and ignore the problem for now.

Saancreed (approved):
One warning but it's not necessarily something we need to act on.

PR #202, opened 2024-08-20
nvapi-gpu: Report size in KB for NvAPI_GPU_GetMemoryInfo

jp7677:
Only NvAPI_GPU_GetMemoryInfoEx wants bytes.


Saancreed (approved):
Derp indeed.

Saancreed:
Missed some test fixup though.

Issue #199, opened 2024-08-17
Use classic switch fallthrough for versionized structs


jp7677:
See https://github.com/jp7677/dxvk-nvapi/pull/198#discussion_r1718709030
We should do this consistently everywhere, if possible.

jp7677:
This is indeed trickier than it looks since a lots of structs do not just contain additions prior to earlier versions, so we can't do this everywhere consistently.

PR #206, opened 2024-08-24
nvapi: Use fallthrough when checking struct versions

SveSop:
And skip casting versioned structs that should be the correct version

Attempt to solve #199 

Some of these makes a lot of sense, others - not so sure. Just for consistency (i hope). Will work on it.
In theory it should not be necessary to reinterpret_cast structs that should already be the correct version, as pointed out in https://github.com/jp7677/dxvk-nvapi/pull/198#discussion_r1718709030

### DIFF
--- a/src/nvapi_d3d.cpp
+++ b/src/nvapi_d3d.cpp
@@ -65,18 +65,11 @@ extern "C" {
             return InvalidArgument(n);
 
         switch (pSliState->version) {
-            case NV_GET_CURRENT_SLI_STATE_VER1: {
-                auto pSliStateV1 = reinterpret_cast<NV_GET_CURRENT_SLI_STATE_V1*>(pSliState);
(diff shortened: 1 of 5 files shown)


jp7677:
The crashed were a result of the changes in `NvAPI_GPU_CudaEnumComputeCapableGpus`, here the v2 struct had non-backwards compatible changes, even if it compiled. So yeah, this requires careful inspection of the different struct versions.

SveSop:
> The crashed were a result of the changes in `NvAPI_GPU_CudaEnumComputeCapableGpus`, here the v2 struct had non-backwards compatible changes, even if it compiled. So yeah, this requires careful inspection of the different struct versions.

Indeed. But that was not the only one. I think i spot 2 more of those that would en up with the same problem as this one, so that would throw consistency out the window, since you would need to cast those anyway.

PR #207, opened 2024-08-25
Ensure that we don't lose struct version

jp7677:
Based on logs of `NvAPI_GPU_GetMemoryInfo` usage, structs can and will be reused. The one's where we zero'ed the struct version are probably not of that category, but lets be consistent and safe everywhere.

TODO:
- [x] Test  `NvAPI_D3D11_MultiGPU_GetCaps` explicitly with Senua's Sacrifice VR, that title made us introduce that function

### DIFF
--- a/src/nvapi_d3d11.cpp
+++ b/src/nvapi_d3d11.cpp
@@ -69,25 +69,26 @@ extern "C" {
         if (pMultiGPUCaps == nullptr)
             return InvalidArgument(n);
 
-        switch (pMultiGPUCaps->version) {
-            case 0: // NV_MULTIGPU_CAPS_V1 has no version field
-            case NV_MULTIGPU_CAPS_VER1: {
+        auto version = pMultiGPUCaps->version;
+        switch (version) { // NOLINT(*-multiway-paths-covered)
+            case NV_MULTIGPU_CAPS_VER2:
+                *pMultiGPUCaps = {};
+                pMultiGPUCaps->version = version;
+                pMultiGPUCaps->multiGPUVersion = 3; // Observed on Windows
+                pMultiGPUCaps->nTotalGPUs = nvapiAdapterRegistry->GetAdapterCount();
+                // Report on purpose that no SLI GPUs are available by
+                // keeping pMultiGPUCaps->nSLIGPUs at zero
+                break;
+            default: {
+                // NV_MULTIGPU_CAPS_V1 has no version field
                 auto pMultiGPUCapsV1 = reinterpret_cast<NV_MULTIGPU_CAPS_V1*>(pMultiGPUCaps);
                 *pMultiGPUCapsV1 = {};
-                // Report that SLI is not available
+                pMultiGPUCapsV1->reserved = version;  // This is a union field with version in V2
+                pMultiGPUCapsV1->multiGPUVersion = 3; // Observed on Windows
                 pMultiGPUCapsV1->nTotalGPUs = nvapiAdapterRegistry->GetAdapterCount();
-                pMultiGPUCapsV1->nSLIGPUs = 0;
+                // See above about reporting no SLI
                 break;
             }
-            case NV_MULTIGPU_CAPS_VER2: {
-                *pMultiGPUCaps = {};
-                // Report that SLI is not available
-                pMultiGPUCaps->nTotalGPUs = nvapiAdapterRegistry->GetAdapterCount();
-                pMultiGPUCaps->nSLIGPUs = 0;
-                break;
-            }
-            default:
-                return IncompatibleStructVersion(n, pMultiGPUCaps->version);
         }
 
         return Ok(n);
--- a/.clang-tidy-ignore
+++ b/.clang-tidy-ignore
@@ -1,0 +1,2 @@
+external/*
+inc/*
--- a/src/nvapi_disp.cpp
+++ b/src/nvapi_disp.cpp
@@ -59,6 +59,7 @@ extern "C" {
             case NV_HDR_CAPABILITIES_VER1: {
                 auto pHdrCapabilitiesV1 = reinterpret_cast<NV_HDR_CAPABILITIES_V1*>(pHdrCapabilities);
                 *pHdrCapabilitiesV1 = {};
+                pHdrCapabilitiesV1->version = NV_HDR_CAPABILITIES_VER1;
                 pHdrCapabilitiesV1->isST2084EotfSupported = data.HasST2084Support;
                 pHdrCapabilitiesV1->display_data.displayPrimary_x0 = data.RedPrimaryX;
                 pHdrCapabilitiesV1->display_data.displayPrimary_y0 = data.RedPrimaryY;
@@ -76,6 +77,7 @@ extern "C" {
             case NV_HDR_CAPABILITIES_VER2: {
                 auto pHdrCapabilitiesV2 = reinterpret_cast<NV_HDR_CAPABILITIES_V2*>(pHdrCapabilities);
                 *pHdrCapabilitiesV2 = {};
+                pHdrCapabilitiesV2->version = NV_HDR_CAPABILITIES_VER2;
                 pHdrCapabilitiesV2->isST2084EotfSupported = data.HasST2084Support;
                 pHdrCapabilitiesV2->display_data.displayPrimary_x0 = data.RedPrimaryX;
                 pHdrCapabilitiesV2->display_data.displayPrimary_y0 = data.RedPrimaryY;
@@ -92,6 +94,7 @@ extern "C" {
             }
             case NV_HDR_CAPABILITIES_VER3:
                 *pHdrCapabilities = {};
+                pHdrCapabilities->version = NV_HDR_CAPABILITIES_VER3;
                 pHdrCapabilities->isST2084EotfSupported = data.HasST2084Support;
                 pHdrCapabilities->display_data.displayPrimary_x0 = data.RedPrimaryX;
                 pHdrCapabilities->display_data.displayPrimary_y0 = data.RedPrimaryY;
--- a/src/nvapi_gpu.cpp
+++ b/src/nvapi_gpu.cpp
@@ -36,10 +36,12 @@ extern "C" {
             return InsufficientBuffer(n);
         }
 
-        switch (pDisplayIds->version) {
+        auto version = pDisplayIds->version;
+        switch (version) {
             case NV_GPU_DISPLAYIDS_VER1: // Both versions use the same NV_GPU_DISPLAYIDS struct
             case NV_GPU_DISPLAYIDS_VER2: {
                 *pDisplayIds = {};
+                pDisplayIds->version = version;
                 for (auto i = 0U; i < count; i++) {
                     auto output = nvapiAdapterRegistry->GetOutput(adapter, i);
                     pDisplayIds[i].displayId = output->GetId();
@@ -693,10 +695,12 @@ extern "C" {
             case NV_GPU_INFO_VER1: {
                 auto pGpuInfoV1 = reinterpret_cast<NV_GPU_INFO_V1*>(pGpuInfo);
                 *pGpuInfoV1 = {};
+                pGpuInfoV1->version = NV_GPU_INFO_VER1;
                 break;
             }
             case NV_GPU_INFO_VER2:
                 *pGpuInfo = {};
+                pGpuInfo->version = NV_GPU_INFO_VER2;
                 if (architectureId >= NV_GPU_ARCHITECTURE_TU100) {
                     // Values are taken from RTX4080
                     pGpuInfo->rayTracingCores = 76;
--- a/tests/nvapi_d3d11.cpp
+++ b/tests/nvapi_d3d11.cpp
@@ -453,6 +453,7 @@ TEST_CASE("D3D11 MultiGPU methods succeed", "[.d3d11]") {
         NV_MULTIGPU_CAPS_V1 multiGPUCaps{};
 
         REQUIRE(NvAPI_D3D11_MultiGPU_GetCaps(reinterpret_cast<PNV_MULTIGPU_CAPS>(&multiGPUCaps)) == NVAPI_OK);
+        REQUIRE(multiGPUCaps.multiGPUVersion == 3);
         REQUIRE(multiGPUCaps.nTotalGPUs == 1);
         REQUIRE(multiGPUCaps.nSLIGPUs == 0);
     }
@@ -462,21 +463,9 @@ TEST_CASE("D3D11 MultiGPU methods succeed", "[.d3d11]") {
         multiGPUCaps.version = NV_MULTIGPU_CAPS_VER2;
 
         REQUIRE(NvAPI_D3D11_MultiGPU_GetCaps(&multiGPUCaps) == NVAPI_OK);
+        REQUIRE(multiGPUCaps.multiGPUVersion == 3);
         REQUIRE(multiGPUCaps.nTotalGPUs == 1);
         REQUIRE(multiGPUCaps.nSLIGPUs == 0);
-    }
-
-    SECTION("MultiGPU_GetCaps with unknown struct version returns incompatible-struct-version") {
-        NV_MULTIGPU_CAPS multiGPUCaps{};
-        multiGPUCaps.version = NV_MULTIGPU_CAPS_VER2 + 1;
-        REQUIRE(NvAPI_D3D11_MultiGPU_GetCaps(&multiGPUCaps) == NVAPI_INCOMPATIBLE_STRUCT_VERSION);
-    }
-
-    SECTION("MultiGPU_GetCaps with current struct version returns not incompatible-struct-version") {
-        // This test fails when a header update provides a newer not yet implemented struct version
-        NV_MULTIGPU_CAPS multiGPUCaps{};
-        multiGPUCaps.version = NV_MULTIGPU_CAPS_VER;
-        REQUIRE(NvAPI_D3D11_MultiGPU_GetCaps(&multiGPUCaps) != NVAPI_INCOMPATIBLE_STRUCT_VERSION);
     }
 
     SECTION("MultiGPU_Init returns OK") {


Saancreed:
That `NvAPI_D3D11_MultiGPU_GetCaps` change is the only one I'm not sure about (the rest are fine). Documentation about the struct it accepts is fairly disappointing so I would feel safer if we knew Windows NVAPI behavior here. I can try to figure it out around tomorrow or the day after, if you don't mind waiting a bit.

Also, have a multiple instances of a typo in PR title and commit messages where "loose" should be "lose".

jp7677:
> That `NvAPI_D3D11_MultiGPU_GetCaps` change is the only one I'm not sure about (the rest are fine). Documentation about the struct it accepts is fairly disappointing so I would feel safer if we knew Windows NVAPI behavior here. I can try to figure it out around tomorrow or the day after, if you don't mind waiting a bit.

That would be nice, thanks. I'm also going to test the title that was affected by this (the todo item). 
 
> Also, have a multiple instances of a typo in PR title and commit messages where "loose" should be "lose".

Oh embarrassing, fixed ;)

SveSop:
https://github.com/jp7677/dxvk-nvapi/blob/36affbe6c768822ddaef99f86af0128defae11e1/src/nvapi_d3d.cpp#L121

jp7677:
> https://github.com/jp7677/dxvk-nvapi/blob/36affbe6c768822ddaef99f86af0128defae11e1/src/nvapi_d3d.cpp#L121

That struct doesn't have a version field, here it come as a separate argument in the function, See https://github.com/jp7677/dxvk-nvapi/blob/36affbe6c768822ddaef99f86af0128defae11e1/src/nvapi_d3d.cpp#L108

jp7677:
> That `NvAPI_D3D11_MultiGPU_GetCaps` change is the only one I'm not sure about (the rest are fine). Documentation about the struct it accepts is fairly disappointing so I would feel safer if we knew Windows NVAPI behavior here. I can try to figure it out around tomorrow or the day after, if you don't mind waiting a bit.

Tested with `Senua's Sacrifice VR`, that title kept `version` and `multiGPUVersion` at `0`, so added that specific case as well. I don't know if we should also support uninitialized version fields, so random garbage. Would be cool if you could include that scenario when testing against Windows.

SveSop:
I did a test with this:
`NvAPI_SYS_GetDisplayDriverInfo(NV_DISPLAY_DRIVER_INFO* pDriverInfo);`
And i "set" the version by doing:
`NV_DISPLAY_DRIVER_INFO_V2 pDriverInfo = { NV_DISPLAY_DRIVER_INFO_VER };`

The value in the version field is then: 131216 (from the make_ver macro).

If i then set it using:
`NV_DISPLAY_DRIVER_INFO_V2 pDriverInfo = { NV_DISPLAY_DRIVER_INFO_VER1 };`

the value in the version field is: 65612
Calling `NvAPI_SYS_GetDisplayDriverInfo` by setting some random number in the version field returns -9 `NVAPI_INCOMPATIBLE_STRUCT_VERSION`, so it seems that is working as it is supposed to do

Testing the call to `NvAPI_D3D11_MultiGPU_GetCaps(PNV_MULTIGPU_CAPS pMultiGPUCaps);`
with `NV_MULTIGPU_CAPS_V2 pMultiGPUCaps = { NV_MULTIGPU_CAPS_VER2 };` vs `NV_MULTIGPU_CAPS_V2 pMultiGPUCaps = { NV_MULTIGPU_CAPS_VER1 };` returns 3 for `multiGPUVersion`

Calling it by setting `pMultiGPUCaps.multiGPUVersion = 1234;` , still returns 3.
Setting `pMultiGPUCaps.version` to some random number also does not really seem to do anything, and returns `NVAPI_OK` without changing `pMultiGPUCaps.version` field.

So, not really sure.

Saancreed:
> Setting `pMultiGPUCaps.version` to some random number also does not really seem to do anything, and returns `NVAPI_OK` without changing `pMultiGPUCaps.version` field.

If I were to guess, it should control whether `NvLinkPresent` and `fastNvLinkReads` fields are written into or not, with value of `version` equal to `NV_MULTIGPU_CAPS_VER2` meaning "please write them", value equal to `0` or `NV_MULTIGPU_CAPS_VER1` meaning "don't" and others being unspecified… which would be nice to know (but hopefully we won't have to handle cases of other garbage here).

The fact that you never saw `NVAPI_INCOMPATIBLE_STRUCT_VERSION`, even for random versions, is strange and perhaps a bit concerning.

But it does seem to support the theory where `version` is strictly an input field (and never touched by the function) and `multiGPUVersion` is strictly an output field (and is never read by the function, only written to).

SveSop:
> But it does seem to support the theory where `version` is strictly an input field (and never touched by the function) and `multiGPUVersion` is strictly an output field (and is never read by the function, only written to).

Does seem like this. What i have not really tested is to use the "V1" struct when calling, as this means i have to create a "new" function call to the same address with a different struct, as using the `nvapi.h` header will default to setting it to "latest version".

I can do a couple of tests for this just for lulz and see if it changes things, but from other nvapi calls, fiddling with the `->version` field will return `NVAPI_INCOMPATIBLE_STRUCT_VERSION`, so it is indeed a bit strange.

The V1 struct did not really seem to have a `->version` field, so it could be somewhat of a internal nvapi tweak NVIDIA have done as a safeguard "in case of old versions", that this field is ignored? `multiGPUVersion` might be some sort of indication of "supported SLI version whatever"?? 4070 that i tested this on does not support SLI afaik, so i do not know what it really means.

SveSop:
V1 input:
```
NV_MULTIGPU_CAPS_V1 pMultiGPUCapsV1 = { NV_MULTIGPU_CAPS_VER1 };
pMultiGPUCapsV1.nTotalGPUs = 0xbbed;
pMultiGPUCapsV1.nSLIGPUs = 0xffef;
pMultiGPUCapsV1.videoBridgePresent = 0xbbce;
pMultiGPUCapsV1.multiGPUVersion = 0xefe;
```
V1 return:
```
D3D11 MultiGPU Caps (V1):
MGPU Version  : 3
Total GPUs    : 1
SLI GPUs      : 1
VideoBridge   : 0
```

V2 input:
```
NV_MULTIGPU_CAPS_V2 pMultiGPUCaps = { NV_MULTIGPU_CAPS_VER2 };
pMultiGPUCaps.nTotalGPUs = 0xbbed;
pMultiGPUCaps.nSLIGPUs = 0xffef;
pMultiGPUCaps.videoBridgePresent = 0xbbce;
pMultiGPUCaps.multiGPUVersion = 0xefe;
pMultiGPUCaps.version = 0xef;
pMultiGPUCaps.NvLinkPresent = 0xfef;
pMultiGPUCaps.fastNvLinkReads = 0xabc;
```

V2 return:
```
D3D11 MultiGPU Caps (V2):
MGPU Version  : 3
Struct version: 239
Total GPUs    : 1
SLI GPUs      : 1
VideoBridge   : 0
NV Link       : 4079
NV link Reads : 2748
```

So.. the `pMultiGPUCaps.NvLinkPresent` and `pMultiGPUCaps.fastNvLinkReads` will return the input number. If this is some sort of setting, i have no idea of. Maybe you are supposed to "set" that you are using NvLinkPresent (eg. True - 1)? Using other numbers does not change anything, nor does setting the version field to 0 or whatever number.

Without testing this on an actual SLI setup, the settings may just be disregarded? Maybe `pMultiGPUCaps.NvLinkPresent` would return 1 in a SLI setup? Dunno. I suppose SLI is not overly supported with DXVK anyway, so...

Setting `NV_MULTIGPU_CAPS_V2 pMultiGPUCaps = { NV_MULTIGPU_CAPS_VER1 };` but using the V2 struct did not change this behavior.

Saancreed:
>  V2 input:
> `NV_MULTIGPU_CAPS_V2 pMultiGPUCaps = { NV_MULTIGPU_CAPS_VER2 };`

I've meant to say "try `NV_MULTIGPU_CAPS_V2 pMultiGPUCaps = { .version = NV_MULTIGPU_CAPS_VER2 };`" instead of what you did, sorry for being imprecise :sweat_smile:

Saancreed:
> Doing this for `NvAPI_GPU_GetMemoryInfoEx` worked either way:

Yes, because this is the first field in the struct so the two are equivalent. But `version` is not the first field in `NV_MULTIGPU_CAPS` struct (neither V1 nor V2) so `NV_MULTIGPU_CAPS_V2 pMultiGPUCaps = { NV_MULTIGPU_CAPS_VER2 };` is actually equivalent to `NV_MULTIGPU_CAPS_V2 pMultiGPUCaps = { .multiGPUVersion = NV_MULTIGPU_CAPS_VER2, .version = 0, … };`.

Now I'm just mightly confused about what `version` is supposed to do.

SveSop:
I see what you mean. It does not seem like this is "working right" on this particular call, as this is the only versioned struct that i know of that do NOT have the .version field "in the correct spot" so to speak... or atleast "one of few".

1. `pMultiGPUCaps.MultiGPUVersion != MAKE_NVAPI_VERSION` (struct version), as it always **returns** "3" (for me atleast on RTX card non-SLI)
2. V1 - `pMultiGPUCaps.reserved` == V2 - `pMultiGPUCaps.reserved` == V2 - `pMultiGPUCaps.version` (due to "union" usage in V2) and is completely disregarded no matter what value it is set to from the tests i have done.

**Theory:** Some long long time ago, the V1 struct was distributed without the .version field, and as some kind of band-aid "fix", this union field was made in V2 as a possible solution to this missing .version field, and due to whatever problems someone decided to just scrap the idea of version checking and just provide data as-is.

Personally i am quite happy to write off this strangeness as a somewhat lingering bug that "just works" without a set version. It might explain why the `pMultiGPUCaps.NvLinkPresent` and `pMultiGPUCaps.fastNvLinkReads` fields does not really seem to be used for anything, although that might be debatable since i am not able to actually test this on a SLI setup.

SveSop:
Using V2 struct for `NvAPI_GPU_CudaEnumComputeCapableGpus` also does not work https://github.com/jp7677/dxvk-nvapi/commit/4966fb57b76d219f0fb85ae5caec405a34dc1114 even tho nvapi.h clearly states to use `//! \retval ::NVAPI_INCOMPATIBLE_STRUCT_VERSION  Structure version is not supported, initialize to NV_COMPUTE_GPU_TOPOLOGY_VER.`

`#define NV_COMPUTE_GPU_TOPOLOGY_VER  MAKE_NVAPI_VERSION(NV_COMPUTE_GPU_TOPOLOGY_V2,2)`

So.. sometimes there ARE bugs 😏

Saancreed:
I see, thanks for checking. Looks like doing

```cpp
pMultiGPUCaps->multiGPUVersion = 3;
pMultiGPUCaps->nTotalGPUs = nvapiAdapterRegistry->GetAdapterCount();
pMultiGPUCaps->nSLIGPUs = 0
pMultiGPUCaps->videoBridgePresent = 0;
```

and completely ignoring the version should be a reasonable no-empty-promises reimplementation of the function in question.

SveSop:
I do also get `nSLIGpus == nTotalGPUs` for me in windows, but as we cannot really test this with actual SLI setup i suppose it might not matter? Dunno..

Saancreed:
> I mean, lets say you have 3 adapters where 2 of them are "SLI capable", the result would be `nTotalGPUs = 3` and `nSLIGPUs = 2` maybe?

I would guess so but it might be safer to keep reporting `0` in dxvk-nvapi because we don't actually support SLI. Even in Vulkan device groups are kind of a meme feature.

SveSop:
> I would guess so but it might be safer to keep reporting `0` in dxvk-nvapi because we don't actually support SLI. Even in Vulkan device groups are kind of a meme feature.

Probably has no "real" meaning anyway. Just found it odd that both the 2070 card AND the 4070 reported `nSLIGPUs = 1`, since 40xx series do absolutely NOT support SLI/NVLink.

jp7677:
Testing on a Windows box here, this code 

<details>
  <summary>code</summary>

```c++
    std::cout << "NV_MULTIGPU_CAPS Ver2 " << std:: endl;
    NV_MULTIGPU_CAPS multiGPUCaps{};
    multiGPUCaps.version = NV_MULTIGPU_CAPS_VER2;
    multiGPUCaps.multiGPUVersion = 99;
    multiGPUCaps.nTotalGPUs = 98;
    multiGPUCaps.nSLIGPUs = 97;
    multiGPUCaps.videoBridgePresent = 96;
    multiGPUCaps.NvLinkPresent = 95;
    multiGPUCaps.fastNvLinkReads = 94;
    REQUIRE(nvAPI_D3D11_MultiGPU_GetCaps(&multiGPUCaps) == NVAPI_OK);
    std::cout << multiGPUCaps.version << std::endl;
    std::cout << multiGPUCaps.multiGPUVersion << std::endl;
    std::cout << multiGPUCaps.nTotalGPUs << std::endl;
    std::cout << multiGPUCaps.nSLIGPUs << std::endl;
    std::cout << multiGPUCaps.videoBridgePresent << std::endl;
    std::cout << multiGPUCaps.NvLinkPresent << std::endl;
    std::cout << multiGPUCaps.fastNvLinkReads << std::endl;

    std::cout << "NV_MULTIGPU_CAPS Ver1 " << std:: endl;
    NV_MULTIGPU_CAPS multiGPUCaps1{};
    multiGPUCaps1.version = NV_MULTIGPU_CAPS_VER1;
    multiGPUCaps1.multiGPUVersion = 89;
    multiGPUCaps1.nTotalGPUs = 88;
    multiGPUCaps1.nSLIGPUs = 87;
    multiGPUCaps1.videoBridgePresent = 86;
    multiGPUCaps1.NvLinkPresent = 85;
    multiGPUCaps1.fastNvLinkReads = 84;
    REQUIRE(nvAPI_D3D11_MultiGPU_GetCaps(&multiGPUCaps1) == NVAPI_OK);
    std::cout << multiGPUCaps1.version << std::endl;
    std::cout << multiGPUCaps1.multiGPUVersion << std::endl;
    std::cout << multiGPUCaps1.nTotalGPUs << std::endl;
    std::cout << multiGPUCaps1.nSLIGPUs << std::endl;
    std::cout << multiGPUCaps1.videoBridgePresent << std::endl;
    std::cout << multiGPUCaps1.NvLinkPresent << std::endl;
    std::cout << multiGPUCaps1.fastNvLinkReads << std::endl;

    std::cout << "NV_MULTIGPU_CAPS Ver0 " << std:: endl;
    NV_MULTIGPU_CAPS multiGPUCaps0{};
    multiGPUCaps0.version = 88;
    multiGPUCaps0.multiGPUVersion = 89;
    multiGPUCaps0.nTotalGPUs = 88;
    multiGPUCaps0.nSLIGPUs = 87;
    multiGPUCaps0.videoBridgePresent = 86;
    multiGPUCaps0.NvLinkPresent = 85;
    multiGPUCaps0.fastNvLinkReads = 84;
    REQUIRE(nvAPI_D3D11_MultiGPU_GetCaps(&multiGPUCaps0) == NVAPI_OK);
    std::cout << multiGPUCaps0.version << std::endl;
    std::cout << multiGPUCaps0.multiGPUVersion << std::endl;
    std::cout << multiGPUCaps0.nTotalGPUs << std::endl;
    std::cout << multiGPUCaps0.nSLIGPUs << std::endl;
    std::cout << multiGPUCaps0.videoBridgePresent << std::endl;
    std::cout << multiGPUCaps0.NvLinkPresent << std::endl;
    std::cout << multiGPUCaps0.fastNvLinkReads << std::endl;

    std::cout << "NV_MULTIGPU_CAPS Ver55" << std:: endl;
    NV_MULTIGPU_CAPS multiGPUCaps55{};
    multiGPUCaps55.version = 58;
    multiGPUCaps55.multiGPUVersion = 59;
    multiGPUCaps55.nTotalGPUs = 58;
    multiGPUCaps55.nSLIGPUs = 57;
    multiGPUCaps55.videoBridgePresent = 56;
    multiGPUCaps55.NvLinkPresent = 55;
    multiGPUCaps55.fastNvLinkReads = 54;
    REQUIRE(nvAPI_D3D11_MultiGPU_GetCaps(&multiGPUCaps55) == NVAPI_OK);
    std::cout << multiGPUCaps55.version << std::endl;
    std::cout << multiGPUCaps55.multiGPUVersion << std::endl;
    std::cout << multiGPUCaps55.nTotalGPUs << std::endl;
    std::cout << multiGPUCaps55.nSLIGPUs << std::endl;
    std::cout << multiGPUCaps55.videoBridgePresent << std::endl;
    std::cout << multiGPUCaps55.NvLinkPresent << std::endl;
    std::cout << multiGPUCaps55.fastNvLinkReads << std::endl;
```
</details>

shows:

<details>
  <summary>code</summary>

```
NV_MULTIGPU_CAPS Ver2
131100
3
1
1
0
0
0
NV_MULTIGPU_CAPS Ver1
65556
3
1
1
0
85
84
NV_MULTIGPU_CAPS Ver0
88
3
1
1
0
85
84
NV_MULTIGPU_CAPS Ver55
58
3
1
1
0
55
54
```
</details>

So it echos the version/reserved field and when `Ver2` is specified, also the last two fields are filled.

src/nvapi_d3d11.cpp
```diff
@@ -69,25 +69,27 @@ extern "C" {
...
-            case NV_MULTIGPU_CAPS_VER1: {
+        auto version = pMultiGPUCaps->version;
+        switch (version) { // NOLINT(*-multiway-paths-covered)
+            case NV_MULTIGPU_CAPS_VER2:
+                *pMultiGPUCaps = {};
+                pMultiGPUCaps->version = version;
+                pMultiGPUCaps->multiGPUVersion = 3; // Observed on Windows
+                // Report on purpose that no SLI GPU is available
+                pMultiGPUCaps->nTotalGPUs = nvapiAdapterRegistry->GetAdapterCount();
+                pMultiGPUCaps->nSLIGPUs = 0;
```
Saancreed, line 80: `*pMultiGPUCaps = {};` implicitly zeroes last two fields, but then `pMultiGPUCaps->nSLIGPUs = 0;` explicitly sets only one of them. Why should we handle this one differently?

Saancreed (approved):
Only one very nitpickish question, but LGTM anyway.

src/nvapi_d3d11.cpp
jp7677: > `*pMultiGPUCaps = {};` implicitly zeroes all (including last two) fields, but then `pMultiGPUCaps->nSLIGPUs = 0;` explicitly (and redundantly) sets another one. Why should we handle this one differently?

Yeah, it's completely redundant. I think the former me wanted to make it even more obvious that we report zero SLI GPUs when reading the source. Should turn this into a comment.

PR #208, opened 2024-08-26
Allow disabling entrypoints by name via environment variable

Saancreed:
Includes some C++ concepts fun, hopefully not too crazy but I wanted the string splitting helper to be fairly generic. Interestingly, I found it easier to do with regex-based separators than matching plain substrings or single characters.

Testing this automatically is a bit tricky because entrypoints are cached, and so is the processed variable value, and I'm kinda afraid it would interfere with other tests. Any ideas, or are you fine with no tests for this thing?

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ The following environment variables tweak DXVK-NVAPI's runtime behavior:
 
 - `DXVK_NVAPI_DRIVER_VERSION` lets you override the reported driver version. Valid values are numbers between 100 and 99999. Use e.g. `DXVK_NVAPI_DRIVER_VERSION=47141` to report driver version `471.41`.
 - `DXVK_NVAPI_ALLOW_OTHER_DRIVERS`, when set to `1`, allows DXVK-NVAPI to initialize for a non-NVIDIA GPU.
+- `DXVK_NVAPI_DISABLE_ENTRYPOINTS`, when set to a comma-separated list of NVAPI entrypoint names, will hide their implementation from the application. For example, `DXVK_NVAPI_DISABLE_ENTRYPOINTS=NvAPI_D3D11_BeginUAVOverlap,NvAPI_D3D11_EndUAVOverlap` will report D3D11 barrier control extensions as not available.
 - `DXVK_NVAPI_GPU_ARCH`, when set to one of supported NVIDIA GPU architecture IDs will override reported GPU architecture. Currently supported values are:
   - `GK100` (Kepler)
   - `GM200` (Maxwell)
--- a/src/nvapi_interface.cpp
+++ b/src/nvapi_interface.cpp
@@ -6,6 +6,9 @@
 extern "C" {
     using namespace dxvk;
 
+    static const auto disabledString = str::fromnullable(std::getenv("DXVK_NVAPI_DISABLE_ENTRYPOINTS"));
+    static const auto disabled = str::split<std::unordered_set<std::string_view>>(disabledString, std::regex(","));
+
     static std::unordered_map<NvU32, void*> registry;
 
     __declspec(dllexport) void* __cdecl nvapi_QueryInterface(NvU32 id) {
@@ -23,8 +26,15 @@ extern "C" {
             return registry.insert({id, nullptr}).first->second;
         }
 
+        auto name = std::string_view(it->func);
+
+        if (disabled.contains(name)) {
+            log::info(str::format("NvAPI_QueryInterface (", name, "): Disabled"));
+            return registry.insert({id, nullptr}).first->second;
+        }
+
 #define INSERT_AND_RETURN_WHEN_EQUALS(method) \
-    if (std::string(it->func) == #method)     \
+    if (name == #method)                      \
         return registry.insert({id, (void*)method}).first->second;
 
         // This block will be validated for completeness when running package-release.sh. Do not remove the comments.
@@ -149,7 +159,7 @@ extern "C" {
 
 #undef INSERT_AND_RETURN_WHEN_EQUALS
 
-        log::info(str::format("NvAPI_QueryInterface ", it->func, ": Not implemented method"));
+        log::info(str::format("NvAPI_QueryInterface ", name, ": Not implemented method"));
         return registry.insert({id, nullptr}).first->second;
     }
 }
--- a/src/nvapi_interface.cpp
+++ b/src/nvapi_interface.cpp
@@ -6,8 +6,38 @@
 extern "C" {
     using namespace dxvk;
 
-    static const auto disabledString = str::fromnullable(std::getenv("DXVK_NVAPI_DISABLE_ENTRYPOINTS"));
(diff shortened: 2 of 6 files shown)


src/nvapi_interface.cpp
```diff
@@ -6,6 +6,9 @@
 extern "C" {
     using namespace dxvk;
 
+    static const auto disabledString = str::fromnullable(std::getenv("DXVK_NVAPI_DISABLE_ENTRYPOINTS"));
```
jp7677, line 9: What about also converting to upper or lowercase and doing the same later during `contains` with `name` for a case insensitive check? I know that function names are case sensitive but may be we can assume that we won’t see two function names that only differ in casing in the headers.

src/util/util_string.h
```diff
@@ -38,4 +38,33 @@ namespace dxvk::str {
         append(stream, args...);
         return stream.str();
     }
+
+    std::string fromnullable(const char* str);
```
jp7677, line 42: Would be nice to add a few normal unit tests for the new string utils here https://github.com/jp7677/dxvk-nvapi/blob/master/tests/util.cpp#L37

jp7677 (commented):
Looks good! Added a suggestion. W.r.t. tests, agree that testing the full thing will become awkward, but we can test the new string magic separately. Catch should also provide nice matchers for this.

src/nvapi_interface.cpp
jp7677: I had this in mind: https://github.com/jp7677/dxvk-nvapi/blob/master/src/util/util_env.cpp#L190

Saancreed: Hm, okay, it's still fairly naive but should do the job.

But I'll probably wrap this into some kind of `Compare` for `std::set` to use internally.

jp7677: Thanks. Yeah, I agree we should avoid any complex string operations or new dependencies and don't dive into string rabbit holes.
My reasoning is the following: when I'm typing, I'm pretty good at making small typos and while indeed most function names will be copied, I would likely from your example just paste `NvAPI_D3D11_BeginUAVOverlap` twice and edit the second one to `..End..`. The user feedback of using this env variable here is only indirect because we never log the input, so  I'm ok with being a little bit forgiving by matching case-insensitive.. if it doesn't introduce madness.

PS: While at it, would you mind also improving the formatting at https://github.com/jp7677/dxvk-nvapi/blob/master/src/util/util_env.cpp#L190 like you did in `util::split`?

jp7677: Yes good idea, let's do this and using `std::call_once` sounds cool. I would probably do this after the first early-return of `nvapi_QueryInterface`.

Saancreed: Added, let me know what you think.

src/util/util_string.h
```diff
@@ -38,4 +38,76 @@ namespace dxvk::str {
...
+        auto last = items.end();
+
+        if (first == last)
+            return {};
+
+        auto init = *first++;
+
+        return std::accumulate(
+            first,
+            items.end(),
```
jp7677, line 107: `last`

src/nvapi_interface.cpp
```diff
@@ -6,13 +6,46 @@
...
+    static std::once_flag logDisabledOnceFlag;
+
     static std::unordered_map<NvU32, void*> registry;
 
     __declspec(dllexport) void* __cdecl nvapi_QueryInterface(NvU32 id) {
         auto entry = registry.find(id);
         if (entry != registry.end())
             return entry->second;
 
+        std::call_once(logDisabledOnceFlag, []() {
```
jp7677, line 22: May be extract the lambda into a separate function to not distract to much from the important things in `nvapi_QueryInterface`?

README.md
```diff
@@ -76,6 +76,7 @@ The following environment variables tweak DXVK-NVAPI's runtime behavior:
 
 - `DXVK_NVAPI_DRIVER_VERSION` lets you override the reported driver version. Valid values are numbers between 100 and 99999. Use e.g. `DXVK_NVAPI_DRIVER_VERSION=47141` to report driver version `471.41`.
 - `DXVK_NVAPI_ALLOW_OTHER_DRIVERS`, when set to `1`, allows DXVK-NVAPI to initialize for a non-NVIDIA GPU.
+- `DXVK_NVAPI_DISABLE_ENTRYPOINTS`, when set to a comma-separated list of NVAPI entrypoint names, will hide their implementation from the application. For example, `DXVK_NVAPI_DISABLE_ENTRYPOINTS=NvAPI_D3D11_BeginUAVOverlap,NvAPI_D3D11_EndUAVOverlap` will report D3D11 barrier control extensions as not implemented.
```
jp7677, line 79: `as not available` is probably better since it suggests otherwise that we return `NVAPI_NO_IMPLEMENTATION`.

tests/util.cpp
```diff
@@ -49,6 +49,52 @@ TEST_CASE("String", "[.util]") {
         dxvk::str::tonvss(ss, std::string("Longer-Than-Short-String-Longer-Than-Short-String-Longer-Than-Short-String"));
         REQUIRE_THAT(ss, SizeIs(64));
     }
+
+    SECTION("std::string") {
```
jp7677, line 53: `SECTION(“fromnullable”)`. Feel free to also updates the section name in the two function above.

src/nvapi_interface.cpp
jp7677: Looks cool, thanks.

tests/util.cpp
Saancreed: > Feel free to also updates the section name in the two functions above.

To match exactly the function name they test (so, `fromnvus` and `tonvss`)?

jp7677: Yep.

tests/util.cpp
```diff
@@ -49,6 +49,52 @@ TEST_CASE("String", "[.util]") {
...
+        REQUIRE(result.size() == 3);
+        REQUIRE(result.contains("foo"));
+        REQUIRE(result.contains("bar"));
+        REQUIRE(result.contains("baz"));
+    }
+
+    SECTION("CaseInsensitiveCompare") {
+        dxvk::str::CaseInsensitiveCompare<std::string_view> compare;
+
+        REQUIRE(compare("aa", "aa") == false);
```
jp7677, line 73: PS: you can shorten this and the following test case with `GENERATE` (https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_sysinfo.cpp#L201), but also ok to keep it like this.

src/util/util_string.h
Saancreed: Heh, I extracted `last` just to be able to reuse that iterator only to forget to actually do that :sweat_smile:

jp7677:
Thanks, there isn’t much left we cannot configure runtime now :)

PR #209, opened 2024-08-29
nvapi-gpu: Simplify check

SveSop:
No need to check struct version twice


jp7677:
Thanks, we are somewhat inconsistent with doing a pre-check or not regarding the struct version. Bailing out early safes some work and also plays easier with the tests but indeed duplicates the checks. Looks fine in this case!

PR #212, opened 2024-10-13
sysinfo: Detect Ampere on NVK

jp7677:
We can use the same check since Mesa NVK started advertising VK_KHR_fragment_shading_rate, which happened two days ago: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/31585 and esp. https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/31585/diffs?commit_id=2fb4aed9d8b51b741fde7bb6a9c8c263a7be5fdd


jp7677:
While at it, also simplify Volta detection which allows to remove quite a bit of code.

Saancreed (approved):
Looks fine, assuming we don't have to care about AMPERE_A (and I'm not actually sure which GPUs are AMPERE_A even).

jp7677:
> Looks fine, assuming we don't have to care about AMPERE_A (and I'm not actually sure which GPUs are AMPERE_A even).

Thanks. Worst case is that we fall back to Turing like we did earlier, so may be don't catch all Ampere's, but it is better than before. Since this matches the check for the proprietary driver I think we are not far off.

PR #213, opened 2024-10-17
Add implementation of nvofapi64.dll

liam-middlebrook:
Introduce a second library output to dxvk-nvapi for NVIDIA Optical Flow API. This serves as a pass-through implementation for D3D12 and Vulkan NVOF backends to `VK_NV_optical_flow`


src/nvofapi/nvofapi_vulkan.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include "util/util_string.h"
+
+#include "../inc/nvofapi/nvOpticalFlowVulkan.h"
+
+#include "nvofapi.h"
+
+namespace nvofapi {
+
+    bool NvOFInstanceVk::Initialize() {
+        m_library = LoadLibraryA("winevulkan.dll");
```
Saancreed, line 34: While for D3D12 it does make sense to load VK funcs from winevulkan directly (because vkd3d-proton does the same thing), doesn't doing the same for native Vulkan OFA risk a mismatch if there's native Vulkan loader installed in the Wine/Proton prefix? Will funcs exported by `winevulkan.dll` be able to handle VK handles that were created using functions exported by `vulkan-1.dll` if that's the library the application loaded an is using to access Vulkan API?

robotman40:
On an unrelated note, and with all due respect, may I ask what causes features like this to take awhile to arrive?

jp7677:
Thanks for the PR @liam-middlebrook !

Looking at the change, `nvofapi` is added on the sideline. What about making it more first-class citizen? I've pulled the changes locally and played with it, resulting in: https://github.com/jp7677/dxvk-nvapi/tree/of-playground/src (it still compiles, but not tested)
- `nvofapi` entrypoints are now directly in `src` next to the `nvapi` entrypoints. I've also separated those into `nvofapi`, `nvofapi_d3d12` and `nvofapi_vulkan` (using forward declaration for the function pointers) to match the `nvapi` entrypoints.
- `NvofapiD3d12Instance` lives now next to `NvapiD3d12Device` in `d3d12` and `NvofapiVulkanInstance` in a new directory `vulkan`. I've left the base class and `NvOFImageVk` in `nvofapi`, but the more I think of it, the existing `shared` directory is propably a better place. PS: I've replaced the `instance` member in `NvOFImageVk` with `VkDevice m_vkDevice` to avoid the circular dependency. 
- The `nvofapi` classes are in the `nvofapi` namespace. I would be fine to just use `dxvk` namespace like we do with the `nvapi` classes. 

Some additional thoughts:
-  It would be nice if we can use the same logging and return code approach for both `nvapi` and `nvofapi`. I've tried this in https://github.com/jp7677/dxvk-nvapi/blob/of-playground/src/nvofapi.cpp . Separating the statuscodes by namespace (https://github.com/jp7677/dxvk-nvapi/blob/of-playground/src/util/util_statuscode.h#L157) and using `using` (https://github.com/jp7677/dxvk-nvapi/blob/of-playground/src/nvapi.cpp#L14) to avoid to much nesting works imho OK. Having the entry point tracing and log statements in the status code functions also automatically replaces some of the log statements.
- Most functions in the `nvofapi` classes return `NV_OF_STATUS`, but in practice it is just success or error or even void (when only success is returned). What about just using true/false/void to be in line with most `nvapi` classes? We can revisit this later for all classes, because true/false is indeed much less clear to read.
- I guess it is safe to assume that we don't need `nvofapi` for 32bits? I've guarded that in `meson.build`.
- There are a lot of compile warning, would be nice to have those resolved. Also clang-tidy (in CLion) offers a lot of simple suggestions, would be nice to apply those.

Let me know what you think of this slightly alternative approach/structure for combining `nvapi`/`nvofapi`.  Putting all `nvapi` entrypoints and classes into a new `nvapi` directory (after this PR) and thus only keeping `nvapi`, `nvofapi`, `util`, `dxvk` and `vkd3d-proton` top-level is also a way forward. I'm not sure yet which one is better.

@Saancreed , also of course curious about your thoughts.

Saancreed:
I don't have any strong opinions about how the repo/project should be structured (providing two libraries now is a bit tricky anyway) but for the remaining points:

> * It would be nice if we can use the same logging and return code approach for both `nvapi` and `nvofapi`. I've tried this in https://github.com/jp7677/dxvk-nvapi/blob/of-playground/src/nvofapi.cpp . Separating the statuscodes by namespace (https://github.com/jp7677/dxvk-nvapi/blob/of-playground/src/util/util_statuscode.h#L157) and using `using` (https://github.com/jp7677/dxvk-nvapi/blob/of-playground/src/nvapi.cpp#L14) to avoid to much nesting works imho OK. Having the entry point tracing and log statements in the status code functions also automatically replaces some of the log statements.

:+1: to this.

> * Most functions in the `nvofapi` classes return `NV_OF_STATUS`, but in practice it is just success or error or even void (when only success is returned). What about just using true/false/void to be in line with most `nvapi` classes? We can revisit this later for all classes, because true/false is indeed much less clear to read.

Tbh I'd _slightly_ prefer just using status types everywhere. Nothing wrong with having a little wider return type that what we currently need (because it greatly simplifies adding another code paths with possible failures in them) and if it matches the public API there's less work to do in the outermost layer.

> * I guess it is safe to assume that we don't need `nvofapi` for 32bits? I've guarded that in `meson.build`.

As far as I can tell:

* NVIDIA does not seem to report `VK_NV_optical_flow` as supported for Vulkan physical devices given to 32-bit applications on Linux.
* On Linux, there is no 32-bit `libnvidia-opticalflow.so`.
* On Windows, NVIDIA does ship 32-bit `nvofapi.dll` as part of their driver, but I don't know how usable and feature-complete it is.
  * The only entrypoints it exports are `NvOFAPICreateInstanceCuda`, `NvOFAPICreateInstanceD3D11`, `NvOFAPICreateInstanceD3D12` and `NvOFGetMaxSupportedApiVersion` as opposed to `nvofapi64.dll` which also exports `NvOFAPICreateInstanceVk`.
* Streamline (which is currently the only public SDK that enables DLFG integrations) requires 64-bit.

However, once the WoW64 work in Wine completes, it's might be possible to leverage 64-bit Optical Flow in Linux drivers to implement/expose 32-bit Optical Flow in Wine. That said, I'm not sure if there are any 32-bit Windows applications that use NVOF out there so I doubt there's any value in doing this.

> * There are a lot of compile warning, would be nice to have those resolved. Also clang-tidy (in CLion) offers a lot of simple suggestions, would be nice to apply those.

:+1:

> * Have to think about the vulkan dependency i.c.w. testing. For `nvapi` we had wrapped those in https://github.com/jp7677/dxvk-nvapi/blob/master/src/sysinfo/vulkan.h and have it created by https://github.com/jp7677/dxvk-nvapi/blob/master/src/resource_factory.cpp#L21 and being replaced by a mock for tests. Something similar is also needed for `nvofapi`. But this is for later.

We will also probably have to mock Vulkan functions for Vulkan OF which could be a bit harder.

jp7677:
> > * Most functions in the `nvofapi` classes return `NV_OF_STATUS`, but in practice it is just success or error or even void (when only success is returned). What about just using true/false/void to be in line with most `nvapi` classes? We can revisit this later for all classes, because true/false is indeed much less clear to read.
> 
> Tbh I'd _slightly_ prefer just using status types everywhere. Nothing wrong with having a little wider return type that what we currently need (because it greatly simplifies adding another code paths with possible failures in them) and if it matches the public API there's less work to do in the outermost layer.

Yes, this is what I wanted to say with my last sentence. With hindsight we/I should have used wider return types in the inner layers/classes from the beginning on. I guess I couldn't decide at that time between COM and NVAPI and just went for bool ;). Not sure yet if using `NvAPI_Status` or `NV_OF_STATUS` in the inner layers plays out. But on the other hands, considering our small code base, we can change this easily later.

> > * Have to think about the vulkan dependency i.c.w. testing. For `nvapi` we had wrapped those in https://github.com/jp7677/dxvk-nvapi/blob/master/src/sysinfo/vulkan.h and have it created by https://github.com/jp7677/dxvk-nvapi/blob/master/src/resource_factory.cpp#L21 and being replaced by a mock for tests. Something similar is also needed for `nvofapi`. But this is for later.
> 
> We will also probably have to mock Vulkan functions for Vulkan OF which could be a bit harder.

Should be doable but indeed requires some refactorings/extraction and hopefully reuse of our existing wrapper.

liam-middlebrook:
I've addressed some of the feedback above. I think the big outstanding items to address are:
- limiting builds to 64-bit targets only
- figuring out what to do about tests and if we want to defer that as discussed over chat
- merging nvofapi functionality into the dxvk namespace

Let me know if there's anything else that comes to mind.

src/nvofapi/nvofapi_common.cpp
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        // Need to get the size/id for the private data to pass it along to VK...
+        VkOpticalFlowSessionCreatePrivateDataInfoNV privData{};
+        privData.sType = VK_STRUCTURE_TYPE_OPTICAL_FLOW_SESSION_CREATE_PRIVATE_DATA_INFO_NV;
+        privData.size = ((NV_OF_PRIV_DATA*)initParams->hPrivData)->size;
+        privData.id = ((NV_OF_PRIV_DATA*)initParams->hPrivData)->id;
+        privData.pPrivateData = ((NV_OF_PRIV_DATA*)initParams->hPrivData)->data;
+
+        createInfo.pNext = &privData;
```
Saancreed, line 128: The comment on `hPrivData` states:

> Optical flow private data. It is reserved field and should be set to NULL.

Therefore I believe we should be checking `initParams->hPrivData` for `nullptr` before attempting to pass along this private data.

Same applies to `inParams->hPrivData` in `RecordCmdBuf` below.

src/d3d12/nvofapi_d3d12_instance.cpp
```diff
@@ -0,0 +1,176 @@
...
+        }
+        return true;
+    }
+
+    void NvOFInstanceD3D12::RegisterBuffer(const NV_OF_REGISTER_RESOURCE_PARAMS_D3D12* registerParams) {
+        NV_OF_REGISTER_RESOURCE_PARAMS_VK vkParams{};
+        dxvk::log::info(
+            dxvk::str::format("RegisterBuffer DX: resource: ",
+                registerParams->resource, " inputFencePoint: ",
+                registerParams->inputFencePoint.fence, " ouputFencePoint: ",
```
Saancreed, line 111: Typo:

```suggestion
                registerParams->inputFencePoint.fence, " outputFencePoint: ",
```

src/nvofapi/nvofapi_common.cpp
```diff
@@ -0,0 +1,218 @@
...
+        NvOFGPUBufferHandle reference[6];
+        uint8_t reserved[100];
+    } NV_OF_EXECUTE_PRIV_DATA_INPUT_MIPS;
+
+    constexpr uint32_t NV_OF_EXECUTE_PRIV_DATA_ID_INPUT_MIPS = 6;
+
+    uint32_t NvOFInstance::GetVkOFAQueue() {
+        uint32_t count = 0;
+        m_vkGetPhysicalDeviceQueueFamilyProperties(m_vkPhysicalDevice, &count, nullptr);
+        VkQueueFamilyProperties* queueFamProps = (VkQueueFamilyProperties*)calloc(sizeof(VkQueueFamilyProperties), count);
```
Saancreed, line 53: I'll leave this to Jens but I believe using `std::vector` is preferable over manual `malloc`/`calloc` and `free`. It also lets us ignore the mythical allocation failures without causing segfaults.

src/nvofapi/nvofapi_common.cpp
```diff
@@ -0,0 +1,218 @@
...
+        auto ret = m_vkCreateOpticalFlowSessionNV(m_vkDevice, &createInfo, NULL, &m_vkOfaSession);
+
+        if (ret == VK_SUCCESS) {
+            return Success();
+        }
+
+        return ErrorGeneric();
+    }
+
+    NV_OF_STATUS NvOFInstance::BindImageToSession(NvOFImage* image, VkOpticalFlowSessionBindingPointNV bindingPoint) {
```
Saancreed, line 139: Looks like at each callsite there is a `reinterpret_cast<NvOFImage*>` which causes a lot of visual noise, maybe this function could instead accept `NvOFGPUBufferHandle` and do the cast on its own?

src/nvofapi/nvofapi_common.cpp
```diff
@@ -0,0 +1,218 @@
...
+        VkImageLayout layout = VK_IMAGE_LAYOUT_GENERAL;
```
Saancreed, line 140: Nitpicking, but if this is going to always be `VK_IMAGE_LAYOUT_GENERAL` then it might as well be inlined in the call below.

src/nvofapi/nvofapi_common.cpp
```diff
@@ -0,0 +1,218 @@
...
+        BindImageToSession(reinterpret_cast<NvOFImage*>(inParams->inputFrame), VK_OPTICAL_FLOW_SESSION_BINDING_POINT_INPUT_NV);
+        BindImageToSession(reinterpret_cast<NvOFImage*>(inParams->referenceFrame), VK_OPTICAL_FLOW_SESSION_BINDING_POINT_REFERENCE_NV);
+        BindImageToSession(reinterpret_cast<NvOFImage*>(outParams->outputBuffer), VK_OPTICAL_FLOW_SESSION_BINDING_POINT_FLOW_VECTOR_NV);
+        BindImageToSession(reinterpret_cast<NvOFImage*>(outParams->outputCostBuffer), VK_OPTICAL_FLOW_SESSION_BINDING_POINT_COST_NV);
+        BindImageToSession(reinterpret_cast<NvOFImage*>(outParams->bwdOutputBuffer), VK_OPTICAL_FLOW_SESSION_BINDING_POINT_BACKWARD_FLOW_VECTOR_NV);
+        BindImageToSession(reinterpret_cast<NvOFImage*>(outParams->bwdOutputCostBuffer), VK_OPTICAL_FLOW_SESSION_BINDING_POINT_BACKWARD_COST_NV);
+        BindImageToSession(reinterpret_cast<NvOFImage*>(outParams->globalFlowBuffer), VK_OPTICAL_FLOW_SESSION_BINDING_POINT_GLOBAL_FLOW_NV);
+        // Support INPUT_MIPS execute priv data
+        if (((NV_OF_PRIV_DATA*)inParams->hPrivData)->id == NV_OF_EXECUTE_PRIV_DATA_ID_INPUT_MIPS) {
```
Saancreed, line 184: Besides the missing null check for `hPrivData`, is there no need to also check the size of what `data` points to?

src/vulkan/nvofapi_vulkan_instance.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include "../inc/nvofapi/nvOpticalFlowVulkan.h"
+
+#include "vulkan/nvofapi_vulkan_instance.h"
+
+namespace nvofapi {
+
+    bool NvOFInstanceVk::Initialize() {
+        // For VK we cannot load winevulkan directly or we may break handle
+        // opacity.
+        m_library = LoadLibraryA("vulkan-1.dll");
```
Saancreed, line 36: One more thing about this: I just remembered that Streamline also exports its own VK funcs from `sl.interposer.dll`. This sounds like another possible problem but I have no idea how to detect if some game is using Streamline exports or if it's calling Vulkan loader directly.

jp7677 (commented):
Some styling nits and a few suggestion. Thanks a lot!

jp7677:
> I've addressed some of the feedback above. I think the big outstanding items to address are:

Thanks a lot, yes, and points 1 and 3 are easy to address regardless the direction.

>  limiting builds to 64-bit targets only

No strong preference, if there is no practical benefit to build a 32bit target, I would leave that out. But I don't mind building it either.     
 
> figuring out what to do about tests and if we want to defer that as discussed over chat

I'm fine with leaving tests out for now. 
 
> merging nvofapi functionality into the dxvk namespace

No strong preference. Using namespace `dxvk` would make it consistent with the `nvapi` classes. We could then defer a stricter and clearer separation between `nvapi` and `nvofapi` on both directory and namespaces for later.  
 
> 
> Let me know if there's anything else that comes to mind.

See the comments above. Mostly smaller things.

src/nvofapi/nvofapi_common.cpp
liam-middlebrook: I don't think that's necessary, the size can be inferred by the type.

src/nvofapi/nvofapi_common.cpp
```diff
@@ -0,0 +1,218 @@
...
+            bindingPoint,
+            image->ImageView(),
+            layout);
+        if (ret != VK_SUCCESS) {
+            return ErrorGeneric();
+        }
+        return Success();
+    }
+
+    NV_OF_STATUS NvOFInstance::getCaps(NV_OF_CAPS param, uint32_t* capsVal, uint32_t* size) {
```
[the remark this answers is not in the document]
liam-middlebrook: I suppose it technically could right now, but eventually a full implementation of it would make use of `m_vkPhysicalDevice` to query information about the optical flow capabilities of the desired GPU. I'm not sure that adding the static keyword now provides much value, although neither does the current (mostly-stub) implementation.

src/nvofapi_d3d12.cpp
```diff
@@ -0,0 +1,151 @@
...
+            log::trace(n, log::fmt::ptr(pD3D12Device), log::fmt::ptr(hOFInstance));
+
+        nvofapi::NvOFInstanceD3D12* nvOF = nullptr;
+        try {
+            nvOF = new nvofapi::NvOFInstanceD3D12(pD3D12Device);
+        } catch (std::exception const& e) {
+            log::info(str::format("CreateOpticalFlowD3D12 exception, %s", e.what()));
+        }
+
+        if (!nvOF) {
```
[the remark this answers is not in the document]
liam-middlebrook: I'm not sure that I agree given that if the constructor throws an exception (either OOM, or otherwise) the assignment shouldn't complete and the value should remain `nullptr`. Either way I re-worked this a bit into a way that I think will mesh a bit better.

src/vulkan/nvofapi_vulkan_instance.cpp
liam-middlebrook: That's used for Streamline's "implicit hooking". Your intuition that we should be loading `vulkan-1.dll` is correct because that's what other middleware would be doing (presuming it is not passed `vkGetInstanceProcAddr()` directly by the application).

Saancreed (approved):
Only some minor nitpicks and suggestions but nothing too important so I'm going to approve anyway.

src/d3d12/nvofapi_d3d12_instance.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+        VkCommandBuffer vkCmdBuf;
+        m_device->BeginVkCommandBufferInterop(m_cmdLists[m_cmdListIndex], &vkCmdBuf);
+
+        this->RecordCmdBuf(&vkInputParams, &vkOutputParams, vkCmdBuf);
+
+        m_device->EndVkCommandBufferInterop(m_cmdLists[m_cmdListIndex]);
+        m_cmdLists[m_cmdListIndex]->Close();
+
+        m_commandQueue->ExecuteCommandLists(1, (ID3D12CommandList**)&m_cmdLists[m_cmdListIndex]);
```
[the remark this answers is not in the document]
liam-middlebrook: Yes, because it's a `ID3D12GraphicsCommandList` vs `ID3D12CommandList` mismatch. I'll change it to a c++ style cast though.

jp7677 (approved):

PR #215, opened 2024-10-21
Proton CI

Saancreed:
I just used the `latest` tag for the image, this way we will have to bump this only when Proton switches the Steam Runtime it uses to something other than `sniper`. Seems to work, but…

```
fatal: detected dubious ownership in repository at '/__w/dxvk-nvapi/dxvk-nvapi'
To add an exception for this directory, call:
	git config --global --add safe.directory /__w/dxvk-nvapi/dxvk-nvapi
```

Let me know if this is something we need to care about, for now the fallback appears to work well enough.

### DIFF
--- a/.github/workflows/build-proton.yml
+++ b/.github/workflows/build-proton.yml
@@ -1,0 +1,26 @@
+name: Build with Proton SDK
+
+on:
+  push:
+    branches: [ master ]
+    tags: [ v* ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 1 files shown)


jp7677:
ah nice, thanks! I still had this somewhere in the back of my head…

W.r.t the git error, let’s try this https://github.com/actions/runner/issues/2033#issuecomment-1598547465 but I’m fine with the fallback if more is needed.

Saancreed:
Yup, no such error now, nice.

jp7677 (approved):

PR #216, opened 2024-10-24
LLVM builds (against Windows SDK)

Saancreed:
Wasn't as terrible as I expected but I still wasted a lot of time fighting Meson's nonsense (for example, when I tried choosing the linker using more idiomatic `cpp_ld = 'lld-link'` it insisted on passing `-fuse-ld=lld-link` _to the linker_, which at this point `clang-cl` already determined it to be default `link.exe`, which of course doesn't exist).

TODO: ~~Persist the toolchain flavor in logged version string?~~ Perhaps ~~build and~~ run tests as well? But that's something for another day I think.

### DIFF
--- a/.github/workflows/build-llvm.yml
+++ b/.github/workflows/build-llvm.yml
@@ -1,0 +1,72 @@
+name: Build with LLVM against MSVCRT and Windows SDK
+
+on:
+  push:
+    branches: [ master ]
+    tags: [ v* ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  build-llvm:
+    runs-on: ubuntu-latest
+    container: docker.io/archlinux/archlinux:base
+
+    steps:
+    - name: Install packages
+      run: pacman -Syu --needed --noconfirm clang curl git jq lld llvm meson ninja
+
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        submodules: recursive
+        fetch-depth: 0
+
+    - name: Suppress repository ownership mismatch errors
+      run: git config --global --add safe.directory '*'
+
+    - name: Prepare MSVCRT and Windows SDK
+      run: |
+        xwin_version=$(curl -s https://api.github.com/repos/Jake-Shadle/xwin/releases/latest | jq -r .tag_name)
+        curl -sL "https://github.com/Jake-Shadle/xwin/releases/download/${xwin_version}/xwin-${xwin_version}-x86_64-unknown-linux-musl.tar.gz" | bsdtar -xf-
+        "./xwin-${xwin_version}-x86_64-unknown-linux-musl/xwin" \
+          --accept-license \
+          --arch x86_64,x86 \
+          splat \
+          --preserve-ms-arch-notation \
+          --use-winsysroot-style \
+          --output ./winsysroot
+        cd './winsysroot/Windows Kits/10'
+        test -e ./Include || mv ./include ./Include
+        test -e ./Lib || mv ./lib ./Lib
+
+    - name: Build x86
+      shell: bash
+      run: |
+        meson setup build32 . \
+          --cross-file ./build-win32-llvm.txt \
+          --buildtype release \
+          --prefix ~+/install \
+          --bindir bin \
+          --libdir lib \
+          --strip
+        ninja -C build32 install
+
+    - name: Build x64
+      shell: bash
+      run: |
+        meson setup build64 . \
+          --cross-file ./build-win64-llvm.txt \
+          --buildtype release \
+          --prefix ~+/install \
+          --bindir bin \
+          --libdir lib \
+          --strip -Denable_tests=true
+        ninja -C build64 install
+
+    - name: Upload artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: dxvk-nvapi-llvm
+        path: install/bin/*
--- a/build-win32-llvm.txt
+++ b/build-win32-llvm.txt
@@ -1,0 +1,20 @@
+[constants]
+winsysroot = '@GLOBAL_SOURCE_ROOT@/winsysroot'
+
+[binaries]
+cpp = ['clang-cl', '/clang:--target=i686-pc-windows-msvc', '/clang:-fuse-ld=lld-link', '/winsysroot', winsysroot]
+strip = 'llvm-strip'
+ar = 'llvm-ar'
+
+[built-in options]
+cpp_args = ['/MT', '-Wno-c++11-narrowing']
+cpp_link_args = ['/winsysroot:' + winsysroot]
+
+[properties]
+needs_exe_wrapper = true
+
+[host_machine]
+system = 'windows'
+cpu_family = 'x86'
+cpu = 'x86'
+endian = 'little'
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 build
 buildDir
 out
+winsysroot
 .idea/**
 .vscode/**
 version.h
--- a/build-win64-llvm.txt
+++ b/build-win64-llvm.txt
@@ -1,0 +1,20 @@
+[constants]
+winsysroot = '@GLOBAL_SOURCE_ROOT@/winsysroot'
+
+[binaries]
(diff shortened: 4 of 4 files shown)


jp7677:
> TODO: Persist the toolchain flavor in logged version string?

I created a start for that in https://github.com/jp7677/dxvk-nvapi/pull/217

build-win32-llvm.txt
```diff
@@ -0,0 +1,20 @@
+[constants]
+winsysroot = '@GLOBAL_SOURCE_ROOT@/winsysroot'
+
+[binaries]
+cpp = ['clang-cl', '/clang:--target=i686-pc-windows-msvc', '/clang:-fuse-ld=lld-link', '/winsysroot', winsysroot]
+strip = 'llvm-strip'
+ar = 'llvm-ar'
+
+[built-in options]
+cpp_args = ['/MT', '-Wno-c++11-narrowing']
```
Saancreed, line 10: If you're wondering why is this `-Wno-c++11-narrowing` needed, it's to suppress this error:

```
FAILED: src/nvapi64.dll.p/nvapi_drs.cpp.obj 
clang-cl /clang:-fuse-ld=lld-link /winsysroot …/dxvk-nvapi/winsysroot -Isrc/nvapi64.dll.p -Isrc -I../src -I../external/nvapi -I../external/Vulkan-Headers/include -I. /MD /nologo /showIncludes /utf-8 /W3 /EHsc /std:c++20 /permissive- /O2 /Gw -DNOMINMAX -DWIN32_LEAN_AND_MEAN -D_CRT_SECURE_NO_WARNINGS /MT /Fdsrc/nvapi64.dll.p/nvapi_drs.cpp.pdb /Fosrc/nvapi64.dll.p/nvapi_drs.cpp.obj /c ../src/nvapi_drs.cpp
In file included from ../src/nvapi_drs.cpp:2:
../external/nvapi/NvApiDriverSettings.c(932,132): error: constant expression evaluates to -1 which cannot be narrowed to type 'NvU32' (aka 'unsigned long') [-Wc++11-narrowing]
  932 |     {SET_VAB_DATA_ID,                      SET_VAB_DATA_STRING,                       6, (NvU32 *)g_valuesSET_VAB_DATA,            SET_VAB_DATA_USE_API_DEFAULTS},
      |                                                                                                                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../external/nvapi/NvApiDriverSettings.c(932,132): note: insert an explicit cast to silence this issue
  932 |     {SET_VAB_DATA_ID,                      SET_VAB_DATA_STRING,                       6, (NvU32 *)g_valuesSET_VAB_DATA,            SET_VAB_DATA_USE_API_DEFAULTS},
      |                                                                                                                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                                                                    static_cast<NvU32>(          )
../external/nvapi/NvApiDriverSettings.c(934,132): error: constant expression evaluates to -1769598857 which cannot be narrowed to type 'NvU32' (aka 'unsigned long') [-Wc++11-narrowing]
  934 |     {VSYNCTEARCONTROL_ID,                  VSYNCTEARCONTROL_STRING,                   2, (NvU32 *)g_valuesVSYNCTEARCONTROL,        VSYNCTEARCONTROL_DISABLE},
      |                                                                                                                                    ^~~~~~~~~~~~~~~~~~~~~~~~
../external/nvapi/NvApiDriverSettings.c(934,132): note: insert an explicit cast to silence this issue
  934 |     {VSYNCTEARCONTROL_ID,                  VSYNCTEARCONTROL_STRING,                   2, (NvU32 *)g_valuesVSYNCTEARCONTROL,        VSYNCTEARCONTROL_DISABLE},
      |                                                                                                                                    ^~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                                                                    static_cast<NvU32>(     )
2 errors generated.
```

But because the error comes from NVAPI headers which we include from a submodule, I thought it would be preferable to suppress this rather than try fixing them.

Saancreed:
Seems to work, and reports itself like this:

> `DXVK-NVAPI v0.7.1-62-g64147d9 clang-cl 18.1.8 x86_64 release`

or

> `DXVK-NVAPI v0.7.1-62-g64147d9 clang-cl 18.1.8 x86 release`

.github/workflows/build-llvm.yml
```diff
@@ -0,0 +1,72 @@
...
+      with:
+        submodules: recursive
+        fetch-depth: 0
+
+    - name: Suppress repository ownership mismatch errors
+      run: git config --global --add safe.directory '*'
+
+    - name: Prepare MSVCRT and Windows SDK
+      run: |
+        xwin_version=0.6.6-rc.2
```
jp7677, line 31: We could use something like this to always fetch the latest version since they seem to release quite often.

```
curl -s https://api.github.com/repos/Jake-Shadle/xwin/releases/latest | jq  -r '.tag_name'
```

.github/workflows/build-llvm.yml
```diff
@@ -0,0 +1,72 @@
...
+          --arch x86_64,x86 \
+          splat \
+          --preserve-ms-arch-notation \
+          --use-winsysroot-style \
+          --output ./winsysroot
+        cd './winsysroot/Windows Kits/10'
+        mv include Include
+        mv lib Lib
+
+    - name: Build x64
```
jp7677, line 44: nit: other workflows build the x86 version first.

.github/workflows/build-llvm.yml
```diff
@@ -0,0 +1,72 @@
...
+          --bindir bin \
+          --libdir lib \
+          --strip
+        ninja -C build32 install
+
+    - name: Upload artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: dxvk-nvapi-llvm
+        path: install/bin/*
```
jp7677, line 72: Similar with msvc artifacts I would prefer to include the tests executable, especially since they are not (yet) executed here as part of CI.

build-win32-llvm.txt
jp7677: Thanks for the clarification. I think this is fine here. Alternatively, we might wrap https://github.com/jp7677/dxvk-nvapi/blob/master/src/nvapi_drs.cpp#L2 like we do here https://github.com/jp7677/dxvk-nvapi/blob/master/src/nvapi_private.h#L42, but for clang. But I guess that's not worth it, even when moving that include to `nvapi_private.h` to keep the cpp files clean of this stuff (assuming that this doesn't create other issues).

jp7677 (commented):
Looks good. Not yet sure about the extra maintenance in the long term, but it's nicely isolated, so cool to include this workflow. Not much CI variations left after this PR ;).

See some remarks.

.github/workflows/build-llvm.yml
```diff
@@ -0,0 +1,72 @@
...
+      run: pacman -Syu --needed --noconfirm clang curl git lld llvm meson ninja
+
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        submodules: recursive
+        fetch-depth: 0
+
+    - name: Suppress repository ownership mismatch errors
+      run: git config --global --add safe.directory '*'
```
jp7677, line 27: ah indeed :(.

.github/workflows/build-llvm.yml
Saancreed: It is already included :)

Check the artifact from https://github.com/jp7677/dxvk-nvapi/actions/runs/11532212308, it contains `nvapi.dll`, `nvapi64.dll` and `nvapi64-tests.exe`.

jp7677: ah perfect! Sry for the spam ;)

.github/workflows/build-llvm.yml
Saancreed: On one hand, autoupdating this would be nice, but on another we kinda have no guarantees of stable behavior.

But as least I'll know when another release that fixes casing of `include` and `lib` is made because my hack below will likely break :sweat_smile:

Saancreed: Done, hopefully in a way that shouldn't break in the future…

jp7677: Yeah, there is no perfect solution. If it breaks on annoying moment, we can still pin to a specific version in the future.

jp7677 (approved):

PR #219, opened 2024-11-01
Revert "util: Spoof Ada and later GPUs as Ampere for Ghost of Tsushima"

Saancreed:
This reverts commit f86f56b24737d117098a80133516edfb553fe39a.

To be merged after DLSS FG support is in place. Let me know if I should actually write a new commit message.

### DIFF
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -117,19 +117,10 @@ namespace dxvk::env {
         }
     }
 
-    bool isGhostOfTsushima() {
-        return getExecutableName() == std::string("GhostOfTsushima.exe");
-    }
-
     bool needsAmpereSpoofing(NV_GPU_ARCHITECTURE_ID architectureId, void* pReturnAddress) {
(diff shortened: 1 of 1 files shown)


jp7677:
Thanks, PR is good ofc. Would be nice if the commit message at least briefly states why we don't need this any longer.

jp7677 (approved):

shelterx:
The workaround was made for a reason, are we sure it isn’t needed?

https://github.com/HansKristian-Work/vkd3d-proton/issues/2008

Saancreed:
@shelterx The cause of this issue will be gone from Proton soon so yes, we are… fairly confident about it.

But feel free to verify this and report back; while I can no longer notice the problem myself, another confirmation would be appreciated. Just don't forget to add Liam's D3DKMT patches to the mix and install nvofapi64.dll into Proton prefix.

shelterx:
@Saancreed seems to work yes from what I could tell, with DLSS FG and Reflex On

Saancreed:
That's great, thanks for checking!

PR #220, opened 2024-11-05
Slightly reorganize includes

jp7677:
Introduce `nvofapi_private.h`.
Move all external COM interface to new `interfaces` folder.
Use relative includes. This makes adding tests easier.

Also align a few smaller things, e.g. log statement for `initilize`.


src/nvofapi_private.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+#include "./nvapi_private.h"
+
+#include "../inc/nvofapi/nvOpticalFlowD3D12.h"
+#include "../inc/nvofapi/nvOpticalFlowVulkan.h"
```
Saancreed, line 6: Do we care about trailing newlines?

Saancreed (approved):

PR #221, opened 2024-11-06
nvapi,nvofapi: Reorganize file layout

jp7677:
Move things to `nvapi` or `nvofapi` folder.


src/meson.build
```diff
@@ -2,17 +2,17 @@ nvapi_src = files([
...
-  'd3d/lfx.cpp',
-  'd3d/nvapi_d3d_instance.cpp',
-  'd3d/nvapi_d3d_low_latency_device.cpp',
-  'd3d11/nvapi_d3d11_device.cpp',
-  'd3d12/nvapi_d3d12_device.cpp',
+  'nvapi/lfx.cpp',
+  'nvapi/nvapi_d3d_instance.cpp',
+  'nvapi/nvapi_d3d_low_latency_device.cpp',
+  'nvapi/nvapi_d3d11_device.cpp',
+  'nvapi/nvapi_d3d12_device.cpp',
```
Saancreed, line 15: Perhaps the stuff below could also be moved to `nvapi/` (and `nvofapi{,_d3d12,_vulkan},cpp` to `nvofapi/`)? Then we could have each library be built from sources only in its own directory plus some shared utils/interfaces and this `meson.build` would just define variables for shared sources and then do `subdir('nvapi')` and `subdir('nvofapi')`.

src/meson.build
```diff
@@ -2,17 +2,17 @@ nvapi_src = files([
...
-  'sysinfo/nvml.cpp',
-  'sysinfo/nvapi_output.cpp',
-  'sysinfo/nvapi_adapter.cpp',
-  'sysinfo/nvapi_adapter_registry.cpp',
+  'nvapi/vulkan.cpp',
+  'nvapi/nvml.cpp',
+  'nvapi/nvapi_output.cpp',
+  'nvapi/nvapi_adapter.cpp',
+  'nvapi/nvapi_adapter_registry.cpp',
   'resource_factory.cpp',
```
Saancreed, line 10: Do you intend to share this between nvapi and nvofapi? Maybe it could be moved to `util` if that's the case (and to `nvapi/` otherwise, or just renamed to `nvapi_resource_factory.cpp`).

Saancreed (approved):
I have some random suggestions…

src/meson.build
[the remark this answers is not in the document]
jp7677: Good point, should be better now.

src/meson.build
jp7677: Yes, for testing we need to abstract the Vulkan interaction away behind our own Vulkan wrapper, like we did within `sysinfo`. The resource factory has to create the wrapper so that we can mock this away for tests (the tests have their own resource factory which returns the mocks). I hope it works nicely to share the wrapper and resource creation between `nvapi` and `nvofapi`, but I haven't played yet.

src/meson.build
jp7677: Sounds good. Ok to wait with this until we have an idea how the tests plays out?

Saancreed: Yup!

Saancreed (approved):

PR #222, opened 2024-11-07
NvOFInstanceD3D12: Implement input/output fence points

liam-middlebrook:
Needed for older versions of nvngx_dlssg.dll where OF buffer registration was synchronized with GPU work.

### DIFF
--- a/src/nvofapi/nvofapi_d3d12_instance.cpp
+++ b/src/nvofapi/nvofapi_d3d12_instance.cpp
@@ -100,13 +100,15 @@ namespace dxvk {
         uint64_t offset;
         m_device->GetVulkanResourceInfo1(registerParams->resource, reinterpret_cast<UINT64*>(&vkParams.image), &offset, &vkParams.format);
 
-        // ID3D12 fence to timeline semaphore
-        // no inputFencePoint/outputFencePoint equivalents for VK, leaving as
-        // no-op :(
-        assert(registerParams->inputFencePoint.fence == nullptr);
-        assert(registerParams->inputFencePoint.value == 0);
-        assert(registerParams->outputFencePoint.fence == nullptr);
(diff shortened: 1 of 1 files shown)


liam-middlebrook:
Thanks to @adamdmoss for reporting this bug. If you're able to confirm this resolves your issue I'd appreciate it.

jp7677 (approved):

adamdmoss:
Gollum is very happy now.  Thanks!

adamdmoss:
Oh, it fixes the rat game crash too!  Hooray!  It has weird unpleasant out-of-order frame issues when FG is enabled but that's better than being in a crashloop.

adamdmoss:
> Gollum is very happy with this PR. Thanks!

Oops, spoke too soon.  It has the same problem as Plague Tale Requiem with this fix; it no longer crashes but it eventually starts presenting frames out-of-order.  Just took longer to repro (like a couple of minutes) with Gollum.

adamdmoss:
So -> IMHO this fix is a definite improvement (visual glitches are an improvement over complete crashing) and I'd love it to land, but it either reveals or causes further issues with the two titles it stops crashing so it's not an ultimate fix.

jp7677:
@liam-middlebrook Do you want to address Adams remaining points as part of this PR or do you prefer if we land this now as intermediate improvement?

Saancreed (approved):

liam-middlebrook:
> @liam-middlebrook Do you want to address Adams remaining points as part of this PR (if it is even doable here) or do you prefer if we land this now as an intermediate improvement?

I don't think we should wait for that. There are a few different modules at work here, and the issues Adam describes don't sound like they would be due to `nvofapi64.dll`.

jp7677:
Thanks!

PR #224, opened 2024-11-09
Use COM template

jp7677:
This should take care of calling `Release()`. The members have been reordered to ensure that release order doesn't change.

While at it, also move `ResourceFactory` to `nvapi`, I don't see us sharing this since it would pull `nvml` and `lfx` into `nvofapi`.


tests/meson.build
```diff
@@ -2,6 +2,7 @@ nvapi_src = files([
   '../src/util/util_string.cpp',
   '../src/util/util_env.cpp',
   '../src/util/util_log.cpp',
+    '../src/nvapi/resource_factory.cpp',
```
Saancreed, line 5: Indentation mismatch?

Saancreed (approved):
Looks reasonable, just one style nitpick.

PR #227, opened 2024-11-16
Maxwell GM107/GM108 support

pchome:
IDK what `GM000` for but lets use it for now.

Test reports:
```patch
-    Architecture ID:            0x000000e0 (Kepler)
+    Architecture ID:            0x00000110 (Maxwell)
 ...
-    Compute capable:            -
+    Compute capable:            Yes (Compute GPU topology flags: 0x0b)
```

```log
Filters: [d3d],[d3d11],[d3d12],[drs],[sysinfo],[sysinfo-topo],[sysinfo-nvml],[sysinfo-hdr],[util],[system]
...
All tests passed (54839 assertions in 18 test cases)
```

### DIFF
--- a/src/nvapi/nvapi_adapter.cpp
+++ b/src/nvapi/nvapi_adapter.cpp
@@ -268,6 +268,10 @@ namespace dxvk {
             || (HasNvkDriver() && IsVkDeviceExtensionSupported(VK_EXT_SHADER_VIEWPORT_INDEX_LAYER_EXTENSION_NAME)))
             return NV_GPU_ARCHITECTURE_GM200;
 
+        // VK_EXT_shader_image_atomic_int64 is supported on Maxwell 1 (GM10x) and newer
+        if (IsVkDeviceExtensionSupported(VK_EXT_SHADER_IMAGE_ATOMIC_INT64_EXTENSION_NAME))
+            return NV_GPU_ARCHITECTURE_GM000;
+
         // Fall back to Kepler
         return NV_GPU_ARCHITECTURE_GK100;
     }
--- a/src/nvapi_d3d12.cpp
+++ b/src/nvapi_d3d12.cpp
@@ -236,6 +236,7 @@ extern "C" {
                 pGraphicsCaps->minorSMVersion = 0;
                 break;
             case NV_GPU_ARCHITECTURE_GM200:
+            case NV_GPU_ARCHITECTURE_GM000:
                 pGraphicsCaps->majorSMVersion = 5;
                 pGraphicsCaps->minorSMVersion = 0;
                 break;
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -184,6 +184,7 @@ namespace dxvk::env {
     if (overrideStr == #arch) \
         override = NV_GPU_ARCHITECTURE_##arch;
             CHECK_ARCH(GK100)
+            CHECK_ARCH(GM000)
             CHECK_ARCH(GM200)
             CHECK_ARCH(GP100)
             CHECK_ARCH(GV100)
--- a/README.md
+++ b/README.md
@@ -79,7 +79,8 @@ The following environment variables tweak DXVK-NVAPI's runtime behavior:
 - `DXVK_NVAPI_DISABLE_ENTRYPOINTS`, when set to a comma-separated list of NVAPI entrypoint names, will hide their implementation from the application. For example, `DXVK_NVAPI_DISABLE_ENTRYPOINTS=NvAPI_D3D11_BeginUAVOverlap,NvAPI_D3D11_EndUAVOverlap` will report D3D11 barrier control extensions as not available.
 - `DXVK_NVAPI_GPU_ARCH`, when set to one of supported NVIDIA GPU architecture IDs will override reported GPU architecture. Currently supported values are:
   - `GK100` (Kepler)
-  - `GM200` (Maxwell)
+  - `GM000` (Maxwell Gen1)
+  - `GM200` (Maxwell Gen2)
   - `GP100` (Pascal)
   - `GV100` (Volta)
   - `TU100` (Turing)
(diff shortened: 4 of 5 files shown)


jp7677:
Thanks. What is your motivation for this PR, does this changes something in a real life scenario?

pchome:
Reports 'Compute capable' for PhysX.

FluidMark_0001.log
```log
FluidMark 1.5.2 startup...
Checking NVIDIA PhysX hardware...
NVIDIA PhysX engine version 2.8.4 successfully started.
NVIDIA PhysX engine - found 1 PhysX GPU(s).
...
NVIDIA PhysX engine init - gpuHeapSize=512MB
NVIDIA PhysX engine - # of PPUs detected: 0
NVIDIA PhysX: PhysX hardware not detected (NX_HW_VERSION_NONE).
NVIDIA PhysX engine version 2.8.4 successfully started.
NVIDIA PhysX shutdown ok.
...
```

I usually play Borderlands w/ Full HW PhysX enabled, adds a lot of details and fun.

jp7677:
Please be a bit more verbose, I don't want to read between the lines.

So you have a Maxwell GM107/108 card and this change allows you to enable PhysX and it also works? Above log lines are from FluidMark without this patch, but with this patch the log changes and the benchmark works on GM107/108?

pchome:
> So you have a Maxwell GM107/108 card and this change allows you to enable PhysX and it also works?

Yes.
750Ti, GM107
GM108 - are something 800M (mobile(?))

> Above log lines are from FluidMark without this patch, but with this patch the log changes and the benchmark works on GM107/108?

**With** this patch.
`NVIDIA PhysX engine - found 1 PhysX GPU(s).` - indication of success.
and
`NVIDIA PhysX: PhysX hardware not detected (NX_HW_VERSION_NONE).` - seems just reporting that it's not PPU (dedicated PhysX processor(?))

Without this patch (or `nvcuda.dll`) log lines are a bit different:
`NVIDIA PhysX engine - found 0 PhysX GPU(s).` - 'Compute capable' not found
no `NVIDIA PhysX engine init` line and no other PhysX related lines an the end.

Also there is a checkmark to enable GPU PhysX in FluidMark, which available with patch and disabled w/o.

src/nvapi_d3d12.cpp
```diff
@@ -236,6 +236,7 @@ extern "C" {
                 pGraphicsCaps->minorSMVersion = 0;
                 break;
             case NV_GPU_ARCHITECTURE_GM200:
```
jp7677, line 238: Accoording to https://en.wikipedia.org/wiki/Maxwell_(microarchitecture) we should correct this and report sm52 for Maxwell Gen2 and sm50 for Maxwell Gen2.

src/nvapi/nvapi_adapter.cpp
```diff
@@ -268,6 +268,10 @@ namespace dxvk {
             || (HasNvkDriver() && IsVkDeviceExtensionSupported(VK_EXT_SHADER_VIEWPORT_INDEX_LAYER_EXTENSION_NAME)))
             return NV_GPU_ARCHITECTURE_GM200;
 
+        // VK_EXT_shader_image_atomic_int64 is supported on Maxwell 1 (GM10x) and newer
```
jp7677, line 271: This is also the case for NVK (https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/nouveau/vulkan/nvk_physical_device.c?ref_type=heads#L244) which is nice!

src/util/util_env.cpp
```diff
@@ -184,6 +184,7 @@ namespace dxvk::env {
     if (overrideStr == #arch) \
         override = NV_GPU_ARCHITECTURE_##arch;
             CHECK_ARCH(GK100)
+            CHECK_ARCH(GM000)
```
jp7677, line 187: We should also add this to the readme at https://github.com/jp7677/dxvk-nvapi/blob/master/README.md#tweaks-debugging-and-troubleshooting

jp7677 (changes requested):
Thanks for the clarification. I would love to see a few more additions, but looks good in general.

We need to extend/adjust the tests to cover the new cases:
https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_sysinfo.cpp#L742 and https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_sysinfo.cpp#L309 and https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_d3d12.cpp#L243

It would also be nice to make a distinction between `Maxwell (Gen1)` and `Maxwell (Gen2)` here https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_system.cpp#L46

src/nvapi/nvapi_adapter.cpp
pchome: This is the soure ;) so no double check.
As double check I filtered all devices by this thing on [vulkan.gpuinfo.org](https://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_EXT_shader_image_atomic_int64&platform=linux) with additional finter for NV -- only gen1 maxwells and some weird setups was lowest.

src/nvapi_d3d12.cpp
pchome: See info about `Series_*` too.

pchome: Well, `*_series`.
https://en.wikipedia.org/wiki/GeForce_700_series
https://en.wikipedia.org/wiki/GeForce_800M_series
https://en.wikipedia.org/wiki/GeForce_900_series

All contains info about `numbers` for Maxwell Gen1.
Just in case.

jp7677:
Hi @pchome I took the liberty to add the remaining bits here #228 
I've added your name as co-author since nearly all of the work came from this PR.

jp7677:
Merged manually with https://github.com/jp7677/dxvk-nvapi/pull/228

PR #228, opened 2024-11-19
Maxwell GM107/GM108 support (2)

jp7677:
Supersedes https://github.com/jp7677/dxvk-nvapi/pull/227


src/nvapi_gpu.cpp
```diff
@@ -595,6 +595,8 @@ extern "C" {
                 implementationId = NV_GPU_ARCH_IMPLEMENTATION_GP102;
                 break;
             case NV_GPU_ARCHITECTURE_GM200:
+            case NV_GPU_ARCHITECTURE_GM000:
+                // There is no NV_GPU_ARCH_IMPLEMENTATION_GM107
```
Saancreed, line 599: I'm now curious how native NVAPI reports it.

src/interfaces/vkd3d-proton_interfaces.h
```diff
@@ -130,7 +130,6 @@ ID3D12DXVKInteropDevice : public IUnknown {
 };
 MIDL_INTERFACE("902d8115-59eb-4406-9518-fe00f991ee65")
 ID3D12DXVKInteropDevice1 : public ID3D12DXVKInteropDevice {
-
```
Saancreed, line 133: Unrelated but if you're touching this, could you please clang-format interface headers, even if that requires temporarly getting rid of `MIDL_INTERFACE`s to unconfuse it? That `TYPE * name` is starting to hurt my eyes.

Saancreed (approved):

src/nvapi_gpu.cpp
jp7677: If I had to guess it would be `0x00000007` considering the pattern of the values in `NV_GPU_ARCH_IMPLEMENTATION_ID` ;)

src/interfaces/vkd3d-proton_interfaces.h
jp7677: not part of this pr, but yeah agree ;)

PR #229, opened 2024-11-25
nvapi-d3d: Fix NV_LATENCY_MARKER_TYPE forwarding

jp7677:
Unfortunately `NV_LATENCY_MARKER_TYPE` doesn't match `VkLatencyMarkerNV`, so we need to convert manually. Silently acknowledge, but ignore, calls with unsupported values.

cc: @esullivan-nvidia

### DIFF
--- a/src/nvapi/nvapi_d3d_low_latency_device.cpp
+++ b/src/nvapi/nvapi_d3d_low_latency_device.cpp
@@ -1,4 +1,6 @@
 #include "nvapi_d3d_low_latency_device.h"
+#include "../util/util_string.h"
+#include "../util/util_log.h"
 
 namespace dxvk {
     LowLatencyFrameIdGenerator::LowLatencyFrameIdGenerator()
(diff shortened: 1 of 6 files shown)


src/nvapi_d3d.cpp
```diff
@@ -272,7 +272,16 @@ extern "C" {
         if (nvapiD3dInstance->IsUsingLfx() || !NvapiD3dLowLatencyDevice::SupportsLowLatency(pDev))
             return NoImplementation(n, alreadyLoggedNoImpl);
 
-        if (!NvapiD3dLowLatencyDevice::SetLatencyMarker(pDev, pSetLatencyMarkerParams->frameID, pSetLatencyMarkerParams->markerType))
+        auto markerType = NvapiD3dLowLatencyDevice::ToMarkerType(pSetLatencyMarkerParams->markerType);
```
Saancreed, line 275: Same thing should be done in `NvAPI_D3D12_SetAsyncFrameMarker`.

jp7677: ah, should practice to not only search in the current file ;)

src/nvapi/nvapi_d3d_low_latency_device.cpp
```diff
@@ -110,6 +110,31 @@ namespace dxvk {
...
+            case (PC_LATENCY_PING):
+                return {};
+            case (OUT_OF_BAND_RENDERSUBMIT_START):
+            case (OUT_OF_BAND_RENDERSUBMIT_END):
+            case (OUT_OF_BAND_PRESENT_START):
+            case (OUT_OF_BAND_PRESENT_END):
+                return markerType - 1;
+        }
+
+        return {};
```
Saancreed, line 135: One concern about throwing `PC_LATENCY_PING` and invalid values in the same bag is that if new values are introduced in the future, detecting that they are used will be much harder. So, I'd prefer just logging invalid values here when one is encountered (at least once when we are passed anything outside known `SIMULATION_START` .. `OUT_OF_BAND_PRESENT_END` range), but it's not impossible that this won't ever matter.

Saancreed (commented):
Tests will need some adjustments. It would be nice if we also verified that `PC_LATENCY_PING` and invalid values don't cause any calls to `ID3DLowLatencyDevice` but we can deal with that later.

src/nvapi/nvapi_d3d_low_latency_device.cpp
jp7677: Good idea, I've added the log statement, also unconditionally because this should be in our faces.

src/nvapi_d3d.cpp
```diff
@@ -272,7 +272,16 @@ extern "C" {
...
+        if (!markerType.has_value()) {
+            // Silently drop unsupported marker types
+            if (!alreadyLoggedOk)
+                log::info(str::format("Not supported NV_LATENCY_MARKER_TYPE: ", pSetLatencyMarkerParams->markerType));
```
Saancreed, line 279: This can probably be removed now (and from the other function as well), especially since it's unlikely to ever log anything when first valid call with correct marker type will set `alreadyLoggedOk`.

Saancreed (approved):
Last minor cleanup and LGTM otherwise.

src/nvapi_d3d.cpp
jp7677: thx, changed with new `bool` to log this at least once.

Saancreed: Yup, that should work fine.

Saancreed (approved):

adamdmoss:
Good catch.  Out of interest, was this spotted by code inspection or was a title malfunctioning?

jp7677:
> Good catch. Out of interest, was this spotted by code inspection or was a title malfunctioning?

This was spotted by chance on LGD by looking at results from a run with Vulkan Validation Layers to hunt an unrelated issue.

PR #232, opened 2024-11-29
util: Use build target name as log file name

jp7677:
So write to nvapi.log, nvapi64.log and nvofapi64.log separately.

### DIFF
--- a/src/util/util_log.cpp
+++ b/src/util/util_log.cpp
@@ -26,7 +26,7 @@ namespace dxvk::log {
 
         constexpr auto logLevelEnvName = "DXVK_NVAPI_LOG_LEVEL";
         constexpr auto logPathEnvName = "DXVK_NVAPI_LOG_PATH";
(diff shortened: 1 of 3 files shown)


src/util/util_log.cpp
```diff
@@ -81,8 +81,7 @@ namespace dxvk::log {
             std::setfill('0'), std::setw(3), milliseconds, ":",
             std::setfill('0'), std::setw(4), std::hex, ::GetCurrentProcessId(), ":",
             std::setfill('0'), std::setw(4), std::hex, ::GetCurrentThreadId(), ":",
-            level, ":",
-            DXVK_NVAPI_TARGET_NAME, ":",
+            level, ":" DXVK_NVAPI_TARGET_NAME ":",
```
jp7677, line 84: I needed some time to get used to the idea ;)

Saancreed (approved):
I kind of liked having a single timeline but that should still be the case in Wine log so :ok_hand:

PR #237, opened 2024-12-17
Vulkan Reflex support

### LINKED ISSUE #166: vk_streamline: Vulkan reflex support (frame limiter) fails to work
opened an issue, as I asked on discord..
I'm running vk_streamline NV Pro sample:
https://github.com/nvpro-samples/vk_streamline 

it's interesting sample, because is a simple example using Reflex Vulkan, DLSS Vulkan and DLSS-G Vulkan..
Reflex Vulkan seems uses NvLowLatencyVk.dll which has this entrypoints:
```
NvLL_VK_DestroyLowLatencyDevice
NvLL_VK_GetLatency
NvLL_VK_GetSleepStatus
NvLL_VK_InitLowLatencyDevice
NvLL_VK_Initialize
NvLL_VK_NotifyOutOfBandQueue
NvLL_VK_SetLatencyMarker
NvLL_VK_SetSleepMode
NvLL_VK_Sleep
NvLL_VK_Unload
```
really, all I see is a nvapi_QueryInterface string but no nvapi64.dll reference in the DLL..

I attach sample binary in case you trust me to do your simple experiments:

https://drive.google.com/file/d/13IrihRqwYVVNPGjf3wU-wIKHGqNSFwp_/view?usp=sharing

tested on Ubuntu 23.10.. but when I compiled I tested initially on Ubu 23.04 I believe and worked also..

note I use:

```
export WINEHAGS=0
export DXVK_ENABLE_NVAPI=1

```
if on a hags enabled Wine build, as using WINEHAGS=1 crashes due to dlss-g issues..

using latest dxvk-nvapi (reflex enabled) and a enabled reflex (ll2 driver) I see following errors:

```
[07.03.2024 19-52-30][streamline][warn]vulkan.cpp:867[init] Low latency API for VK failed to initialize device -3
[07.03.2024 19-52-31][streamline][error]vulkan.cpp:3069[setSleepMode] NvLL_VK_SetSleepMode(m_device, &params) failed - error 4294967293
[07.03.2024 19-52-31][streamline][error]reflexentry.cpp:321[slSetData] ctx.compute->setSleepMode(ctx.constants) failed
[07.03.2024 19-52-31][streamline][error]vulkan.cpp:3115[sleep] NvLL_VK_Sleep(m_device, reflexSemaphoreValue) failed - error 4294967293
[07.03.2024 19-52-35][streamline][warn]reflexentry.cpp:275[slSetData] Low-latency modes are only supported on NVIDIA hardware through Reflex, collecting latency stats only
```

Saancreed:
Implemented via both `NvAPI_Vulkan_*` entrypoints and corresponding Linux-side Vulkan layer.

Closes #166.


Saancreed:
Okay, I think this is now ready for initial round of review. Still has some hacks but I have no idea how to rid of them. And I'll try to test this version again with various games I happen to have installed, for now I know that at least vk_streamline sample works with this.

Feel free to ask about strange stuff in there; I lived with a variant of this code for months so I might have got used to the weirdness.

Saancreed:
Tested with Enshrouded, Portal RTX and Portal Prelude RTX, appears to work fine with them. Still no idea how to convince No Man's Sky and Indiana Jones and the Great Circle to let me activate Reflex.

Remaining titles that would be nice to have tested (that I know of):

* Path of Exile 2
* Red Dead Redemption 2
* Tom Clancy's Rainbow Six Extraction
* Tom Clancy's Rainbow Six Siege? Does it still have its Vulkan renderer?

Saancreed:
I figured out the problem NMS and Indy had: Streamline was thinking that `NvLowLatencyVk.dll` shipped with the game wasn't correctly signed, most likely due to Wine not implementing the signature verification functions (correctly?). With that issue… sidestepped, both NMS and Indy load the library and make Reflex calls, but:

* NMS only calls `SetSleepMode` and `SetLatencyMarker`, it doesn't seem to be performing the actual latency sleeps.
* Indy works, but spams the log with `unsupported NV_VULKAN_LATENCY_MARKER_TYPE (13)` which doesn't exist as a Vulkan latency marker type but is likely related to [PCLMarker::eControllerInputSample](https://github.com/NVIDIAGameWorks/Streamline/blob/v2.4.15/include/sl_pcl.h#L73).

jp7677:
Thanks for the PR! Cool to have Vulkan Reflex.

I did a first pass on the nvapi side of things for some initial thoughts. The gist is more or less that it would be nice to integrate more with the existing structure:
-   `nvapi_vkreflex.cpp` should be `nvapi_vulkan.cpp` to be consistent with other entry points file (due to  `NvAPI_Vulkan_` )
- On the D3D side we have `NvapiD3dLowLatencyDevice`, what about introducing `NvapiVulkanLowLatencyDevice` for details and Vulkan initialization so that the entry point files are mostly for logging, arguments checking and status codes? The `VkReflexContextData` is I guess already quite close to this.
- Other global things do live in `nvapi_globals`, the context map (or map of VulkanLowLatencyDevices) could also live there.
-  With the mind of tests it would be cool to reuse and extend  the existing `Vulkan` class for abstracting the VK calls. Creating should still go through ResourceFactory::CreateVulkan with a new overload for  just the library name.

jp7677:
>     * Indy works, but spams the log with `unsupported NV_VULKAN_LATENCY_MARKER_TYPE (13)` which doesn't exist as a Vulkan latency marker type but is likely related to [PCLMarker::eControllerInputSample](https://github.com/NVIDIAGameWorks/Streamline/blob/v2.4.15/include/sl_pcl.h#L73).

Impressive find. Probably a good idea to log only once per invalid marker type (also on the D3D side).

Saancreed:
> * `nvapi_vkreflex.cpp` should be `nvapi_vulkan.cpp` to be consistent with other entry points file (due to  `NvAPI_Vulkan_` )

I actually tried that first, only to be told to go away with my nonsense by Meson:

> `ERROR: Multiple producers for Ninja target "src/nvapi64.dll.p/nvapi_vulkan.cpp.obj". Please rename your targets.`

I think `nvapi_vulkan.cpp` _somehow_ collides with `nvapi/vulkan.cpp`, but that sounds incredibly stupid.

jp7677:
> > * `nvapi_vkreflex.cpp` should be `nvapi_vulkan.cpp` to be consistent with other entry points file (due to  `NvAPI_Vulkan_` )
> 
> I actually tried that first, only to be told to go away with my nonsense by Meson:
> 
> > `ERROR: Multiple producers for Ninja target "src/nvapi64.dll.p/nvapi_vulkan.cpp.obj". Please rename your targets.`
> 
> I think `nvapi_vulkan.cpp` _somehow_ collides with `nvapi/vulkan.cpp`, but that sounds incredibly stupid.

ah yeah, it is exactly this... I remember hitting this when playing with `nvofapi` initially. Meson flattens paths by converting the separator to underscore which conflicts in this case. Maybe rename `nvapi/vulkan.cpp` to `nvapi/vk.cpp`?

Saancreed:
It does not. For now you should assume that any game with Vulkan renderer that isn't Portal RTX won't have DLSS Frame Generation function correctly.

Indy is a bit more of a special case because it's trying to use VK-CUDA interop for DLFG, we are investigating this here: https://github.com/SveSop/nvcuda/issues/9

Saancreed:
Pushed new version, but with changes (when compared to previous revision) as fixup commits, to hopefully make reviewing them easier. Now to figure out what Proton's mingw wants me to do to make it happy…

src/nvapi_globals.h
```diff
@@ -4,8 +4,10 @@
 #include "nvapi/resource_factory.h"
 #include "nvapi/nvapi_adapter_registry.h"
 #include "nvapi/nvapi_d3d_instance.h"
+#include "nvapi/vk.h"
 
 extern uint64_t initializationCount;
 extern std::unique_ptr<dxvk::ResourceFactory> resourceFactory;
 extern std::unique_ptr<dxvk::NvapiAdapterRegistry> nvapiAdapterRegistry;
 extern std::unique_ptr<dxvk::NvapiD3dInstance> nvapiD3dInstance;
+extern std::unique_ptr<dxvk::Vk> vkLowLatency;
```
jp7677, line 13: Since this `Vk` instance is only needed for `NvapiVulkanLowLatencyDevice`, we might also keep it alive next to `static std::unordered_map<VkDevice, NvapiVulkanLowLatencyDevice> m_lowLatencyDeviceMap;` as another static field of `NvapiVulkanLowLatencyDevice` and initialize as part of e.g. `GetOrCreate(ResourceFactory& resourceFactory, VkDevice device)`.

src/nvapi/nvapi_vulkan_low_latency_device.cpp
```diff
@@ -0,0 +1,225 @@
...
+    std::unordered_map<VkDevice, NvapiVulkanLowLatencyDevice> NvapiVulkanLowLatencyDevice::m_lowLatencyDeviceMap = {};
+    std::mutex NvapiVulkanLowLatencyDevice::m_lowLatencyDeviceMutex = {};
+
+    std::pair<NvapiVulkanLowLatencyDevice*, VkResult> NvapiVulkanLowLatencyDevice::GetOrCreate(VkDevice device, PFN_vkGetDeviceProcAddr pvkGetDeviceProcAddr) {
+        std::scoped_lock lock{m_lowLatencyDeviceMutex};
+
+        if (auto lowLatencyDevice = Get(device))
+            return {lowLatencyDevice, VK_SUCCESS};
+
+#define VK_GET_DEVICE_PROC_ADDR(proc) auto proc = reinterpret_cast<PFN_##proc>(pvkGetDeviceProcAddr(device, #proc))
```
jp7677, line 13: Clang-tidy suggestion: Macro argument should be enclosed in parentheses 
`#define VK_GET_DEVICE_PROC_ADDR(proc) auto (proc) = reinterpret_cast<PFN_##proc>(pvkGetDeviceProcAddr(device, #proc))`

src/nvapi_vulkan.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+        auto lowLatencyDevice = NvapiVulkanLowLatencyDevice::Get(device);
+
+        if (!lowLatencyDevice)
+            return HandleInvalidated(n);
+
+        static constexpr auto count = sizeof(pGetLatencyParams->frameReport) / sizeof(*pGetLatencyParams->frameReport);
+        static_assert(count == 64);
+
+        std::array<VkLatencyTimingsFrameReportNV, count> timings;
```
jp7677, line 177: Depending on how often this is called, may be wise to initialize this for safety: 
`std::array<VkLatencyTimingsFrameReportNV, count> timings{}; `

jp7677 (commented):
Thanks for the rework of the `nvapi` side. See comments for a few thoughts. It would also be nice to have some tests covering the basics, but I'm fine to do this in a separate PR (but we shouldn't forget since everything is in place now for that).

src/nvapi_vulkan.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+        if (!pGetSleepStatusParams)
+            return InvalidPointer(n);
+
+        if (pGetSleepStatusParams->version != NV_VULKAN_GET_SLEEP_STATUS_PARAMS_VER1)
+            return IncompatibleStructVersion(n, pGetSleepStatusParams->version);
+
+        auto lowLatencyDevice = NvapiVulkanLowLatencyDevice::Get(device);
+
+        // TODO: check native behavior for this case
```
jp7677, line 93: is this TODO for later of for this PR?

jp7677:
Looking at the layer, some initial thoughts:

- The layer itself looks fine at first look, also good idea to use `vkroots`.
- We should also built the layer as part of CI including proper naming of the artifact on a tag 
- We also need to document building and installing in a user home. 
- May be we can repeat the `version.h` generation and log the version on `Init()`?
- Not sure yet if we should use different env names for the layer. May be it is easier to reuse the existing name.

src/nvapi_vulkan.cpp
Saancreed: It might be called every frame, and this thing is unfortunately quite huge. IIRC when Strudel was testing old flavor of this in Path of Exile II, presence of latency reports actually had negative impact on FPS overall.

I think it's enough that `NvapiVulkanLowLatencyDevice` ensures that important members (`.sType` and `.pNext`) are initialized before calling Vulkan function that fills the rest.

jp7677: Alright, thanks for additional context and agree to keep your version.

src/nvapi/nvapi_vulkan_low_latency_device.cpp
Saancreed: I can do this but it feels weird. I'd never write `auto (variable) = …` myself and this is what it would expand to.

src/nvapi_vulkan.cpp
jp7677: Thanks and fine ofc to keep the todo en chack later. Since the impact is probably zero I would be ok with making the code active now. So: 

```cpp
        // TODO: check native behavior for this case
        if (!lowLatencyDevice)
            return HandleInvalidated(n);
```

src/nvapi_globals.h
Saancreed: I gave this a thought and I'm not sure about this. I wanted `NvapiVulkanLowLatencyDevice` to not be aware of any other implemetation detail like `ResourceFactory` so I kept `NvAPI_Vulkan_InitLowLatencyDevice` as the only piece that connects both worlds; accepting just `vkGetDeviceProcAddr` made it standalone and very self-contained.

But if you really want it like this, I'd suggest separating `GetOrCreate(ResourceFactory& resourceFactory, VkDevice device)` into `InitializeVulkan(ResourceFactory& resourceFactory)` and `GetOrCreate(VkDevice device)`.

src/nvapi/nvapi_vulkan_low_latency_device.cpp
Saancreed: Hmm, for me clang-tidy doesn't report this, instead it reports

```
../src/nvapi/nvml.cpp:52:89: warning: Call to virtual method 'Nvml::ErrorString' during construction bypasses virtual dispatch [clang-analyzer-optin.cplusplus.VirtualCall]
   52 |             log::info(str::format("NVML loaded but initialization failed with error: ", ErrorString(result)));
      |                                                                                         ^~~~~~~~~~~~~~~~~~~
```

and

```
../src/util/util_string.cpp:41:9: warning: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119 [clang-analyzer-security.insecureAPI.strcpy]
   41 |         strcpy(nvss, str.c_str());
      |         ^~~~~~
```

:thinking:

jp7677: Screenshot
...

Interestingly I don't get your warnings, at least the first one sounds somewhat sensible..

Saancreed: Perhaps clion brings its own configuration that doesn't match the default one I get when I run `clang-tidy` from CLI?

Anyway, if this is the only macro it complains about I'll just go ahead and change it.

src/nvapi_globals.h
jp7677: ok, I get your point wrt keeping `NvapiVulkanLowLatencyDevice` clean. My reasoning was that having `std::unique_ptr<dxvk::Vk> vkLowLatency` in `globals` looks somewhat "naked", especially since the other part of the static application skeleton for vkreflex, the list of created `NvapiVulkanLowLatencyDevice`'s, lives in `NvapiVulkanLowLatencyDevice` itself. Also looking at other parts, on the sysinfo side we have the registry for connecting adapters and entrypoints, which also hold the `Vk` instance , for lfx there is `NvapiD3dInstance` which hold the `Lfx` instance. 
Creating  `Initialize(ResourceFactory& resourceFactory)` and `GetOrCreate(VkDevice device)` sounds also fine of course and it fits even better.
If you dislike this approach though, we can keep your initial version. But if I had to choose, I would go for `Initialize()` and `GetOrCreate()` in `NvapiVulkanLowLatencyDevice`.

Saancreed: I think I'll go with `Initialize` then.

src/nvapi/nvapi_vulkan_low_latency_device.cpp
jp7677: > Perhaps clion brings its own configuration that doesn't match the default one I get when I run `clang-tidy` from CLI?

Very likely that.
 
> I can do this but it feels weird. I'd never write `auto (variable) = …` myself and this is what it would expand to.

Actually agree, feel free to also just drop this suggestion.

Saancreed: Ugh, I already committed that, then I noticed it caused the _compiler_ to complain instead:

```
../src/nvapi/nvapi_vulkan_low_latency_device.cpp:36:43: warning: unnecessary parentheses in declaration of ‘vkCreateSemaphore’ [-Wparentheses]
   36 | #define VK_GET_DEVICE_PROC_ADDR(proc) auto(proc) = reinterpret_cast<PFN_##proc>(pvkGetDeviceProcAddr(device, #proc))
      |                                           ^~~~~~
../src/nvapi/nvapi_vulkan_low_latency_device.cpp:38:9: note: in expansion of macro ‘VK_GET_DEVICE_PROC_ADDR’
   38 |         VK_GET_DEVICE_PROC_ADDR(vkCreateSemaphore);
      |         ^~~~~~~~~~~~~~~~~~~~~~~
../src/nvapi/nvapi_vulkan_low_latency_device.cpp:36:43: note: remove parentheses
   36 | #define VK_GET_DEVICE_PROC_ADDR(proc) auto(proc) = reinterpret_cast<PFN_##proc>(pvkGetDeviceProcAddr(device, #proc))
      |                                           ^~~~~~
```

I'll just drop it then. We should probably figure out some reasonable `.clang-tidy` config and commit it to the repo.

jp7677:
LGTM wrt the NVAPI side. Suggested change for avoiding a nullptr check for the resource factory (because that already happens in `NvAPI_Vulkan_InitLowLatencyDevice`) and for keeping the `Initialize()` signature identical with e.g. adapter registry:   

<details>
<summary>Diff</summary>

```diff
diff --git a/src/nvapi/nvapi_vulkan_low_latency_device.cpp b/src/nvapi/nvapi_vulkan_low_latency_device.cpp
index 7472aac..f53e434 100644
--- a/src/nvapi/nvapi_vulkan_low_latency_device.cpp
+++ b/src/nvapi/nvapi_vulkan_low_latency_device.cpp
@@ -5,21 +5,18 @@ namespace dxvk {
     std::unordered_map<VkDevice, NvapiVulkanLowLatencyDevice> NvapiVulkanLowLatencyDevice::m_lowLatencyDeviceMap = {};
     std::mutex NvapiVulkanLowLatencyDevice::m_mutex = {};
 
-    bool NvapiVulkanLowLatencyDevice::TryInitialize(ResourceFactory* resourceFactory) {
+    bool NvapiVulkanLowLatencyDevice::Initialize(ResourceFactory& resourceFactory) {
         std::scoped_lock lock{m_mutex};
 
         if (m_vk && m_vk->IsAvailable())
             return true;
 
-        if (resourceFactory == nullptr)
-            return false;
-
-        m_vk = resourceFactory->CreateVulkan("vulkan-1.dll");
+        m_vk = resourceFactory.CreateVulkan("vulkan-1.dll");
 
         if (m_vk && m_vk->IsAvailable())
             return true;
 
-        m_vk = resourceFactory->CreateVulkan("winevulkan.dll");
+        m_vk = resourceFactory.CreateVulkan("winevulkan.dll");
 
         return m_vk && m_vk->IsAvailable();
     }
diff --git a/src/nvapi/nvapi_vulkan_low_latency_device.h b/src/nvapi/nvapi_vulkan_low_latency_device.h
index fbeb9f8..55b086d 100644
--- a/src/nvapi/nvapi_vulkan_low_latency_device.h
+++ b/src/nvapi/nvapi_vulkan_low_latency_device.h
@@ -6,7 +6,7 @@
 namespace dxvk {
     class NvapiVulkanLowLatencyDevice {
       public:
-        static bool TryInitialize(ResourceFactory* resourceFactory);
+        static bool Initialize(ResourceFactory& resourceFactory);
 
         [[nodiscard]] static std::pair<NvapiVulkanLowLatencyDevice*, VkResult> GetOrCreate(VkDevice device);
         [[nodiscard]] static NvapiVulkanLowLatencyDevice* Get(VkDevice device);
diff --git a/src/nvapi_vulkan.cpp b/src/nvapi_vulkan.cpp
index 74ff113..43449be 100644
--- a/src/nvapi_vulkan.cpp
+++ b/src/nvapi_vulkan.cpp
@@ -16,7 +16,7 @@ extern "C" {
             return ApiNotInitialized(n);
 
         static std::once_flag initialized{};
-        std::call_once(initialized, []() { NvapiVulkanLowLatencyDevice::TryInitialize(resourceFactory.get()); });
+        std::call_once(initialized, []() { NvapiVulkanLowLatencyDevice::Initialize(*resourceFactory); });
 
         auto device = reinterpret_cast<VkDevice>(vkDevice);
         auto semaphore = reinterpret_cast<VkSemaphore*>(signalSemaphoreHandle);
```

</details>

Saancreed:
Weird, when I initially tried passing as reference clangd complained about binding to temporary value. But your version works so I applied it, thanks.

layer/vulkan_reflex_layer.cpp
```diff
@@ -0,0 +1,896 @@
...
+
+// hijack VKAPI_ATTR before vk_layer.h is included to make sure that vkNegotiateLoaderLayerInterfaceVersion
+// has appropriate visibility attribute in its initial declaration / prototype, otherwise MSVC will complain
+
+#undef VKAPI_ATTR
+
+#if defined(_WIN32)
+#define VKAPI_ATTR __declspec(dllexport)
+#elif defined(__GNUC__)
+#define VKAPI_ATTR [[gnu::visibility("default")]]
```
Saancreed, line 24: As a side note, I would prefer to do the attribute check first because at least that is standard, and then leave the `declspec` for MSVC-like compilers…

```cpp
#if __has_cpp_attribute(gnu::visibility)
#define VKAPI_ATTR [[gnu::visibility("default")]]
#elif defined(_MSC_VER)
#define VKAPI_ATTR __declspec(dllexport)
```

But this is how vkroots does it and unfortunately my version would trip `clang-cl` which supports both, causing it to complain about `dllexport` being added later :slightly_frowning_face:

Maybe the idea of having a Windows-side layer is not worth the effort.

jp7677: Indeed lots of extra declarations for `msvc`. That said, since it builds now, I'm fine with keeping this.

Saancreed:
> * I'm fine to enable the layer by default so that explicitly setting `DXVK_NVAPI_VKREFLEX` is not needed. Manually setting up the layer setup is needed anyway, at least for Proton. Up to you with what you are comfortable with.

I'd prefer to keep it disabled by default, even if manual setup is required to set it up. I don't trust myself that it won't interfere with other stuff just yet, and once installed it would activate it in other games for no reason.

> * Also no strong preference here: may be easier to keep the environment variable for logging the same with nvapi/nvofapi, thus `DXVK_NVAPI_LOG_LEVEL` with `info` and `trace`. We can introduce e.g. `warn` if that fits.

I kind of like having it separate.

> * I think it makes sense to issue a clear log statement in `NvAPI_Vulkan_InitLowLatencyDevice` i.c.w. `VK_ERROR_EXTENSION_NOT_PRESENT`  to guide users to a missing/wrong layer setup.

Right, will do.

README.md
```diff
@@ -64,10 +68,33 @@ DXVK 1.10 and older does not support `DXVK_ENABLE_NVAPI`. Disable the `nvapiHack
...
+- `VkLayer_DXVK_NVAPI_reflex.json` (the layer manifest)
+- `libdxvk_nvapi_vkreflex_layer.so` (the layer library)
+
+To make it discoverable by your Vulkan loader:
+
+1. Do one of the following:
+    - Install the layer manifest into one of default implicit layer search paths (e.g. `${XDG_DATA_HOME}/vulkan/implicit_layer.d` where Steam usually installs its layers)
+    - Export full path to the directory containing the layer manifest in `VK_ADD_IMPLICIT_LAYER_PATH` environment variable (e.g. `VK_ADD_IMPLICIT_LAYER_PATH="${HOME}/dxvk-nvapi/layer"`)
+        - Note that `VK_ADD_IMPLICIT_LAYER_PATH` works only with Vulkan loader v1.3.296 or newer
+2. If needed, edit the layer manifest to ensure that its `library_path` points to the layer library
```
jp7677, line 88: May be we can make this a sub point of `Install the layer manifest into one of default implicit layer search paths` since this is only needed in a non default build setup or when the manifest is no longer in the same directory as the layer library.

README.md
```diff
@@ -64,10 +68,33 @@ DXVK 1.10 and older does not support `DXVK_ENABLE_NVAPI`. Disable the `nvapiHack
...
+     - For example, if the layer library is in the same directory as the layer manifest, the path should be `./libdxvk_nvapi_vkreflex_layer.so`
+     - For other scenarios where the layer library is nearby, relative paths work as well (e.g. `../../../lib/libdxvk_nvapi_vkreflex_layer.so`)
+     - It's also possible to provide full path to the layer library (e.g. `/home/user/dxvk-nvapi/layer/libdxvk_nvapi_vkreflex_layer.so`)
+     - Finally, if the layer is in one of the paths searched by dynamic linker (like `/usr/lib`, this depends on the distro), it's possible to set the library path to just the name of the layer library: `"library_path": "libdxvk_nvapi_vkreflex_layer.so"`
+
+Run `vulkaninfo` and check the list of discovered layers to verify there is an entry named `VK_LAYER_DXVK_NVAPI_reflex (DXVK-NVAPI Vulkan Reflex compatibility layer)`.
```
jp7677, line 94: I found `vulkaninfo` a but tricky for validating the layer setup, since `vulkaninfo` shows that the manifest has been found, but not that the `library_path` inside the manifest is correct. May be add the following sentence 

```
Note that `vulkaninfo` does not validate the correctness of the `library_path` property inside the manifest.
```

or something like this.

README.md
```diff
@@ -89,9 +116,20 @@ The following environment variables tweak DXVK-NVAPI's runtime behavior:
...
 - `DXVK_NVAPI_LOG_PATH` enables file logging additionally to console output and sets the path where the log file `nvapi.log`/`nvapi64.log`/`nvofapi64.log` should be written to. Log statements are appended to an existing file. Please remove this file once in a while to prevent excessive grow. This requires `DXVK_NVAPI_LOG_LEVEL` set to `info` or `trace`.
 
+The following environment variables tweak DXVK-NVAPI's Vulkan Reflex layer's runtime behavior:
+
+- `DXVK_NVAPI_VKREFLEX=1` enables the layer
+- `DISABLE_DXVK_NVAPI_VKREFLEX` disables the layer (overriding `DXVK_NVAPI_VKREFLEX=1`) when set to any nonempty value
+- `DXVK_NVAPI_VKREFLEX_LAYER_LOG_LEVEL` configures logging verbosity, supported values are `none` (default), `error`, `warn`, `info`, `trace`, `debug` and integer values from `0` to `5` inclusive that correspond to named log levels
+- `DXVK_NVAPI_VKREFLEX_EXPOSE_EXTENSION=1` causes the layer to report support for `VK_NV_low_latency` device extension, this is usually unnecessary as modern NVIDIA drivers already claim support for it and exposing it from the layer can cause issues in some games (notably Indiana Jones and the Great Circle)
```
jp7677, line 124: May be add a sentence that the following env vars are just meant for troubleshooting and setting them is not needed/ strongly discouraged  in a regular setup.

README.md
```diff
@@ -89,9 +116,20 @@ The following environment variables tweak DXVK-NVAPI's runtime behavior:
 - `DXVK_NVAPI_LOG_LEVEL` set to `info` prints log statements. The default behavior omits any logging. Please fill an issue if using log servery `info` creates log spam. Setting severity to `trace` logs all entry points enter and exits, this has a severe effect on performance. All other log levels will be interpreted as `none`.
 - `DXVK_NVAPI_LOG_PATH` enables file logging additionally to console output and sets the path where the log file `nvapi.log`/`nvapi64.log`/`nvofapi64.log` should be written to. Log statements are appended to an existing file. Please remove this file once in a while to prevent excessive grow. This requires `DXVK_NVAPI_LOG_LEVEL` set to `info` or `trace`.
 
+The following environment variables tweak DXVK-NVAPI's Vulkan Reflex layer's runtime behavior:
+
+- `DXVK_NVAPI_VKREFLEX=1` enables the layer
```
jp7677, line 121: For safety, even if this is mentioned earlier:

```
.. ` , when not set, the layer is disabled.`
```

jp7677 (commented):
Thanks for adding the readme additions. See some minor suggestions.

Once the remaining fixup commits are squashed, this PR LGTM!

README.md
```diff
@@ -64,10 +68,33 @@ DXVK 1.10 and older does not support `DXVK_ENABLE_NVAPI`. Disable the `nvapiHack
...
+The Vulkan Reflex layer, when built as a Linux-side layer, is not installed into Wine prefix. Instead it's installed like any other Vulkan implicit layer. It is distributed as a pair of files:
+
+- `VkLayer_DXVK_NVAPI_reflex.json` (the layer manifest)
+- `libdxvk_nvapi_vkreflex_layer.so` (the layer library)
+
+To make it discoverable by your Vulkan loader:
+
+1. Do one of the following:
+    - Install the layer manifest into one of default implicit layer search paths (e.g. `${XDG_DATA_HOME}/vulkan/implicit_layer.d` where Steam usually installs its layers)
+    - Export full path to the directory containing the layer manifest in `VK_ADD_IMPLICIT_LAYER_PATH` environment variable (e.g. `VK_ADD_IMPLICIT_LAYER_PATH="${HOME}/dxvk-nvapi/layer"`)
```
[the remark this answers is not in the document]
Saancreed: Sure, will do. I'm also wondering if I should show how multiple paths can be provided in `VK_ADD_IMPLICIT_LAYER_PATH` in case the user already has that exported with a path to some other layer manifest but it's probably just unnecessary complication.

README.md
Saancreed: Fair enough, but it does cause the loader to emit an error if layer was requested but the path was invalid. I'll describe this case in greater detail.

README.md
Saancreed: Sure, will do.

Saancreed:
I also added a notice about potential glibc version mismatches because our CI is running on Arch, so less surprises for Ubuntu users is probably a good idea :penguin:

README.md
```diff
@@ -64,10 +68,37 @@ DXVK 1.10 and older does not support `DXVK_ENABLE_NVAPI`. Disable the `nvapiHack
...
+Using DXVK-NVAPI with other GPU vendors / drivers has very limited use. Outside of testing, only Reflex (LatencyFlex) or HDR entry points provide benefits. This requires DXVK to see the GPU as an NVIDIA GPU. Use `DXVK_CONFIG="dxgi.hideAmdGpu = True"` to spoof an AMD GPU as NVIDIA GPU. Use `DXVK_CONFIG="dxgi.customVendorId = 10de"` for generally spoofing an NVIDIA GPU.
 
 Setting `DXVK_NVAPI_ALLOW_OTHER_DRIVERS=1` is needed for successful DXVK-NVAPI initialization when using a driver other than the NVIDIA proprietary driver or Mesa NVK. The reported driver version on drivers other than the NVIDIA proprietary driver will be 999.99. Overriding the reported driver version is still recommended. The reported GPU arrchitecture for other vendors is always Pascal to prevent attempts to initialize DLSS. This behavior cannot be changed without modifying the source code.
 
+### Vulkan Reflex layer
+
+The Vulkan Reflex layer, when built as a Linux-side layer, is not installed into Wine prefix. Instead it's installed like any other Vulkan implicit layer. It is distributed as a pair of files:
```
jp7677, line 77: … into `a` Wine prefix

README.md
```diff
@@ -64,10 +68,37 @@ DXVK 1.10 and older does not support `DXVK_ENABLE_NVAPI`. Disable the `nvapiHack
...
+To make it discoverable by your Vulkan loader:
+
+1. Do one of the following:
+    - Export full path to the directory containing the layer manifest in `VK_ADD_IMPLICIT_LAYER_PATH` environment variable (e.g. `VK_ADD_IMPLICIT_LAYER_PATH="${HOME}/dxvk-nvapi/layer"`)
+        - Note that `VK_ADD_IMPLICIT_LAYER_PATH` works only with Vulkan loader v1.3.296 or newer
+        - If you already have `VK_ADD_IMPLICIT_LAYER_PATH` set and pointing to a directory with some other implicit layer manifest, it is possible to pass multiple paths by separating them with `:`
+    - Install the layer manifest into one of default implicit layer search paths (e.g. `${XDG_DATA_HOME}/vulkan/implicit_layer.d` where Steam usually installs its layers)
+2. If needed, edit the layer manifest to ensure that its `library_path` points to the layer library
+     - For example, if the layer library is in the same directory as the layer manifest, the path should be `./libdxvk_nvapi_vkreflex_layer.so`
+         - Note that the distrbuted layer manifest by default assumes this scenario; therefore there is no need to edit the manifest if the layer library is kept in the same directory as the manifest file
```
jp7677, line 91: `distributed`

jp7677 (commented):
2 typos

README.md
Saancreed: Thanks, fixed.

jp7677 (approved):
LGTM!

jp7677 (approved):

PR #238, opened 2024-12-25
nvapi: Rename Vulkan → Vk and rework its construction

Saancreed:
Also gets rid of our custom `VkStructure` used in tests in favor of `VkBaseOutStructure` provided by Vulkan headers.

`MockFactory` currently does not override new factory overload, I left this for later, when it's actually used in tests.

Should I also rename `ResourceFactory::CreateVulkan` to `ResourceFactory::CreateVk`?

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -11,7 +11,7 @@ nvapi_src = files([
   'nvapi/nvapi_adapter_registry.cpp',
   'nvapi/nvapi_adapter.cpp',
   'nvapi/nvapi_output.cpp',
-  'nvapi/vulkan.cpp',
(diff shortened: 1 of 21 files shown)


jp7677 (approved):
LGTM! Feel free to land this.

Suggestion: add the reasoning for this (I mean working around a meson limitation that’s preventing introducing `nvapi_vulkan.cpp`) to the commit message.

Wrt the factory methods, I’m fine with both versions. So up to you to rename or keep the current name.

Saancreed:
Explanation added, and I think I'll keep the longer name, at least for now.

PR #239, opened 2025-01-07
Add GB20x Support to NvAPI_GPU_GetArchInfo

liam-middlebrook:


jp7677:
@liam-middlebrook I took the liberty to take over in #240 and add a few commits on top of yours because the testing part required some preparations. Would you mind doing a review on that newer PR, esp. the SM part?

PR #240, opened 2025-01-07
Add GB20x Support to NvAPI_GPU_GetArchInfo (follow up)

jp7677:
Follow up on #239 

cc: @liam-middlebrook

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ The following environment variables tweak DXVK-NVAPI's runtime behavior:
   - `TU100` (Turing)
   - `GA100` (Ampere)
   - `AD100` (Ada)
+  - `GB200` (Blackwell)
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -86,6 +86,7 @@ The following environment variables tweak DXVK-NVAPI's runtime behavior:
   - `TU100` (Turing)
   - `GA100` (Ampere)
   - `AD100` (Ada)
+  - `GB100` (Blackwell)
```
Saancreed, line 89: This is incorrect, we are actually looking for `GB200`, not `GB100`.

jp7677: Thanks, rebased again on top of the original PR which includes the correct line.

Saancreed (approved):
:ok_hand:

Still not sure if we should keep the compute capabillity hardcoded fallback or not. But at worst we can bring it back later.

PR #243, opened 2025-01-18
includes: Remove d3d12 header copy

jp7677:
Newer MinGW headers are good enough again.

Let see if CI passes everywhere...


mati865:
I do not know where you are getting mingw-w64 toolchain from but for this change to work you need at least mingw-w64 v12 from last year: https://github.com/mingw-w64/mingw-w64/commit/c9af10e9d145a62e1db90ebe441a621ec034b7af

Is it coming from https://github.com/ValveSoftware/Proton/blob/1e52f95633c5145aaf54bc27476021d9e1baade2/docker/Makefile#L14? GCC version does match.

jp7677:
Yes, as CI shows, recent build chains are fine, but we indeed have to wait until Proton gets an updated build chain to not break Proton builds.

PR #244, opened 2025-01-20
Drop repeated frame IDs

jp7677:
Not tested ...

Based on https://github.com/HansKristian-Work/vkd3d-proton/issues/2294

### DIFF
--- a/src/nvapi/nvapi_d3d_low_latency_device.cpp
+++ b/src/nvapi/nvapi_d3d_low_latency_device.cpp
@@ -127,23 +127,23 @@ namespace dxvk {
         static_assert(static_cast<int>(OUT_OF_BAND_PRESENT_END) - 1 == static_cast<int>(VK_LATENCY_MARKER_OUT_OF_BAND_PRESENT_END_NV));
 
         switch (markerType) {
-            case (SIMULATION_START):
-            case (SIMULATION_END):
-            case (RENDERSUBMIT_START):
-            case (RENDERSUBMIT_END):
-            case (PRESENT_START):
-            case (PRESENT_END):
-            case (INPUT_SAMPLE):
-            case (TRIGGER_FLASH):
+            case SIMULATION_START:
+            case SIMULATION_END:
+            case RENDERSUBMIT_START:
+            case RENDERSUBMIT_END:
+            case PRESENT_START:
+            case PRESENT_END:
+            case INPUT_SAMPLE:
+            case TRIGGER_FLASH:
                 return markerType;
             // VkLatencyMarkerNV misses PC_LATENCY_PING and all following enum values are offset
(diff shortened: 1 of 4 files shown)


src/nvapi/nvapi_d3d_low_latency_device.h
```diff
@@ -40,12 +40,14 @@ namespace dxvk {
       private:
         inline static std::unordered_map<IUnknown*, ID3DLowLatencyDevice*> m_lowLatencyDeviceMap;
         inline static std::unordered_map<ID3DLowLatencyDevice*, std::unique_ptr<LowLatencyFrameIdGenerator>> m_frameIdGeneratorMap;
+        inline static std::map<std::pair<IUnknown*, uint64_t>, uint64_t> m_lastFrameIdMap;
```
Saancreed, line 43: Probably debatable, but I think instead of a map with pairs as keys, it would be a bit simpler if this was a map with 13-element arrays as values.

Saancreed (commented):
I'd move this to `LowLatencyFrameIdGenerator` (so something like `bool LowLatencyFrameIdGenerator::CheckRepeatedMarker(frameID, markerType)`) but the general idea appears to be sound.

src/nvapi/nvapi_d3d_low_latency_device.cpp
```diff
@@ -42,6 +42,15 @@ namespace dxvk {
         return m_applicationIdList[((lowLatencyDeviceFrameId - 1) % applicationIdListSize)];
     }
 
+    bool LowLatencyFrameIdGenerator::IsRepeatedFrame(uint64_t frameID, uint32_t markerType) {
+        if (m_lastFrameId.contains(markerType) && m_lastFrameId[markerType] == frameID) {
+            return true;
+        }
+
+        m_lastFrameId[markerType] = frameID;
+        return false;
```
Saancreed, line 51:
```suggestion
        return std::exchange(m_lastFrameId[markerType], frameID) == frameID;
```

While this does have different behavior for `frameID == 0`, we can fix it by changing type of `m_lastFrameId` to `std::unordered_map<uint32_t, std::optional<uint64_t>>` (some of our tests rely on this).

But I still believe it should be an array instead of a map though :see_no_evil:

Saancreed (approved):
Even though it doesn't fix the issue, it's still a bug on our side so it's probably a good idea do this anyway…

(I'm not happy with the map but ultimately it's not important.)

Maybe we should ask Eric to take a look?

Saancreed (approved):
:ok_hand:

shelterx:
I'm a bit lost in the issues, Is this supposed to fix the reflex stuttering? I know vkd3d-proton got a draft workaround for it too.

jp7677:
> I'm a bit lost in the issues, Is this supposed to fix the reflex stuttering? I know vkd3d-proton got a draft workaround for it too (which it clearly says in this commit, did not see that)

No, as far as I'm aware this only fixes the log spam in VKD3D-Proton.

PR #252, opened 2025-01-30
Implement some DRS stuff

Saancreed:
Just an idea for now, let me know what you think about it.

### DIFF
--- a/src/nvapi_drs.cpp
+++ b/src/nvapi_drs.cpp
@@ -38,7 +38,12 @@ extern "C" {
         if (log::tracing())
             log::trace(n, log::fmt::hnd(hSession), log::fmt::ptr(profileName), log::fmt::ptr(phProfile));
 
-        return ProfileNotFound(str::format(n, " (", str::fromnvus(profileName), ")"));
+        if (!phProfile)
+            return InvalidArgument(n);
+
+        *phProfile = nvapiDrsProfile;
+
+        return Ok(n);
     }
 
     NvAPI_Status __cdecl NvAPI_DRS_FindApplicationByName(NvDRSSessionHandle hSession, NvAPI_UnicodeString appName, NvDRSProfileHandle* phProfile, NVDRS_APPLICATION* pApplication) {
@@ -47,7 +52,12 @@ extern "C" {
         if (log::tracing())
             log::trace(n, log::fmt::hnd(hSession), log::fmt::ptr(appName), log::fmt::ptr(phProfile), log::fmt::ptr(pApplication));
 
-        return ExecutableNotFound(str::format(n, " (", str::fromnvus(appName), ")"));
+        if (!phProfile)
+            return InvalidArgument(n);
+
+        *phProfile = nvapiDrsProfile;
+
+        return Ok(n);
     }
 
     NvAPI_Status __cdecl NvAPI_DRS_GetBaseProfile(NvDRSSessionHandle hSession, NvDRSProfileHandle* phProfile) {
--- a/src/nvapi_drs.cpp
+++ b/src/nvapi_drs.cpp
@@ -1,6 +1,8 @@
 #include "nvapi_private.h"
 #include "NvApiDriverSettings.c"
+#include "util/util_env.h"
 #include "util/util_statuscode.h"
+#include "util/util_string.h"
 
 static auto drs = 1U;
 static auto nvapiDrsSession = reinterpret_cast<NvDRSSessionHandle>(&drs);
@@ -120,9 +122,17 @@ extern "C" {
 
     NvAPI_Status __cdecl NvAPI_DRS_GetSetting(NvDRSSessionHandle hSession, NvDRSProfileHandle hProfile, NvU32 settingId, NVDRS_SETTING* pSetting) {
         constexpr auto n = __func__;
+        static const auto nvapiDrsSettingsString = dxvk::env::getEnvVariable("DXVK_NVAPI_DRS_SETTINGS");
+        static const auto nvapiDrsDwords = dxvk::str::parsedwords(nvapiDrsSettingsString);
 
         if (log::tracing())
             log::trace(n, log::fmt::hnd(hSession), log::fmt::hnd(hProfile), settingId, log::fmt::ptr(pSetting));
+
+        if (!pSetting)
+            return InvalidArgument(n);
+
+        if (pSetting->version != NVDRS_SETTING_VER1)
+            return IncompatibleStructVersion(n, pSetting->version);
 
         auto id = str::format("0x", std::hex, settingId);
         auto name = std::string("Unknown");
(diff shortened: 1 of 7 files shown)


jp7677:
Yeah, this is really cool! Thanks also for the research for the profiles.

What about doing the UI like this:

`DXVK_NVAPI_DRS_SETTINGS=setting1=type1:value1,setting2=type2:value2` whereas type can be something like `DW|dw`/`S|s`/`WS|ws`  corresponding to `NVDRS_SETTING_TYPE`. Ofc only `DW` for now. So your example would look like:

`DXVK_NVAPI_DRS_SETTINGSS=0x10E41DF3=DW:0xffffff,0x10E41DF7=DW:0xffffff`. Imho this env var name is less implementation specific and makes it easier to extend _if_ adding string types makes sense in the future.

Saancreed:
Well, I'm torn. On one hand, having just one variable for the configuration would be nice. On another, not only this would severly complicate the parsing process, but requiring the type for each value upfront would make usage much more awkward for end users.

(By the way, why is our `getEnvVar` implementation only allowing 260-chars long strings? Doesn't Windows allow up to 32k chars?)

Adding support for binary type would be quite simple, adding strings would be hell because at this point we'd probably need to ensure that `setting1=s:"weird value,stuff=s:\"with escaped\"quotation marks, commas and backslashes\\, just to confuse everyone"` is parsed correctly :weary:

jp7677:
Fair thoughts. If, on the other hand, chances are close to zero that we ever support strings, what about doing exactly what you proposed but still name the env var `DXVK_NVAPI_DRS_SETTINGS`. If we ever add binary, we can just use `b` as a prefix for the values. 

So:
`DXVK_NVAPI_DRS_SETTINGSS=0x10E41DF3=0xffffff,0x10E41DF7=0xffffff` and eventually in the future: `DXVK_NVAPI_DRS_SETTINGSS=0xABCDABCD=b<binary-somthing>`. Additionally postponing thinking about strings until there is a real use case, if ever.

Saancreed:
Sure, sounds good. I'll get this done later today.

src/nvapi_drs.cpp
```diff
@@ -47,7 +56,33 @@ extern "C" {
...
+                [[fallthrough]];
+            case NVDRS_APPLICATION_VER_V3:
+                pApplication->isCommandLine = 0;
+                pApplication->isMetro = 0;
+                [[fallthrough]];
+            case NVDRS_APPLICATION_VER_V2:
+                std::memset(pApplication->fileInFolder, 0, sizeof(pApplication->fileInFolder));
+                [[fallthrough]];
+            case NVDRS_APPLICATION_VER_V1:
+                std::memset(pApplication->launcher, 0, sizeof(pApplication->fileInFolder));
```
jp7677, line 74: /
```suggestion
                std::memset(pApplication->launcher, 0, sizeof(pApplication->launcher));
```

Result is the same, but still..

tests/nvapi_drs.cpp
```diff
@@ -50,6 +104,22 @@ TEST_CASE("DRS methods succeed", "[.drs]") {
...
+
+    SECTION("GetSetting with current struct version returns not incompatible-struct-version") {
+        NvDRSSessionHandle handle{};
+        NvDRSProfileHandle profile{};
+        NVDRS_SETTING setting;
+        setting.version = NVDRS_SETTING_VER;
+        REQUIRE(NvAPI_DRS_GetSetting(handle, profile, FXAA_ALLOW_ID, &setting) != NVAPI_INCOMPATIBLE_STRUCT_VERSION);
+    }
+
     SECTION("GetSetting returns setting-not-found") {
```
jp7677, line 123: Would be cool to have a test for `GetSetting` that uses `DXVK_NVAPI_DRS_SETTINGS`.  That said, I guess reading the env var needs to move to `NvAPI_DRS_GetSetting` (or e.g. `NvAPI_DRS_CreateSession`) for that. Since this is not really a hot path, I would be fine with the small penalty.

PS: Resetting env vars between tests happens in https://github.com/jp7677/dxvk-nvapi/blob/master/tests/section_listener.h#L11

src/util/util_string.cpp
```diff
@@ -44,4 +44,41 @@ namespace dxvk::str {
...
         return str ? std::string(str) : std::string();
     }
+
+    bool parsedword(const std::string_view& str, NvU32& value) {
+        auto end = str.data() + str.size();
+        std::from_chars_result result;
+
+        if (str.empty())
+            return false;
+        else if (str.starts_with("0x") || str.starts_with("0X"))
```
jp7677, line 54: not needed `else`.

src/util/util_string.cpp
```diff
@@ -44,4 +44,41 @@ namespace dxvk::str {
     std::string fromnullable(const char* str) {
         return str ? std::string(str) : std::string();
     }
+
+    bool parsedword(const std::string_view& str, NvU32& value) {
+        auto end = str.data() + str.size();
+        std::from_chars_result result;
+
+        if (str.empty())
```
jp7677, line 52: no real preference: this check can be done before declaring the variables.

jp7677 (commented):
See inline comments.

tests/nvapi_drs.cpp
Saancreed: And here I liked being able to make the map `static const`… :sweat_smile:

jp7677: If it’s ok move the declaration into NvAPI_DRS_GetSetting, static const would still be ok, as long as the env var is set before NvAPI_DRS_GetSetting is being tested. No need to have more than one variation in the tests (but a short comment would be nice noting that there is only one shot)

jp7677:
PS: We should also log when a key/value has been applied from the env var, may be also once the full env env var together with all successfully parsed pairs.

Saancreed:
> PS: We should also log when a key/value pair has been applied from the env var

That's kind of already done:

> `return Ok(str::format(n, " (", id, ": 0x", value, ")"));`

jp7677 (approved):
Looks cool, thanks!

I would suggest to also add something like this wrapped in `std::call_once`  for having some feedback if parsing was successful: 

```c++
        log::info(str::format(nvapiDrsSettingsEnvName, " is set, applying the following DRS settings when requested by the application:"));
        for (auto& [key, value] : nvapiDrsDwords)
            log::info(str::format("    0x", std::hex, key, " = 0x", std::hex, value));

```

May be we also need a `std::hex` also in `return Ok(str::format(n, " (", id, ": 0x", value, ")"));`. Also do we still want to add `name` to this `OK` statement? Eventually future NVAPI headers will contain the interesting setting names.

Saancreed:
@jp7677 Added:

```
98775.346:0020:0024:info:nvapi64:DXVK_NVAPI_DRS_SETTINGS is set, applying the following DRS settings when requested by the application (5 total):
98775.346:0020:0024:info:nvapi64:    0x10e41df7 = 0xffffff
98775.346:0020:0024:info:nvapi64:    0x10e41df3 = 0xffffff
98775.346:0020:0024:info:nvapi64:    0x10e41e03 = 0x1
98775.346:0020:0024:info:nvapi64:    0x10e41e02 = 0x1
98775.346:0020:0024:info:nvapi64:    0x10e41e01 = 0x1
```

Zenzi0:
Sorry to be a bother, a question if you don't mind. How do I use this? And have I understood correctly that this is for the new transformer model for Nvidia DLSS?

Do I simply use Proton Experimental Bleeding Edge and set DXVK_NVAPI_DRS_SETTINGS=1 in the Steam launch options? Does it somehow fetch the new .dll automatically? Or do we still need to replace them manually?
There has been some talk about achieving this functionality of the Nvidia App on Linux and as you can see I'm excited but also a bit confused about this. Thank you for any clarification.

Saancreed:
@Zenzi0

> Sorry to be a bother, a question if you don't mind. How do I use this?

Well, that depends on what you want to achieve.

> And have I understood correctly that this is for the new transformer model for Nvidia DLSS?

You can use it for this purpose among other things, yes.

> Do I simply use Proton Experimental Bleeding Edge and set DXVK_NVAPI_DRS_SETTINGS=1 in the Steam launch options?

For the former, yes. For the latter, no; you are required to provide _comma-separated list of `setting=value` pairs_, as [documented in the README](https://github.com/jp7677/dxvk-nvapi?tab=readme-ov-file#tweaks-debugging-and-troubleshooting). Please refer to [official public NVAPI headers](https://github.com/NVIDIA/nvapi/blob/main/NvApiDriverSettings.h) for a complete list of known settings and values. Here is one example [setting](https://github.com/NVIDIA/nvapi/blob/d08488fcc82eef313b0464db37d2955709691e94/NvApiDriverSettings.h#L211) and its [values](https://github.com/NVIDIA/nvapi/blob/d08488fcc82eef313b0464db37d2955709691e94/NvApiDriverSettings.h#L809-L827).

> Does it somehow fetch the new .dll automatically?

No. But there is another component that can do it: NGX updater.

> Or do we still need to replace them manually?

Depends. You can use NGX updater functionality built into NVIDIA's proprietary Linux driver to fetch new snippets and combine it with relevant `DXVK_NVAPI_DRS_SETTINGS` entries (namely,  `NGX_DLSS_SR_OVERRIDE`, `NGX_DLSS_FG_OVERRIDE` and/or `NGX_DLSS_RR_OVERRIDE`) to override snippets shipped with applications with ones fetched by NGX updater.

Or, if the updater fails and hangs your application (because it apparently doesn't support _all_ applications out there), just replace the snippet(s) as you suggested.

PR #253, opened 2025-02-01
nvofapi: NvOFAPICreateInstanceCuda - stub

SveSop:
This will return an error with a logtext indicating to upgrade DLSS version since the new DLSS4 version do not use opticalflow for DLSS anymore ref NVIDIA

### DIFF
--- a/src/nvofapi_cuda.cpp
+++ b/src/nvofapi_cuda.cpp
@@ -1,0 +1,17 @@
+#include "nvofapi_private.h"
+#include "util/util_log.h"
+#include "util/util_statuscode.h"
+
+extern "C" {
+
+    using namespace dxvk;
+
+    NV_OF_STATUS __cdecl NvOFAPICreateInstanceCuda(uint32_t apiVer, void* functionList) {
+        constexpr auto n = __func__;
+
+        if (log::tracing())
+            log::trace(n, apiVer, log::fmt::ptr(functionList));
+
+        return OFNotAvailable(n);
+    }
+}
--- a/src/meson.build
+++ b/src/meson.build
@@ -51,6 +51,7 @@ nvofapi_src = files([
   'nvofapi.cpp',
   'nvofapi_d3d12.cpp',
   'nvofapi_vk.cpp',
+  'nvofapi_cuda.cpp',
 ])
 
 # Only build 64-bit versions of nvofapi
--- a/src/nvofapi_entrypoints.h
+++ b/src/nvofapi_entrypoints.h
@@ -34,4 +34,7 @@ extern "C" {
     NV_OF_STATUS NVOFAPI OFSessionGetLastError(NvOFHandle hOf, char lastError[], uint32_t* size);
 
     NV_OF_STATUS NVOFAPI OFSessionGetCaps(NvOFHandle hOf, NV_OF_CAPS capsParam, uint32_t* capsVal, uint32_t* size);
+
+    // Overwrite CUDA entrypoint to avoid dependency to cuda.h
+    NV_OF_STATUS NVOFAPI NvOFAPICreateInstanceCuda(uint32_t apiVer, void* functionList);
 }
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -67,12 +67,14 @@ nvofapi_src = files([
   '../src/nvofapi.cpp',
   '../src/nvofapi_d3d12.cpp',
   '../src/nvofapi_vk.cpp',
+  '../src/nvofapi_cuda.cpp',
 ])
 
 nvofapi_tests_src = files([
   'nvofapi_main.cpp',
   'nvofapi_d3d12.cpp',
   'nvofapi_vk.cpp',
+  'nvofapi_cuda.cpp',
 ])
 
 target_name = 'nvofapi'+target_suffix+'-tests'
--- a/tests/nvofapi_cuda.cpp
+++ b/tests/nvofapi_cuda.cpp
@@ -1,0 +1,10 @@
+#include "nvofapi_tests_private.h"
+#include "nvofapi/mock_factory.h"
+
+using namespace trompeloeil;
+
+TEST_CASE("CreateInstanceVk returns not-available", "[.cuda]") {
+    struct NV_OF_CUDA_API_FUNCTION_LIST {
+    } functionList;
+    REQUIRE(NvOFAPICreateInstanceCuda(80, &functionList) == NV_OF_ERR_OF_NOT_AVAILABLE);
+}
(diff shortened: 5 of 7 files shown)


src/nvofapi_cuda.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "util/util_statuscode.h"
+#include "../version.h"
+#include "../config.h"
+
+extern "C" {
+
+    using namespace dxvk;
+
+    // CUDA entrypoint. Returning error with a message to upgrade DLSS version
+    NV_OF_STATUS __cdecl NvOFAPICreateInstanceCuda(uint32_t apiVer, void* functionList) {
```
jp7677, line 36: I suggest to do this without all the ceremony:

```c++
    NV_OF_STATUS NVOFAPI NvOFAPICreateInstanceCuda(uint32_t apiVer, void* functionList) {
        constexpr auto n = __func__;

        if (log::tracing())
            log::trace(n, apiVer, log::fmt::ptr(functionList));

        return UnsupportedFeature(n);
    }
```

src/nvofapi_cuda.cpp
```diff
@@ -0,0 +1,58 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jp7677, line 2: Lets leave this info out here since this file is a community contribution as part of this PR .

jp7677 (dismissed):
See comments, lets strip some of the ceremony . For completeness, could add a very brief corresponding `nvofapi_cuda.cpp` file like https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvofapi_vk.cpp

src/nvofapi_cuda.cpp
SveSop: I would prefer a "not supported" tho.. and would not hurt to create one for NvOFAPICreateInstanceD3D11 too.. just for completeness.. Instead of making suggestions to replace DLSS i mean.
Gives the USER a bit more understanding? :)

`log::info(str::format("OpticalFlow CUDA is not supported."));`

instead of just silently dropping?

SveSop: Well.. depends on the output of "UnsupportedFeature" i guess.. have not looked at the actual returntext when that happens? hmm...

jp7677: Returning `NV_OF_ERR_UNSUPPORTED_FEATURE` says exactly that we do not support this feature, this should be sufficient imho, no need to add more logging to that. Agree on doing the same for D3D11.

jp7677: Looking at the error code description, `NV_OF_ERR_OF_NOT_AVAILABLE` is even better:

```c++
    /**
    * This indicates that HW Optical flow functionality is not supported
    */
    NV_OF_ERR_OF_NOT_AVAILABLE,
```

jp7677: Ah, seeing the `void*` just now… that’s not a good idea imho, we should stick to the headers, but those have a cuda dependency, that’s not gonna work without some ugly things I prefer to avoid.

jp7677:
Not tested, but this is probably all that is needed (excluding comments why not having included the nvofapi_cuda header):

```diff
diff --git a/src/meson.build b/src/meson.build
index 2d35672..8e0b999 100644
--- a/src/meson.build
+++ b/src/meson.build
@@ -51,6 +51,7 @@ nvofapi_src = files([
   'nvofapi.cpp',
   'nvofapi_d3d12.cpp',
   'nvofapi_vk.cpp',
+  'nvofapi_cuda.cpp',
 ])
 
 # Only build 64-bit versions of nvofapi
diff --git a/src/nvofapi_cuda.cpp b/src/nvofapi_cuda.cpp
new file mode 100644
index 0000000..5f99d97
--- /dev/null
+++ b/src/nvofapi_cuda.cpp
@@ -0,0 +1,19 @@
+#include "nvofapi_private.h"
+#include "util/util_log.h"
+#include "util/util_statuscode.h"
+
+extern "C" {
+
+    using namespace dxvk;
+
+    NV_OF_STATUS NVOFAPI NvOFAPICreateInstanceCuda(uint32_t apiVer, void* functionList) {
+        constexpr auto n = __func__;
+
+        if (log::tracing())
+            log::trace(n, apiVer, log::fmt::ptr(functionList));
+
+        return OFNotAvailable(n);
+    }
+}
diff --git a/tests/meson.build b/tests/meson.build
index e4ba6ce..66bdc40 100644
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -73,6 +73,7 @@ nvofapi_tests_src = files([
   'nvofapi_main.cpp',
   'nvofapi_d3d12.cpp',
   'nvofapi_vk.cpp',
+  'nvofapi_cuda.cpp',
 ])
 
 target_name = 'nvofapi'+target_suffix+'-tests'
diff --git a/tests/nvofapi_cuda.cpp b/tests/nvofapi_cuda.cpp
new file mode 100644
index 0000000..59d73ec
--- /dev/null
+++ b/tests/nvofapi_cuda.cpp
@@ -0,0 +1,11 @@
+#include "nvofapi_tests_private.h"
+#include "nvofapi/mock_factory.h"
+
+#include "../src/nvofapi_cuda.cpp"
+
+using namespace trompeloeil;
+
+TEST_CASE("CreateInstanceVk returns not-available", "[.cuda]") {
+    struct NV_OF_CUDA_API_FUNCTION_LIST{} functionList;
+    REQUIRE(NvOFAPICreateInstanceCuda(80, &functionList) == NV_OF_ERR_OF_NOT_AVAILABLE);
+}
diff --git a/tests/nvofapi_main.cpp b/tests/nvofapi_main.cpp
index d62e6ab..5177908 100644
--- a/tests/nvofapi_main.cpp
+++ b/tests/nvofapi_main.cpp
@@ -3,6 +3,6 @@
 #include "nvofapi/section_listener.h"
 
 CATCH_REGISTER_TAG_ALIAS("[@unit-tests]", "[d3d12],[vk]")
-CATCH_REGISTER_TAG_ALIAS("[@all]", "[d3d12],[vk]")
+CATCH_REGISTER_TAG_ALIAS("[@all]", "[d3d12],[vk],[cuda]")
 
 CATCH_REGISTER_LISTENER(SectionListener)
```

But still very meh that it hacks around the actual header

SveSop:
`NV_OF_CUDA_API_FUNCTION_LIST` is not known unless you use the header as you probably discovered.

So.. non-elegant solution is to make a `nvofapi_cuda_private.h` perhaps and just declare these for usage, but then again, it would not use the ACTUAL correct struct.

Saancreed (approved):

jp7677 (approved):

PR #257, opened 2025-02-02
nvofapi: Only export used symbols

SveSop:
This will make the nvofapi64.dll binary export only the used symbols, and make it similar to the windows version.

The usage of __cdecl for the calls SHOULD not be needed, as the `nvOpticalFlowCommon.h` header spesifically sets this:

```
#ifdef _WIN32
#define NVOFAPI __stdcall
#else
#define NVOFAPI
#endif
```

Testing on the OpticalFlow SDK samples, it seems to work as intended. Lets have a talk about that :+1: 

The exported symbols is then like this when using winedump -j export

```
  Name:            nvofapi64.dll
  Characteristics: 00000000
  TimeDateStamp:   679FAE1F Sun Feb  2 18:40:47 2025
  Version:         0.00
  Ordinal base:    1
  # of functions:  5
  # of Names:      5
  Functions RVA:   0014B028
  Ordinals RVA:    0014B050
  Names RVA:       0014B03C

  Entry Pt  Ordn  Name
  00029DF0     1  NvOFAPICreateInstanceCuda
  00017150     2  NvOFAPICreateInstanceD3D11
  0001B250     3  NvOFAPICreateInstanceD3D12
  00027EF0     4  NvOFAPICreateInstanceVk
  00013B30     5  NvOFGetMaxSupportedApiVersion
```

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -63,5 +63,6 @@ if dxvk_cpu_family == 'x86_64'
     name_prefix         : '',
     dependencies        : [ lib_version ],
     include_directories : [ nvapi_headers, vk_headers ],
+    vs_module_defs      : target_name+'.def',
     install             : true)
 endif
--- a/src/nvofapi64.def
+++ b/src/nvofapi64.def
@@ -1,0 +1,7 @@
+LIBRARY nvofapi.dll
+EXPORTS
+    NvOFAPICreateInstanceCuda @1
+    NvOFAPICreateInstanceD3D11 @2
+    NvOFAPICreateInstanceD3D12 @3
+    NvOFAPICreateInstanceVk @4
+    NvOFGetMaxSupportedApiVersion @5
--- a/src/nvofapi_cuda.cpp
+++ b/src/nvofapi_cuda.cpp
@@ -6,7 +6,7 @@ extern "C" {
 
     using namespace dxvk;
 
-    NV_OF_STATUS __cdecl NvOFAPICreateInstanceCuda(uint32_t apiVer, void* functionList) {
+    NV_OF_STATUS NVOFAPI NvOFAPICreateInstanceCuda(uint32_t apiVer, void* functionList) {
         constexpr auto n = __func__;
 
         if (log::tracing())
--- a/src/nvofapi_d3d11.cpp
+++ b/src/nvofapi_d3d11.cpp
@@ -6,7 +6,7 @@ extern "C" {
 
     using namespace dxvk;
 
-    NV_OF_STATUS __cdecl NvOFAPICreateInstanceD3D11(uint32_t apiVer, NV_OF_D3D11_API_FUNCTION_LIST* functionList) {
+    NV_OF_STATUS NVOFAPI NvOFAPICreateInstanceD3D11(uint32_t apiVer, NV_OF_D3D11_API_FUNCTION_LIST* functionList) {
         constexpr auto n = __func__;
 
         if (log::tracing())
--- a/src/nvofapi_d3d12.cpp
+++ b/src/nvofapi_d3d12.cpp
@@ -126,7 +126,7 @@ extern "C" {
     }
 
     // ETBLs
-    NV_OF_STATUS __stdcall NvOFAPICreateInstanceD3D12(uint32_t apiVer, NV_OF_D3D12_API_FUNCTION_LIST* functionList) {
+    NV_OF_STATUS NVOFAPI NvOFAPICreateInstanceD3D12(uint32_t apiVer, NV_OF_D3D12_API_FUNCTION_LIST* functionList) {
         uint32_t apiVerMajor = (apiVer & 0xfffffff0) >> 4;
         uint32_t apiVerMinor = (apiVer & 0xf);
         constexpr auto n = __func__;


SveSop:
Hmm.. was `__stdcall` added like it was due to the msvc build job :thinking:

Saancreed:
I think you ran into the exact same thing I did when working on vkreflex layer: https://github.com/jp7677/dxvk-nvapi/commit/4a8b9522f632294302a8939fc5b004c6dccdc767

You need to ensure that both definition and declaration of each function have matching `__declspec(dllexport)`; if one has it and the other doesn't, MSVC will complain. Windows people usually use spec files to sidestep this nonsense and never have to bother with `__declspec(dllexport)` anywhere.

SveSop:
Lets see if this builds then.. Probably needs some more tests 😄 

Added benefit of using a .def file is ofc that you can control ordinal number in those cases where that could be important.

jp7677:
Thanks! Assuming that this works for all CI pipelines, would be nice to change all NVOFAPI endpoints to `NV_OF_STATUS NVOFAPI` as part of this PR, currently it is inconsistent.

It would also be cool to have a separate Commit (or PR) for the NVAPI side, thus just creating a .def file and and removing `__declspec(dllexport)` from https://github.com/jp7677/dxvk-nvapi/blob/master/src/nvapi_interface.cpp#L42 .

Saancreed:
… actually, I just realized, why _don't_ `NvOFAPICreateInstance{Cuda,D3D11}` have `NVOFAPI`? I thought such a mistake shouldn't even compile…

jp7677:
> … actually, I just realized, why _don't_ `NvOFAPICreateInstance{Cuda,D3D11}` have `NVOFAPI`? I thought such a mistake shouldn't even compile…

It just resolves to `__stdcall` on windows. https://github.com/jp7677/dxvk-nvapi/blob/master/inc/nvofapi/nvOpticalFlowCommon.h#L48, using `__cdecl` instead as SveSop did for cuda and d3d11 should be fine. In fact, may be we should just drop `NVOFAPI` and use `__cdecl` everywhere? That would also be safe just in case someone wants to create a 32bit `nvofapi.dll`.

Saancreed:
Well yeah, `__stdcall` is very much not the same thing as `__cdecl`, unless we close our eyes and pretend that x64 is the only CPU arch we care about, that's a bug.

jp7677:
May be there is a way to detect the calling convention with NVIDIA's nvofapi64.dll for NvOFAPICreateInstanceVk for ruling out that this isn't a mistake in the headers?

SveSop:
Heh.. i did not actually notice that the headers did not use NVOFAPI for CreateInstanceVulkan.. How odd.

Saancreed:
> May be there is a way/tool to detect/see the calling convention with NVIDIA's nvofapi64.dll for NvOFAPICreateInstanceVk for ruling out that this isn't a mistake in the headers?

Nope, as Nvidia's 32-bit `nvofapi.dll` does not export `NvOFAPICreateInstanceVk` on Windows. Maybe they realized "eh, it doesn't matter" and didn't bother decorating it if it's supported on 64-bit only where there's only one calling convention anyway.

jp7677:
> > May be there is a way/tool to detect/see the calling convention with NVIDIA's nvofapi64.dll for NvOFAPICreateInstanceVk for ruling out that this isn't a mistake in the headers?
> 
> Nope, as Nvidia's 32-bit `nvofapi.dll` does not export `NvOFAPICreateInstanceVk` on Windows. Maybe they realized "eh, it doesn't matter" and didn't bother decorating it if it's supported on 64-bit only where there's only one calling convention anyway.

Thanks. Then indeed lets continue like you proposed.

SveSop:
Changed __cdecl -> NVOFAPI for d3d1x and cuda for consitency with the headers. Compiler resolves these the same anyway, but atleast it looks a bit tidier.

src/meson.build
```diff
@@ -63,5 +63,6 @@ if dxvk_cpu_family == 'x86_64'
     name_prefix         : '',
     dependencies        : [ lib_version ],
     include_directories : [ nvapi_headers, vk_headers ],
+    vs_module_defs      : 'nvofapi64.def',
```
jp7677, line 66: Let's reuse `target_name` here. So `target_name+'.def'`.

SveSop: Also changed the name to lowercase `nvofapi.dll` as that seems to be what is printed using dumpbin on windows for both nvofapi.dll and nvofapi64.dll

Just to attempt to make it somewhat consistent to the windows version.

jp7677 (approved):

src/meson.build
Saancreed: Are you sure? Winedump on my system (with 566.36 drivers) produces

```
Contents of /windows/Windows/System32/DriverStore/FileRepository/nvami.inf_amd64_60e995a01594d69e/nvofapi.dll: 505504 bytes


  Name:            nvofapi.dll
  Characteristics: 00000000
  TimeDateStamp:   FFFFFFFF Sun Feb  7 07:28:15 2106
  Version:         0.00
  Ordinal base:    1
  # of functions:  4
  # of Names:      4
  Functions RVA:   00072898
  Ordinals RVA:    000728B8
  Names RVA:       000728A8

  Entry Pt  Ordn  Name
  000039F0     1  NvOFAPICreateInstanceCuda
  00007AA0     2  NvOFAPICreateInstanceD3D11
  0000C4E0     3  NvOFAPICreateInstanceD3D12
  00004D70     4  NvOFGetMaxSupportedApiVersion

Done dumping /windows/Windows/System32/DriverStore/FileRepository/nvami.inf_amd64_60e995a01594d69e/nvofapi.dll
```

Saancreed: `NvOFAPICreateInstanceVk` is notably missing.

SveSop: This is for 64 bit that we make yeah... So if you dump the nvofapi64.dll file, you see the "Name: nvofapi.dll" is the same for both 64 and 32 bit dll in windows

IF we want to create a 32-bit in the future, we make a new nvofapi.def file that leaves out VK

Saancreed: Ah, wait, I see what you mean, that embedded library name is `nvofapi.dll` for both `nvofapi.dll` _and_ `nvofapi64.dll`. Please ignore what I said :sweat_smile:

Saancreed (approved):

PR #259, opened 2025-02-03
util: Avoid using std::string_view::starts_with

Saancreed:
Weird how GitHub Actions didn't catch it…


jp7677 (approved):

ivyl:
It does build now. Thanks! I'm poking at the actions to see why it succeeds with the release script but not when built via Proton's build system. I'll report back on that.

PR #260, opened 2025-02-03
includes: Update NVAPI headers to R570

Saancreed:


Saancreed:
Hm, actually, let me cleanup define of Blackwell arch now that it's in the headers…

jp7677 (approved):

PR #262, opened 2025-02-04
Fix trace logging before initialization

jp7677:


src/util/util_log.cpp
```diff
@@ -4,11 +4,17 @@
...
-static std::mutex fileStreamMutex;
-static bool traceEnabled = false;
-
 namespace dxvk::log {
+    constexpr auto logLevelEnvName = "DXVK_NVAPI_LOG_LEVEL";
+    constexpr auto logPathEnvName = "DXVK_NVAPI_LOG_PATH";
+    constexpr auto logFileName = DXVK_NVAPI_TARGET_NAME ".log";
+
+    const static auto logLevel = env::getEnvVariable(logLevelEnvName);
+    const static auto traceEnabled = logLevel == "trace";
```
Saancreed, line 13:
```suggestion
    static const auto logLevel = env::getEnvVariable(logLevelEnvName);
    static const auto traceEnabled = logLevel == "trace";
```

Saancreed (approved):
Just one code-style (I think?) related nitpick.

Plus, I suppose we can now trace `nvapi_QueryInterface`.

jp7677:
> Plus, I suppose we can now trace `nvapi_QueryInterface`.

Yep, that works now, added. Thanks for the reminder!

PR #264, opened 2025-02-05
nvapi: Allow switching NGX registry settings

jp7677:
Example:

~`DXVK_NVAPI_SET_DLSS_INDICATOR=1024 DXVK_NVAPI_SET_DLSSG_INDICATOR=2`~

~`DXVK_NVAPI_SET_NGX_INDICATOR=DLSS=1024,DLSSG=2`~

`DXVK_NVAPI_SET_NGX_DEBUG_OPTIONS=DLSSIndicator=1024,DLSSGIndicator=2,Logging=1,LogLevel=1`

Probably out of scope...

Reference:
- https://github.com/NVIDIA/DLSS/tree/main/utils
- https://github.com/NVIDIAGameWorks/Streamline/blob/main/docs/ProgrammingGuideDLSS_G.md#200-dlss-fg-indicator-text

### DIFF
--- a/src/nvapi.cpp
+++ b/src/nvapi.cpp
@@ -5,6 +5,7 @@
 #include "util/util_string.h"
 #include "util/util_env.h"
 #include "util/util_log.h"
+#include "util/util_ngx_debug.h"
 #include "../version.h"
 #include "../config.h"
 
@@ -380,6 +381,10 @@ extern "C" {
             --initializationCount;
             return NvidiaDeviceNotFound(n);
         }
+
+#if _WIN64
+        SetNgxDebugOptions(); // NGX is 64-bit only
+#endif
 
         return Ok(n);
     }
--- a/src/util/util_ngx_debug.h
+++ b/src/util/util_ngx_debug.h
@@ -1,0 +1,56 @@
+#pragma once
+
+#include "../nvapi_private.h"
+#include "util_log.h"
+#include "util_env.h"
+
+namespace dxvk {
+    inline void SetNgxDebugOptions() {
+        static const auto setNgxIndicatorEnvName = "DXVK_NVAPI_SET_NGX_DEBUG_OPTIONS";
+        static const std::unordered_map<std::string_view, std::string_view> settings = {
+            {"EnableConsoleLogging", "EnableConsoleLogging"},
+            {"LogLevel", "LogLevel"},
+            {"ShowDlssIndicator", "ShowDlssIndicator"},
+            {"DLSSG_IndicatorText", "DLSSG_IndicatorText"},
+            {"Logging", "EnableConsoleLogging"},
+            {"DLSSIndicator", "ShowDlssIndicator"},
+            {"DLSSGIndicator", "DLSSG_IndicatorText"},
+        };
+        static const auto setNgxIndicator = env::getEnvVariable(setNgxIndicatorEnvName);
+
+        if (setNgxIndicator.empty())
+            return;
+
+        std::set<std::string_view, str::CaseInsensitiveCompare<std::string_view>> keys;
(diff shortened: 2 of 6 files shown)


SveSop:
Not a huge fan of this idea tbh... but... meh.

adamdmoss:
I was a bit doubtful that this belonged in dxvk-nvapi until I suddenly had a need for this.  :grin:

jp7677:
> I was a bit doubtful that this belonged in dxvk-nvapi until I suddenly had a need for this. 😁

Yeah, same thoughts here :). It is somewhat out of scope, but very convenient. 
I’m leaning towards getting this in, but haven’t made up my mind yet. Protontricks can achieve the same, but it’s not as elegantly as just setting an env var.

Fox2Code:
Well, I think what is making this PR a bit awkward is the registry change made from environment variable are persistent.

Maybe adding something to reset all registry values could be useful?

Something like `DXVK_NVAPI_SET_NGX_DEBUG_OPTIONS=defaults`?

jp7677:
> Maybe adding something to reset all registry values could be useful?
> Something like `DXVK_NVAPI_SET_NGX_DEBUG_OPTIONS=defaults`?

Using `DXVK_NVAPI_SET_NGX_DEBUG_OPTIONS=DLSSIndicator=0,DLSSGIndicator=0,Logging=0,LogLevel=0` does exactly this. Tbh, I don't want to put any more magic in this. Agree though that we should document clearly tha the `SET` means that those values will be persistently set in the registry and that there is no "unset".

LightTreasure:
Just tried this in Cyberpunk 2077 and for me the game's post-processing filters (for example, motion blur) apply to the indicator as well, making it harder to read. Have you encountered this?

jp7677:
> Just tried this in Cyberpunk 2077 and for me the game's post-processing filters (for example, motion blur) apply to the indicator as well, making it harder to read. Have you encountered this?

Yes, this is just how the DLSS indicators work. Note that those are just debug tools from NVIDIA's DLSS/NGX/Streamline SDK and this PR only sets the registry keys. This is not a HUD meant for regular usage.

LightTreasure:
@jp7677 Might be good to mention that in the README.

src/nvapi.cpp
```diff
@@ -381,6 +382,8 @@ extern "C" {
             return NvidiaDeviceNotFound(n);
         }
 
+        SetNgxDebugOptions();
```
Saancreed, line 385: Since NGX is 64-bit only but we are producing both `nvapi64.dll` and 32-bit `nvapi.dll`, I'd surround this with some kind of `#if defined(_WIN64) … #endif` to avoid polluting `WoW64Node` branches of registry.

Saancreed (approved):
Only one small suggestion.

src/util/util_ngx_debug.h
```diff
@@ -0,0 +1,56 @@
...
+        const static std::unordered_map<std::string_view, std::string_view> settings = {
+            {"EnableConsoleLogging", "EnableConsoleLogging"},
+            {"LogLevel", "LogLevel"},
+            {"ShowDlssIndicator", "ShowDlssIndicator"},
+            {"DLSSG_IndicatorText", "DLSSG_IndicatorText"},
+            {"Logging", "EnableConsoleLogging"},
+            {"DLSSIndicator", "ShowDlssIndicator"},
+            {"DLSSGIndicator", "DLSSG_IndicatorText"},
+        };
+        const static auto setNgxIndicator = env::getEnvVariable(setNgxIndicatorEnvName);
```
Saancreed, line 19: Or, actually, one more code-style related: `const static` → `static const`.

PR #266, opened 2025-02-05
Update NVML/Vulkan headers

jp7677:
Changes nothing, just a feel good PR.


Saancreed (approved):
Sure, but why not update VK headers all the way to 307?

jp7677:
> Sure, but why not update VK headers all the way to 307?

No real reason other than that this tag matches a Vulkan SDK release.

PR #267, opened 2025-02-07
nvapi-drs: Copy only initialized bytes

jp7677:
Fixes startup crash in OW2.


Saancreed (approved):
I don't quite understand why does this fix anything. If the contract of `NvAPI_UnicodeString` is that it's supposed to be null-terminated sequence of up to 2047 UTF-16 code units plus `L'\0'` then just copying 4096 bytes we got as input _shouldn't_ cause any issues and the application _should_ use `wcslen(appName)` (either explicitly or implicitly as you did with `std::wstring` constructor) to read the output, ignoring bytes past the first `L'\0'`.

And even if `NvAPI_UnicodeString` was supposed to be padded with zeros, this is what we would have received as input, and again `memcpy` would be correct. So, what am I missing? Which invariant are we violating?

(Not that I like games being broken so this is good to go but I really want to understand the underlying issue.)

jp7677:
I've started with this simple snippet:

```c++
        for (auto i = 0; i < NVAPI_UNICODE_STRING_MAX; i++ ) {
            log::info(str::format(i, ": ", appName[i]));
        }
``` 

which showed valid things up until position 75 (which matches `Z:\home\-------\.local\share\Steam\steamapps\common\Overwatch\Overwatch.exe`). After that zeros and random things up until position 1663 where execution stopped/crashed, because I guess reading from an invalid adress.

I think the passed `appName` wasn't declared as `NvAPI_UnicodeString` but as a NvU16 (or similar) array with a shorter size.

PR #269, opened 2025-02-10
DRS (NGX) settings for human beings

Saancreed:
Please forgive the macro horrors, C++ truly lacks some kind of compile-time reflection…

TODO: tests and README updates…

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,19 @@ The following environment variables tweak DXVK-NVAPI's runtime behavior:
 - `DXVK_NVAPI_ALLOW_OTHER_DRIVERS`, when set to `1`, allows DXVK-NVAPI to initialize for a non-NVIDIA GPU.
 - `DXVK_NVAPI_DISABLE_ENTRYPOINTS`, when set to a comma-separated list of NVAPI entrypoint names, will hide their implementation from the application. For example, `DXVK_NVAPI_DISABLE_ENTRYPOINTS=NvAPI_D3D11_BeginUAVOverlap,NvAPI_D3D11_EndUAVOverlap` will report D3D11 barrier control extensions as not available.
 - `DXVK_NVAPI_DRS_SETTINGS` allows providing custom values for arbitrary driver settings. Currently, only values of DWORD (u32) type are supported. Format is `setting1=value1,setting2=value2,…` where both `settingN` and `valueN` are unsigned 32-bit integers and can be prefixed with `0x` or `0X` to parse them as hexadecimal numbers. For example, `DXVK_NVAPI_DRS_SETTINGS=0x10E41DF3=0xffffff,0x10E41DF7=0xffffff` overrides presets for both DLSS Super Resolution and DLSS Ray Reconstruction to latest.
+  - Additionally, the following settings are supported in their named/text form:
+    - `NGX_DLAA_OVERRIDE`
+    - `NGX_DLSSG_MULTI_FRAME_COUNT`
+    - `NGX_DLSS_FG_OVERRIDE`
+    - `NGX_DLSS_RR_MODE`
+    - `NGX_DLSS_OVERRIDE_OPTIMAL_SETTINGS`
+    - `NGX_DLSS_RR_OVERRIDE`
+    - `NGX_DLSS_RR_OVERRIDE_RENDER_PRESET_SELECTION`
+    - `NGX_DLSS_SR_MODE`
+    - `NGX_DLSS_SR_OVERRIDE`
+    - `NGX_DLSS_SR_OVERRIDE_RENDER_PRESET_SELECTION`
+  - The corresponding values can be deduced from the [NVAPI `NvApiDriverSettings.h` header](https://github.com/NVIDIA/nvapi/blob/d08488fcc82eef313b0464db37d2955709691e94/NvApiDriverSettings.h#L700) with the setting type prefix omitted, for example `DXVK_NVAPI_DRS_SETTINGS=NGX_DLSS_FG_OVERRIDE=on`. Both setting name (`NGX_DLSS_FG_OVERRIDE`) and value (`on`) are case-insensitive.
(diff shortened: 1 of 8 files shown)


Saancreed:
Should be good to go now… can't wait for ProtonDB to be flooded with brand new flavor-of-the-month launch options that we'll have to keep debugging for years to come :weary:

README.md
```diff
@@ -109,6 +109,22 @@ The following environment variables tweak DXVK-NVAPI's runtime behavior:
...
 - `DXVK_NVAPI_DRS_SETTINGS` allows providing custom values for arbitrary driver settings. Currently, only values of DWORD (u32) type are supported. Format is `setting1=value1,setting2=value2,…` where both `settingN` and `valueN` are unsigned 32-bit integers and can be prefixed with `0x` or `0X` to parse them as hexadecimal numbers. For example, `DXVK_NVAPI_DRS_SETTINGS=0x10E41DF3=0xffffff,0x10E41DF7=0xffffff` overrides presets for both DLSS Super Resolution and DLSS Ray Reconstruction to latest.
+  - Additionally, _some_ well-known settings are supported in their named/text form. For each setting supported this way, `settingN` can be provided as case-insensitive name of [`ESetting`](https://github.com/NVIDIA/nvapi/blob/d08488fcc82eef313b0464db37d2955709691e94/NvApiDriverSettings.h#L148) enum member with the `_ID` suffix omitted, and `valueN` can be provided as case-insensitive name of corresponding `EValues_${settingN}` enum member with the longest repetition of setting type prefix omitted, for example `NGX_DLSS_FG_OVERRIDE=on`. Settings currently supported this way are:
```
jp7677, line 112: What about keeping the note brief here and refer to a new page in our wiki. There we can write it out in all variations and even include some context and best practices, if any. Linking to the sources to figure out the setting should work for developers, but i have the feeling this is getting used by a broader audience. The wiki also makes it imho also more accessible for quickly editing/extending the info. 

So something like this? 

- Additionally, _some_ well-known settings are supported in their named/text form. See [Custom driver settings](https://github.com/jp7677/dxvk-nvapi/wiki/Custom-driver-settings) for full instructions.

src/nvapi_drs.cpp
```diff
@@ -149,18 +150,19 @@ extern "C" {
...
         static std::once_flag once;
         std::call_once(once, []() {
-            if (nvapiDrsSettingsString.empty())
+            if (nvapiDrsDwords.empty())
                 return;
 
-            log::info(str::format(nvapiDrsSettingsEnvName, " is set, applying the following DRS settings when requested by the application (", nvapiDrsDwords.size(), " total):"));
+            log::info(str::format("Applying the following DRS settings when requested by the application (", nvapiDrsDwords.size(), " total):"));
             for (auto& [key, value] : nvapiDrsDwords)
                 log::info(str::format("    0x", std::hex, key, "/", GetSettingName(key), " = 0x", std::hex, value));
```
jp7677, line 167: Probably for another time, but would be cool to also translate the values for the 10 wekk-known settings into their human readable versions.

src/util/util_drs.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+        for (auto entry : entries) {
+            auto eq = entry.find('=');
+
+            if (eq == entry.npos || eq == 0 || eq == entry.size() - 1)
+                continue;
+
+            NvU32 key, value;
+
+            if (parsedrssetting(entry.substr(0, eq), key) && parsedrsdwordvalue(key, entry.substr(eq + 1), value))
```
jp7677, line 42: Since this is now specific to `drs`, we could log/trace parsing errors for key/value here. That said, we already show the summary, so this would be quite verbose. So no real preference on my side.

jp7677 (commented):
LGTM code wise, see suggestion about documentation.

README.md
Saancreed: On the other hand, it also makes it a bit harder to keep versioned docs in sync with what is actually supported in code. I can certainly see the appeal of a dedicated wiki page but I think it should be an extension of core information included in the README and not a replacement.

That said, I did have a hard time figuring out how detailed the README should be in this area. If you have an idea what to include in the wiki page and what isn't critical enough to keep it here (examples?), feel free to propose something. Or let's tackle this some other time.

src/nvapi_drs.cpp
Saancreed: Sure, sounds like a nice idea.

src/util/util_drs.cpp
Saancreed: I have a slight preference towards avoiding log statements in "util" code, and yeah, logging what was parsed successfully is probably good enough.

README.md
jp7677: I think we will have a hard time explaining this briefly since imho it needs to be written out for non-technical users. Directing to the source code with a guide will I guess yield a lot of typos and misunderstandings. But sure, we can introduce a wiki page later and then clean things up here.

What about like this for now?

 - Additionally, the following settings are supported in their named/text form:
    - `NGX_DLAA_OVERRIDE`
    - `NGX_DLSSG_MULTI_FRAME_COUNT`
    - `NGX_DLSS_FG_OVERRIDE`
    - `NGX_DLSS_RR_MODE`
    - `NGX_DLSS_OVERRIDE_OPTIMAL_SETTINGS`
    - `NGX_DLSS_RR_OVERRIDE`
    - `NGX_DLSS_RR_OVERRIDE_RENDER_PRESET_SELECTION`
    - `NGX_DLSS_SR_MODE`
    - `NGX_DLSS_SR_OVERRIDE`
    - `NGX_DLSS_SR_OVERRIDE_RENDER_PRESET_SELECTION`

    The corresponding values can be deducted from the [NVAPI NvApiDriverSettings.h header](https://github.com/NVIDIA/nvapi/blob/d08488fcc82eef313b0464db37d2955709691e94/NvApiDriverSettings.h#L700) with the setting type prefix omitted, for example `NGX_DLSS_FG_OVERRIDE=on`. Values are treated case-insensitive.

- Furthermore, each supported named setting can also be configured with a dedicated environment variable ...

Saancreed: Done, should be a bit better now. Let me know if the last point needs further simplification :sweat_smile:

jp7677 (approved):
All good, thanks!

PR #273, opened 2025-02-16
util: Support variables of any length in getEnvVariable

Saancreed:
I wish we could just use `GetEnvironmentVariableA` but it's not exactly trustworthy when non-UTF8 locales could cause it to return nonsense.


jp7677 (approved):
Thanks. Looks good, though not tested myself.

PR #274, opened 2025-02-17
Rework NvapiD3D12Device

jp7677:
Seems to work, but yet happy with this one ....

### DIFF
--- a/src/nvapi/nvapi_d3d12_device.cpp
+++ b/src/nvapi/nvapi_d3d12_device.cpp
@@ -2,49 +2,64 @@
 #include "../util/util_log.h"
 
 namespace dxvk {
+    std::unordered_map<ID3D12Device*, NvapiD3d12Device> NvapiD3d12Device::m_nvapiDeviceMap;
+    std::mutex NvapiD3d12Device::m_mutex;
 
-    std::optional<LUID> NvapiD3d12Device::GetLuid(IUnknown* unknown) {
-        Com<ID3D12Device> device;
-        if (FAILED(unknown->QueryInterface(IID_PPV_ARGS(&device))))
-            return {};
+    std::unordered_map<NVDX_ObjectHandle, NvU32> NvapiD3d12Device::m_cubinSmemMap;
+    std::mutex NvapiD3d12Device::m_cubinSmemMutex;
 
-        return device->GetAdapterLuid();
+    void NvapiD3d12Device::Reset() {
+        std::scoped_lock lock{m_mutex, m_cubinSmemMutex};
+        m_nvapiDeviceMap.clear();
+        m_cubinSmemMap.clear();
     }
 
-    bool NvapiD3d12Device::CreateGraphicsPipelineState(ID3D12Device* device, const D3D12_GRAPHICS_PIPELINE_STATE_DESC* pipelineStateDescription, NvU32 numberOfExtensions, const NVAPI_D3D12_PSO_EXTENSION_DESC** extensions, ID3D12PipelineState** pipelineState) {
-        if (numberOfExtensions != 1 || extensions[0]->psoExtension != NV_PSO_ENABLE_DEPTH_BOUND_TEST_EXTENSION)
-            return false;
+    NvapiD3d12Device* NvapiD3d12Device::GetOrCreate(ID3D12Device* device) {
+        std::scoped_lock lock{m_mutex};
 
-        return SUCCEEDED(device->CreateGraphicsPipelineState(pipelineStateDescription, __uuidof(ID3D12PipelineState), reinterpret_cast<void**>(pipelineState)));
-    }
-
-    bool NvapiD3d12Device::SetDepthBoundsTestValues(ID3D12GraphicsCommandList* commandList, const float minDepth, const float maxDepth) {
-        Com<ID3D12GraphicsCommandList1> commandList1;
-        if (FAILED(commandList->QueryInterface(IID_PPV_ARGS(&commandList1)))) // There is no VKD3D-Proton version out there that does not implement ID3D12GraphicsCommandList1, this should always succeed
-            return false;
-
-        commandList1->OMSetDepthBounds(minDepth, maxDepth);
-        return true;
-    }
-
-    bool NvapiD3d12Device::CreateCubinComputeShaderWithName(ID3D12Device* device, const void* cubinData, NvU32 cubinSize, NvU32 blockX, NvU32 blockY, NvU32 blockZ, const char* shaderName, NVDX_ObjectHandle* pShader) {
-        auto cubinDevice = GetCubinDevice(device);
-        if (cubinDevice == nullptr)
-            return false;
-
-        return CreateCubinComputeShaderEx(device, cubinData, cubinSize, blockX, blockY, blockZ, 0 /* smemSize */, shaderName, pShader);
-    }
-
-    bool NvapiD3d12Device::CreateCubinComputeShaderEx(ID3D12Device* device, const void* cubinData, NvU32 cubinSize, NvU32 blockX, NvU32 blockY, NvU32 blockZ, NvU32 smemSize, const char* shaderName, NVDX_ObjectHandle* pShader) {
-        auto cubinDevice = GetCubinDevice(device);
-        if (cubinDevice == nullptr)
-            return false;
+        auto itF = m_nvapiDeviceMap.find(device);
+        if (itF != m_nvapiDeviceMap.end())
+            return &itF->second;
 
         Com<ID3D12DeviceExt> deviceExt;
-        if (FAILED(cubinDevice->QueryInterface(IID_PPV_ARGS(&deviceExt))))
+        if (FAILED(device->QueryInterface(IID_PPV_ARGS(&deviceExt))))
+            return nullptr;
+
+        auto [itI, inserted] = m_nvapiDeviceMap.emplace(
+            std::piecewise_construct,
+            std::forward_as_tuple(device),
+            std::forward_as_tuple(deviceExt.ptr()));
+
+        if (!inserted)
+            return nullptr;
+
+        return &itI->second;
+    }
+
+    uint32_t NvapiD3d12Device::FindCubinSmem(NVDX_ObjectHandle pShader) {
+        std::scoped_lock lock(m_cubinSmemMutex);
+        auto it = m_cubinSmemMap.find(pShader);
+        if (it != m_cubinSmemMap.end())
+            return it->second;
+
+        log::info("Failed to find CuBIN in m_cubinSmemMap, defaulting to 0");
+        return 0;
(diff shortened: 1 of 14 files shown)


jp7677:
Ready for first review round. Not as elegant as the other two imho. Please let me know if you have an idea to clean up even more. Interestingly, the first commit even yielded an fps in the CP2077 benchmark for me (I haven't looked yet why it used version 0 of the Ext interface).

src/nvapi/nvapi_d3d12_device.cpp
```diff
@@ -2,6 +2,70 @@
...
+    }
+
+    NvapiD3d12Device::NvapiD3d12Device(ID3D12CommandQueueExt* vkd3dCommandQueue)
+        : m_vkd3dCommandQueue(vkd3dCommandQueue) {}
+
+    NvapiD3d12Device::NvapiD3d12Device(ID3D12GraphicsCommandListExt* vkd3dCommandList)
+        : m_vkd3dGraphicsCommandList(static_cast<ID3D12GraphicsCommandListExt1*>(vkd3dCommandList)) { // NOLINT(*-pro-type-static-cast-downcast)
+        Com<ID3D12GraphicsCommandListExt1> commandListExt1 = nullptr;
+        m_supportsExtGraphicsCommandList1 = SUCCEEDED(vkd3dCommandList->QueryInterface(IID_PPV_ARGS(&commandListExt1)));
+    }
```
Saancreed, line 68: I don't like this. This will make the `NvapiD3d12Device` into a class that's actually one of three things under the hood: a device, a command queue, or a command list. Can we please split this into separate classes? `NvapiD3d12Device`, `NvapiD3d12CommandQueue` and `NvapiD3d12CommandList`, then keep the responsibilities of each class limited to their own.

At worst, the command list one would need an additional method to retrieve smem size from the map technically owned by devices but that shouldn't be too much of a trouble.

(I'm not sure if the map with smem sizes should be a global one or per device but it probably doesn't matter. The former is likely simpler, and it's not like we will have applications running cubin stuff on multiple logical devices at the same time.)

src/nvapi/nvapi_d3d12_device.cpp
```diff
@@ -27,180 +91,86 @@ namespace dxvk {
...
         std::scoped_lock lock(m_cubinSmemMutex);
         m_cubinSmemMap.erase(shader);
         return true;
     }
 
-    bool NvapiD3d12Device::GetCudaTextureObject(ID3D12Device* device, D3D12_CPU_DESCRIPTOR_HANDLE srvHandle, D3D12_CPU_DESCRIPTOR_HANDLE samplerHandle, NvU32* cudaTextureHandle) {
-        auto cubinDevice = GetCubinDevice(device);
-        if (cubinDevice == nullptr)
+    bool NvapiD3d12Device::GetCudaTextureObject(D3D12_CPU_DESCRIPTOR_HANDLE srvHandle, D3D12_CPU_DESCRIPTOR_HANDLE samplerHandle, NvU32* cudaTextureHandle) const {
+        if (!m_vkd3dDevice || !m_supportsNvxBinaryImport)
```
Saancreed, line 123: As with D3D11, technically it's `VK_NVX_image_view_handle` that we care about here. But feel free to keep this, in practice it probably doesn't matter as long as NVK doesn't decide to implement one extension without implementing the other.

src/nvapi/nvapi_d3d12_device.cpp
```diff
@@ -27,180 +91,86 @@ namespace dxvk {
...
-            log::info("Failed to find CuBIN in m_cubinSmemMap, defaulting to 0");
-
-        if (interfaceVersion >= 1)
-            return SUCCEEDED(cmdList->LaunchCubinShaderEx(reinterpret_cast<D3D12_CUBIN_DATA_HANDLE*>(pShader), blockX, blockY, blockZ, smem, params, paramSize, nullptr, 0));
-        else {
-            if (smem != 0)
-                log::info("Non-zero SMEM value supplied for CuBIN but ID3D12GraphicsCommandListExt1 not supported! This may cause corruption");
-
-            return SUCCEEDED(cmdList->LaunchCubinShader(reinterpret_cast<D3D12_CUBIN_DATA_HANDLE*>(pShader), blockX, blockY, blockZ, params, paramSize));
+        if (m_supportsExtGraphicsCommandList1 || log::tracing()) {
```
Saancreed, line 148: Why `|| log::tracing()` here? I'd expect something like this shouldn't change the observable behavior… or cause misleading messages to be reported because the underlying implementation doesn't support GCLExt1 even if the application handles it gracefully.

src/nvapi/nvapi_d3d12_device.cpp
```diff
@@ -27,180 +91,86 @@ namespace dxvk {
...
-        Com<ID3D12DeviceExt> deviceExt;
-        if (FAILED(device->QueryInterface(IID_PPV_ARGS(&deviceExt))))
-            return nullptr;
+        if (m_supportsExtGraphicsCommandList1 >= 1)
+            return SUCCEEDED(m_vkd3dGraphicsCommandList->LaunchCubinShaderEx(reinterpret_cast<D3D12_CUBIN_DATA_HANDLE*>(pShader), blockX, blockY, blockZ, smem, params, paramSize, nullptr, 0));
 
-        if (!deviceExt->GetExtensionSupport(extension))
-            return nullptr;
+        if (log::tracing() && smem != 0)
+            log::info("Non-zero SMEM value supplied for CuBIN but ID3D12GraphicsCommandListExt1 not supported! This may cause corruption");
```
Saancreed, line 161: This is probably serious enough that I'd log it even if tracing isn't enabled.

src/nvapi/nvapi_d3d12_device.cpp
jp7677: The idea is to avoid the lock and map lookup when Ext0 is used. When not using Ext1, `smem` is only used for the log statement below. So only logging that statement on tracing, let's us skip the lock and lookup.

jp7677:
> > Interestingly, the first commit even yielded an fps in the CP2077 benchmark for me (I haven't looked yet why it used version 0 of the Ext interface).
> 
> I can see how unlocking the mutex before actually calling the command list to record cubin launch request could bring some uplifts there, but the interface version _should_ be 1 if vkd3d-proton supports it even if the game didn't use it… and Cyberpunk 2077 should be using it because it has Ray Reconstruction.

Scratch that part about CP using Ext0, that was being being set on the wrong foot because of `dynamic_cast`.

src/nvapi/nvapi_d3d12_device.cpp
jp7677: I've restored the log message.

src/nvapi/nvapi_d3d12_device.cpp
jp7677: I've restored the log message, but still kept the lock scoped.

src/nvapi/nvapi_d3d12_device.h
```diff
@@ -5,46 +5,36 @@
...
-        static std::optional<LUID> GetLuid(IUnknown* unknown);
+        static void Reset();
+        [[nodiscard]] static NvapiD3d12Device* GetOrCreate(ID3D12Device* device);
+        [[nodiscard]] static uint32_t FindCubinSmem(NVDX_ObjectHandle);
+
+        explicit NvapiD3d12Device(ID3D12DeviceExt* vkd3dDevice);
 
+        static std::optional<LUID> GetLuid(IUnknown* unknown);
         static bool CreateGraphicsPipelineState(ID3D12Device* device, const D3D12_GRAPHICS_PIPELINE_STATE_DESC* pipelineStateDescription, NvU32 numberOfExtensions, const NVAPI_D3D12_PSO_EXTENSION_DESC** extensions, ID3D12PipelineState** pipelineState);
         static bool SetDepthBoundsTestValues(ID3D12GraphicsCommandList* commandList, float minDepth, float maxDepth);
```
Saancreed, line 19: I'm not yet sure if this should stay here. For one, none of these really access anything NVAPI-specific… at least not yet. Ann top of that:

* `GetLuid` is used only in `NvAPI_D3D12_GetGraphicsCapabilities`, and only because the input is typed as `IUnknown` :weary:;
* `CreateGraphicsPipelineState` just checks for one known extension and forwards to core D3D12 function otherwise.
* `SetDepthBoundsTestValues` upgrades the interface to GCL1 then calls core D3D12 function… but even if you want to avoid inlining this, it should probably live in `NvapiD3d12GraphicsCommandList` instead.

src/nvapi/nvapi_d3d12_device.h
```diff
@@ -5,46 +5,36 @@
...
-        static bool GetCudaSurfaceObject(ID3D12Device* device, D3D12_CPU_DESCRIPTOR_HANDLE uavHandle, NvU32* cudaSurfaceHandle);
-        static bool NotifyOutOfBandCommandQueue(ID3D12CommandQueue* commandQueue, D3D12_OUT_OF_BAND_CQ_TYPE type);
-        static bool LaunchCubinShader(ID3D12GraphicsCommandList* commandList, NVDX_ObjectHandle shader, NvU32 blockX, NvU32 blockY, NvU32 blockZ, const void* params, NvU32 paramSize);
-        static bool CaptureUAVInfo(ID3D12Device* device, NVAPI_UAV_INFO* uavInfo);
-        static bool IsFatbinPTXSupported(ID3D12Device* device);
+        bool CreateCubinComputeShaderWithName(const void* cubinData, NvU32 cubinSize, NvU32 blockX, NvU32 blockY, NvU32 blockZ, const char* shaderName, NVDX_ObjectHandle* pShader);
+        bool CreateCubinComputeShaderEx(const void* cubinData, NvU32 cubinSize, NvU32 blockX, NvU32 blockY, NvU32 blockZ, NvU32 smemSize, const char* shaderName, NVDX_ObjectHandle* pShader);
+        [[nodiscard]] bool DestroyCubinComputeShader(NVDX_ObjectHandle shader);
```
Saancreed, line 23: `[[nodiscard]]` only on some selected methods? Why not all of them?

src/nvapi/nvapi_d3d12_device.h
jp7677: I don't mind them here, but also ok to move those, so moved down them.

src/nvapi/nvapi_d3d12_device.cpp
jp7677: I’ve put this on my todo list for a follow up pr, also for d3d11.

src/nvapi/nvapi_d3d12_device.h
jp7677: PS: will also inline the other two.

Saancreed (approved):

PR #275, opened 2025-02-19
nvapi-d3d11/12: Check for VK_NVX_image_view_handle

jp7677:
Follow up from earlier D3D11 and D3D12 rework.


Saancreed (approved):
Just a few unrelated questions now that you touched an unrelated thing.

PR #276, opened 2025-02-19
Use ID3D12DeviceExt2 for cubin 64-bit support 

Saancreed:


src/nvapi_d3d12.cpp
```diff
@@ -199,6 +199,153 @@ extern "C" {
...
+        params.pNext = nullptr;
+        params.texDesc = pParams->texDesc.ptr;
+        params.smpDesc = pParams->smpDesc.ptr;
+
+        switch (device->GetCudaMergedTextureSamplerObject(&params)) {
+            case S_OK:
+                pParams->textureHandle = params.textureHandle;
+                return Ok(n, alreadyLoggedOk);
+            case E_INVALIDARG:
+                return InvalidArgument(n);
```
jp7677: Slightly unfortunate that we don't know if `InvalidArgument` comes from our side or from VKD3D-Proton if we see this in the logs. I guess the alternative would be just `Error` but that's sounds somewhat dramatic. Well lets see how often we hit this, we can improve logging later.

PR #279, opened 2025-02-21
nvapi-d3d: Just error out of GetCurrentSLIState

K0bin:
Fallout New Vegas seems to think it's running on an SLI system unless we just error out here.

I also tried setting `maxNumAFRGroups` and `numAFRGroups` both to 0 but that didn't fix it either.
Additionally I tried just setting `numAFRGroups` to 0 and keeping `maxNumAFRGroups` at 1 but that didn't work either.

The Nvidia sample suggests what we have right now is a valid implementation:
`bSLIEnabled = (gMaxNumAFRGroups > 1);` but it appears that the check in Fallout isn't particularly robust.

So I suggest we just return an error when the function gets called. Alternatively we could return null pointers for the SLI functions in `QueryInterface`. I don't have a test setup for Nvapi on Windows right now, so I can't check what Nvidia does on Windows these days without a lot of effort.

Fixes: https://github.com/doitsujin/dxvk/issues/4696
https://github.com/ValveSoftware/Proton/issues/356#issuecomment-2665797622

### DIFF
--- a/src/util/util_statuscode.h
+++ b/src/util/util_statuscode.h
@@ -158,6 +158,12 @@ namespace dxvk {
         return NVAPI_INSUFFICIENT_BUFFER;
     }
 
+    inline NvAPI_Status NoActiveSliTopology(const std::string& logMessage, bool& alreadyLogged) {
+        if (log::tracing() || !std::exchange(alreadyLogged, true))
+            log::info(str::format("<-", logMessage, ": No active SLI topology"));
+        return NVAPI_NO_ACTIVE_SLI_TOPOLOGY;
+    }
+
     inline NV_OF_STATUS Success() {
         return NV_OF_SUCCESS;
     }
--- a/src/nvapi_d3d.cpp
+++ b/src/nvapi_d3d.cpp
@@ -65,7 +65,7 @@ extern "C" {
 
     NvAPI_Status __cdecl NvAPI_D3D_GetCurrentSLIState(IUnknown* pDevice, NV_GET_CURRENT_SLI_STATE* pSliState) {
         constexpr auto n = __func__;
-        thread_local bool alreadyLoggedOk = false;
+        thread_local bool alreadyLoggedNoSli = false;
 
         if (log::tracing())
             log::trace(n, log::fmt::ptr(pDevice), log::fmt::ptr(pSliState));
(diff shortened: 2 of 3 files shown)


adamdmoss:
This feels wrong, at least for the purposes of fixing one game... NV docs suggest that an error from the nvapi call should even be interpreted by a game as 'prompt the player to upgrade their driver' which is counterproductive.

https://docs.nvidia.com/gameworks/content/technologies/desktop/sli_using_nvapi.htm

The fix would be to figure out why this game is happy with the result from real nvapi and unhappy with the result from dxvk-nvapi.  dxvk-nvapi's implementation looks sensible at first glance but I wouldn't be surprised if a game expects numAFRGroups to be 0 instead of 1.

K0bin:
> but I wouldn't be surprised if a game expects numAFRGroups to be 0 instead of 1.

I tested that. It did not fix it. I guess I need to see what Windows returns.

K0bin:
Okay I did thorough testing on Windows.

The docs don't list `NVAPI_NO_ACTIVE_SLI_TOPOLOGY` as a possible return value but that's exactly what I got.
The struct fields do **not** get initialized except `numVRSLIGpus` of version 2 of the struct. That one gets set to 0, the others don't get changed. The other errors (incompatible struct, invalid argument) do take precedence, so that part stays as it was.

`NvAPI_D3D_ImplicitSLIControl` does **not** return `NVAPI_NO_ACTIVE_SLI_TOPOLOGY`, regardless of the value passed for `implicitSLIControl`.

K0bin:
> error: XDG_RUNTIME_DIR is invalid or not set in the environment.

The failed CI test isn't caused by the PR.

Saancreed:
It is, you just looked at the wrong message:

```
D3D methods succeed
  GetCurrentSLIState succeeds
  GetCurrentSLIState with unknown struct version returns incompatible-struct-
  version
-------------------------------------------------------------------------------
../../../tests/nvapi_d3d.cpp:52
...............................................................................

../../../tests/nvapi_d3d.cpp:55: FAILED:
  REQUIRE( NvAPI_D3D_GetCurrentSLIState(&unknown, &state) == NVAPI_INCOMPATIBLE_STRUCT_VERSION )
with expansion:
  -113 == -9

===============================================================================
test cases:    20 |    [19](https://github.com/jp7677/dxvk-nvapi/actions/runs/13467194283/job/37635278839?pr=279#step:8:20) passed | 1 failed
assertions: 27591 | 27590 passed | 1 failed
```

K0bin:
Should be fixed now.

Saancreed (approved):

adamdmoss:
Nice (and surprising) discoveries.

src/util/util_statuscode.h
```diff
@@ -158,6 +158,12 @@ namespace dxvk {
         return NVAPI_INSUFFICIENT_BUFFER;
     }
 
+    inline NvAPI_Status NoActiveSliTopology(const std::string& logMessage, bool& alreadyLogged) {
+        if (log::tracing() || !std::exchange(alreadyLogged, true))
+            log::info(str::format("<-", logMessage, ": Error"));
```
jp7677, line 163: This should be `"NoActiveSLITopology"` instead of `"Error"`  to match the corresponding return code in the logs.

jp7677 (changes requested):
Minor nit, otherwise LGTM! If you are at it, you could tweak the commit message slightly since it feels a bit more graceful now than just "error out". But also ok if you want to keep it like it is.

src/util/util_statuscode.h
K0bin: Oops. That's what happens when you finish the commit at 2am.

K0bin: Fixed the log line and updated the commit.

jp7677 (approved):

PR #283, opened 2025-02-22
nvapi-d3d: Make GetCurrentSLIState more accurate

K0bin:
The gift that keeps on giving.

My previous PR apparently breaks DLSS in D3D11 games.
It turns out that what my previous PR does is only correct if the device passed in is a D3D9 device.

D3D9:
image

D3D11:
image

In addition to that, Fallout doesn't even pass its D3D9 device to this function. Instead it passes `IDirect3D9`. In that case the function returns `INVALID_ARGUMENT` which means that this SLI optimization, that we were getting, doesn't even work on Windows. The game is just broken.

cc @adamdmoss

### DIFF
--- a/src/nvapi_d3d.cpp
+++ b/src/nvapi_d3d.cpp
@@ -65,6 +65,7 @@ extern "C" {
 
     NvAPI_Status __cdecl NvAPI_D3D_GetCurrentSLIState(IUnknown* pDevice, NV_GET_CURRENT_SLI_STATE* pSliState) {
         constexpr auto n = __func__;
+        thread_local bool alreadyLoggedOk = false;
         thread_local bool alreadyLoggedNoSli = false;
 
         if (log::tracing())
@@ -73,17 +74,53 @@ extern "C" {
         if (!pDevice || !pSliState)
             return InvalidArgument(n);
 
-        if (pSliState->version >= NV_GET_CURRENT_SLI_STATE_VER2) {
-            // For some reason this is the only field that gets initialized.
-            pSliState->numVRSLIGpus = 0;
-        }
-
         if (pSliState->version > NV_GET_CURRENT_SLI_STATE_VER2)
             return IncompatibleStructVersion(n, pSliState->version);
 
         // The docs don't list NVAPI_NO_ACTIVE_SLI_TOPOLOGY as a return value,
         // but testing on Windows yielded this exact status.
-        return NoActiveSliTopology(n, alreadyLoggedNoSli);
+        // It only returns that for D3D9 devices though.
+        Com<IDirect3DDevice9> device9;
+        Com<ID3D11Device> device11;
+        if (pDevice->QueryInterface(IID_PPV_ARGS(&device9)) == S_OK) {
+            if (pSliState->version == NV_GET_CURRENT_SLI_STATE_VER2) {
+                // Tests on Windows show that this is the only field that gets initialized in this case.
+                pSliState->numVRSLIGpus = 0;
+            }
+            // Report that SLI is not available through the return code.
+            // NvAPI does it this way for D3D9 devices on Windows.
+            return NoActiveSliTopology(n, alreadyLoggedNoSli);
+        } else if (pDevice->QueryInterface(IID_PPV_ARGS(&device11)) != S_OK) {
+            // Calling it with an object that is neither a D3D9 device, nor a D3D11 device returns INVALID_ARGUMENT.
+            // Hello, Fallout New Vegas.
+            return InvalidArgument(n);
+        }
+
+        switch (pSliState->version) {
+            case NV_GET_CURRENT_SLI_STATE_VER1: {
+                auto pSliStateV1 = reinterpret_cast<NV_GET_CURRENT_SLI_STATE_V1*>(pSliState);
+                // Report that SLI is not available
+                pSliStateV1->maxNumAFRGroups = 1;
+                pSliStateV1->numAFRGroups = 1;
+                pSliStateV1->currentAFRIndex = 0;
+                pSliStateV1->nextFrameAFRIndex = 0;
+                pSliStateV1->previousFrameAFRIndex = 0;
+                pSliStateV1->bIsCurAFRGroupNew = false;
+                break;
+            }
+            case NV_GET_CURRENT_SLI_STATE_VER2:
+                // Report that SLI is not available
+                pSliState->maxNumAFRGroups = 1;
+                pSliState->numAFRGroups = 1;
+                pSliState->currentAFRIndex = 0;
+                pSliState->nextFrameAFRIndex = 0;
+                pSliState->previousFrameAFRIndex = 0;
+                pSliState->bIsCurAFRGroupNew = false;
+                pSliState->numVRSLIGpus = 0;
+                break;
+        }
+
+        return Ok(n, alreadyLoggedOk);
     }
 
     NvAPI_Status __cdecl NvAPI_D3D_ImplicitSLIControl(IMPLICIT_SLI_CONTROL implicitSLIControl) {
--- a/src/nvapi_private.h
+++ b/src/nvapi_private.h
@@ -32,6 +32,7 @@
 #include <utility>
 #include <vector>
 
+#include <d3d9.h>
 #include <dxgi1_6.h>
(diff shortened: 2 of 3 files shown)


src/nvapi_d3d.cpp
```diff
@@ -73,17 +74,52 @@ extern "C" {
...
-
         if (pSliState->version > NV_GET_CURRENT_SLI_STATE_VER2)
             return IncompatibleStructVersion(n, pSliState->version);
 
         // The docs don't list NVAPI_NO_ACTIVE_SLI_TOPOLOGY as a return value,
         // but testing on Windows yielded this exact status.
-        return NoActiveSliTopology(n, alreadyLoggedNoSli);
+        // It only returns that for D3D9 devices though.
+        void* unused = nullptr;
+        if (pDevice->QueryInterface(__uuidof(IDirect3DDevice9), &unused) == S_OK) {
```
Saancreed, line 84: This will leak the refcount, won't it?

src/nvapi_d3d.cpp
```diff
@@ -73,17 +74,52 @@ extern "C" {
...
+            if (pSliState->version == NV_GET_CURRENT_SLI_STATE_VER2) {
+                // Tests on Windows show that this is the only field that gets initialized in this case.
+                pSliState->numVRSLIGpus = 0;
+            }
+            // Report that SLI is not available through the return code.
+            // NvAPI does it this way for D3D9 devices on Windows.
+            return NoActiveSliTopology(n, alreadyLoggedNoSli);
+        } else if (pDevice->QueryInterface(__uuidof(ID3D11Device), &unused) != S_OK) {
```
Saancreed, line 92: …and so will this, I suppose.

src/nvapi_d3d.cpp
```diff
@@ -65,6 +65,7 @@ extern "C" {
 
     NvAPI_Status __cdecl NvAPI_D3D_GetCurrentSLIState(IUnknown* pDevice, NV_GET_CURRENT_SLI_STATE* pSliState) {
```
jp7677, line 66: Suggestion to set this up like below. Filling the extra fields when not actually needed wont hurt I guess, but slightly improves readability.

```c++
    NvAPI_Status __cdecl NvAPI_D3D_GetCurrentSLIState(IUnknown* pDevice, NV_GET_CURRENT_SLI_STATE* pSliState) {
        constexpr auto n = __func__;
        thread_local bool alreadyLoggedOk = false;
        thread_local bool alreadyLoggedNoSli = false;

        if (log::tracing())
            log::trace(n, log::fmt::ptr(pDevice), log::fmt::ptr(pSliState));

        if (!pDevice || !pSliState)
            return InvalidArgument(n);

        switch (pSliState->version) {
            case NV_GET_CURRENT_SLI_STATE_VER1: {
                auto pSliStateV1 = reinterpret_cast<NV_GET_CURRENT_SLI_STATE_V1*>(pSliState);
                // Report that SLI is not available
                pSliStateV1->maxNumAFRGroups = 1;
                pSliStateV1->numAFRGroups = 1;
                pSliStateV1->currentAFRIndex = 0;
                pSliStateV1->nextFrameAFRIndex = 0;
                pSliStateV1->previousFrameAFRIndex = 0;
                pSliStateV1->bIsCurAFRGroupNew = false;
                break;
            }
            case NV_GET_CURRENT_SLI_STATE_VER2:
                // Report that SLI is not available
                pSliState->maxNumAFRGroups = 1;
                pSliState->numAFRGroups = 1;
                pSliState->currentAFRIndex = 0;
                pSliState->nextFrameAFRIndex = 0;
                pSliState->previousFrameAFRIndex = 0;
                pSliState->bIsCurAFRGroupNew = false;
                pSliState->numVRSLIGpus = 0;
                break;
            default:
                return IncompatibleStructVersion(n, pSliState->version);
        }

        Com<IDirect3DDevice9> d3d9Device = nullptr;
        if (pDevice->QueryInterface(IID_PPV_ARGS(&d3d9Device)) == S_OK)
            // Report that SLI is not available through the return code.
            // NvAPI does it this way for D3D9 devices on Windows.
            return NoActiveSliTopology(n, alreadyLoggedNoSli);

        Com<IDirect3DDevice9> d3d11Device = nullptr;
        if (pDevice->QueryInterface(IID_PPV_ARGS(&d3d11Device)) != S_OK)
            // Calling it with an object that is neither a D3D9 device, nor a D3D11 device returns INVALID_ARGUMENT.
            // Hello, Fallout New Vegas.
            return InvalidArgument(n);

        return Ok(n, alreadyLoggedOk);
```

Also, do we know that this is never called for D3D12?

src/nvapi_private.h
```diff
@@ -34,6 +34,7 @@
 
 #include <dxgi1_6.h>
 #include <d3d11_1.h>
+#include <d3d9.h>
```
jp7677, line 37: Lets move this one up to keep things somewhat sorted.

Saancreed (commented):
This might need some `Com<IUnknown>` to fix the ref counting if my suspicions are correct…

tests/nvapi_d3d.cpp
```diff
@@ -34,32 +38,90 @@ TEST_CASE("D3D methods succeed", "[.d3d]") {
...
+            REQUIRE(state.numAFRGroups == 1);
+            REQUIRE(state.currentAFRIndex == 0);
+            REQUIRE(state.nextFrameAFRIndex == 0);
+            REQUIRE(state.previousFrameAFRIndex == 0);
+            REQUIRE(state.bIsCurAFRGroupNew == false);
+        }
+
+        SECTION("GetCurrentSLIState (V2) returns OK") {
+            NV_GET_CURRENT_SLI_STATE_V2 state;
+
```
jp7677, line 67: Stray empty line.

src/nvapi_d3d.cpp
K0bin: The docs say pDevice can be either a D3D11 or D3D9 device.

Also, your variant fills all the struct fields. That doesn't match what I'm seeing on Windows. Do we care about that?

jp7677: > The docs say pDevice can be either a D3D11 or D3D9 device.

Thanks, sounds good.

> Also, your variant fills all the struct fields. That doesn't match what I'm seeing on Windows. Do we care about that?

I'm good with that. Having a few more initialized fields instead of keeping them uninitialized shouldn't make a difference for a caller.

src/nvapi_d3d.cpp
K0bin: derp

src/nvapi_d3d.cpp
jp7677: Actually, considering that FONV really does this wrong, I would be ok to do the second check with `… IDirect3D9 … == S_OK` and place it below the other argument checks. Imho no need to make this more safe than needed for this one title.

K0bin: > I'm good with that. Having a few more initialized fields instead of keeping them uninitialized shouldn't make a difference for a caller.

Games often rely on the weirdest little details. I'm fine with changing it though, DXVK-NvAPI has been used with tons of games now and it took this long to find one that's problematic here.

> Actually, considering that FONV really does this wrong, I would be ok to do the second check with … IDirect3D9 … == S_OK and place it below the other argument checks. Imho no need to make this more safe than needed for this one title.

I don't agree with that one though. We have the more correct version ready to go and would essentially just replace it with a near-identical less correct check.

Saancreed (approved):

src/nvapi_d3d.cpp
jp7677: Titles are actually quite robust with regard to NVAPI usage. This is the first case in years where a title expects a non-OK response code for this function. Anyway, the important thing was to fix DLSS for D3D11. I've merged your version manually with some minor formatting changes (mostly no `else` due to early return, using `SUCCEEDED`/`FAILED` and testing for the exact struct versions since those are not just 1 and 2).

Thanks for your work and sry that this turned out to be such a frustrating case.

jp7677:
Merged manually in https://github.com/jp7677/dxvk-nvapi/commit/1e39a9d742418213b1a994e2ac0f2b05e9e9627f

PR #289, opened 2025-04-03
util: Fake GPU query success for The Last of Us Part II

Saancreed:


src/util/util_env.cpp
```diff
@@ -157,6 +157,10 @@ namespace dxvk::env {
             || getExecutableName() == std::string("tlou-i-l.exe");
     }
 
+    bool isTheLastOfUsPartTwo() {
+        return getExecutableName() == std::string("tlou-ii.exe");
```
jp7677, line 161: Any idea if there is also an alternative name, like part one had?

Saancreed: Right, found another one on SteamDB, should be more foolproof now.

jp7677 (approved):

PR #32, opened 2021-08-24
Implement the required NvAPI entrypoints for D3D12 DLSS support

rochaudhari:
Add implementation for new apis D3D12.

Signed-off-by: Roshan Chaudhari <rochaudhari@nvidia.com>

PR #290, opened 2025-04-05
nvapi-d3d12: Add SM version fallback

jp7677:
Lets be prepared for wine versions that don't support VK_NV_cuda_kernel_launch, see https://gitlab.winehq.org/wine/wine/-/merge_requests/7743

BW version taken from https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/34331

### DIFF
--- a/src/nvapi_d3d12.cpp
+++ b/src/nvapi_d3d12.cpp
@@ -285,8 +285,39 @@ extern "C" {
         // Note: One might think that SM here is D3D12 Shader Model, in fact it is the "Streaming Multiprocessor" architecture version
         // Values are valid for Turing and newer only, due to VK_NV_cuda_kernel_launch not being supported by earlier generations
         auto computeCapability = adapter->GetComputeCapability();
-        pGraphicsCaps->majorSMVersion = computeCapability.first;
-        pGraphicsCaps->minorSMVersion = computeCapability.second;
+        if (computeCapability.has_value()) {
+            pGraphicsCaps->majorSMVersion = computeCapability.value().first;
+            pGraphicsCaps->minorSMVersion = computeCapability.value().second;
+        } else if (adapter->HasNvProprietaryDriver()) {
+            // Fallback because older than Turing or no support for VK_NV_cuda_kernel_launch
+            // Based on https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/nouveau/winsys/nouveau_device.c
+            auto architectureId = adapter->GetArchitectureId();
+            if (architectureId >= NV_GPU_ARCHITECTURE_GB200) {
+                pGraphicsCaps->majorSMVersion = 12; // Report highest known version also for future generations
+                pGraphicsCaps->minorSMVersion = 0;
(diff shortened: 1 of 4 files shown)


src/nvapi_d3d12.cpp
```diff
@@ -285,8 +285,50 @@ extern "C" {
...
-        pGraphicsCaps->majorSMVersion = computeCapability.first;
-        pGraphicsCaps->minorSMVersion = computeCapability.second;
+        if (computeCapability.has_value()) {
+            pGraphicsCaps->majorSMVersion = computeCapability.value().first;
+            pGraphicsCaps->minorSMVersion = computeCapability.value().second;
+        } else if (adapter->HasNvProprietaryDriver()) {
+            // Fallback because older than Turing or no support for VK_NV_cuda_kernel_launch
+            // Based on https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/nouveau/winsys/nouveau_device.c
+            switch (adapter->GetArchitectureId()) {
+                case NV_GPU_ARCHITECTURE_GB200:
```
Saancreed, line 295: What about future architectures that we don't know about yet? Maybe it would make sense to do `arch >= NV_GPU_ARCHITECTURE_*` comparisons then, so e.g. Rubin and Feynman GPUs won't get 0.0 until we handle them explicitly.

src/nvapi_d3d12.cpp
```diff
@@ -285,8 +285,50 @@ extern "C" {
         // Note: One might think that SM here is D3D12 Shader Model, in fact it is the "Streaming Multiprocessor" architecture version
         // Values are valid for Turing and newer only, due to VK_NV_cuda_kernel_launch not being supported by earlier generations
         auto computeCapability = adapter->GetComputeCapability();
-        pGraphicsCaps->majorSMVersion = computeCapability.first;
-        pGraphicsCaps->minorSMVersion = computeCapability.second;
+        if (computeCapability.has_value()) {
+            pGraphicsCaps->majorSMVersion = computeCapability.value().first;
+            pGraphicsCaps->minorSMVersion = computeCapability.value().second;
+        } else if (adapter->HasNvProprietaryDriver()) {
```
Saancreed, line 291: `|| HasNvk()` perhaps? IIRC we didn't have any driver checks previously here but I'm also not sure what behavior would be better for NVK.

src/nvapi_d3d12.cpp
jp7677: Yeah, good idea.

src/nvapi_d3d12.cpp
jp7677: Good question. I tend to report `0` for NVK like we currently do (because no `VK_NV_cuda_kernel_launch` for NVK). I guess it also doesn't matter that much, at least the original PR for DLLS/D3D12 didn't set those values at all (https://github.com/jp7677/dxvk-nvapi/pull/32) when introducing this entry point.

Saancreed (approved):

PR #307, opened 2025-07-16
Allow multiple of the same latency marker for a given frameID

wcampbell-nv:
With frame generation apps will send multiple OUT_OF_BAND_PRESENT markers for a given frameID. This lets the Nvidia driver know that frame gen is happening.

dxvk-nvapi currently discards all but the first OUT_OF_BAND_PRESENT marker. This change removes that logic and passes through all of the app's markers.

Tested by running Cyberpunk 2077 with 4x frame generation and logging the markers received by the Nvidia driver. Originally we only saw one VK_LATENCY_MARKER_OUT_OF_BAND_PRESENT_START_NV marker per frameID. With this change we see four.

### DIFF
--- a/src/nvapi/low_latency_frame_id_generator.cpp
+++ b/src/nvapi/low_latency_frame_id_generator.cpp
@@ -4,7 +4,6 @@ namespace dxvk {
     LowLatencyFrameIdGenerator::LowLatencyFrameIdGenerator()
         : m_nextLowLatencyDeviceFrameId(1),
           m_applicationIdList({0}) {
-        m_lastFrameId.fill(std::numeric_limits<uint64_t>::max());
     }
 
     LowLatencyFrameIdGenerator::~LowLatencyFrameIdGenerator() = default;
@@ -40,10 +39,5 @@ namespace dxvk {
             return 0;
 
         return m_applicationIdList[((lowLatencyDeviceFrameId - 1) % applicationIdListSize)];
-    }
-
-    bool LowLatencyFrameIdGenerator::IsRepeatedFrame(uint64_t frameID, uint32_t markerType) {
-        // Should always be within bounds since we drop unsupported marker types in the entrypoints
-        return std::exchange(m_lastFrameId[markerType], frameID) == frameID;
     }
 }
--- a/src/nvapi/low_latency_frame_id_generator.h
+++ b/src/nvapi/low_latency_frame_id_generator.h
@@ -11,7 +11,6 @@ namespace dxvk {
         [[nodiscard]] bool LowLatencyDeviceFrameIdInWindow(uint64_t lowLatencyDeviceFrameId) const;
         uint64_t GetLowLatencyDeviceFrameId(uint64_t applicationFrameId);
         uint64_t GetApplicationFrameId(uint64_t lowLatencyDeviceFrameId);
-        [[nodiscard]] bool IsRepeatedFrame(uint64_t frameID, uint32_t markerType);
 
       private:
         std::mutex m_frameIdGeneratorMutex;
@@ -21,7 +20,5 @@ namespace dxvk {
 
         static constexpr uint32_t applicationIdListSize = 1000;
         std::array<uint64_t, applicationIdListSize> m_applicationIdList;
-
-        std::array<uint64_t, 13> m_lastFrameId;
     };
 }
--- a/src/nvapi/nvapi_d3d_low_latency_device.cpp
+++ b/src/nvapi/nvapi_d3d_low_latency_device.cpp
@@ -137,9 +137,6 @@ namespace dxvk {
     }
 
     HRESULT NvapiD3dLowLatencyDevice::SetLatencyMarker(uint64_t frameID, uint32_t markerType) {
-        if (m_frameIdGenerator.IsRepeatedFrame(frameID, markerType))
-            return S_OK; // Silently drop repeated frame IDs
-
         return m_d3dLowLatencyDevice->SetLatencyMarker(
             m_frameIdGenerator.GetLowLatencyDeviceFrameId(frameID), markerType);
     }
--- a/tests/nvapi_d3d.cpp
+++ b/tests/nvapi_d3d.cpp
@@ -413,20 +413,20 @@ TEST_CASE("D3D Reflex depending methods succeed", "[.d3d]") {
                 REQUIRE(NvAPI_D3D_SetLatencyMarker(reinterpret_cast<IUnknown*>(&d3d11Device), &latencyMarkerParams) == NVAPI_OK);
             }
 
-            SECTION("SetLatencyMarker drops repeated frame IDs and returns OK") {
+            SECTION("SetLatencyMarker passes through repeated frame IDs and returns OK") {
                 sequence seq1, seq2, seq3;
 
                 REQUIRE_CALL(lowLatencyDevice, SetLatencyMarker(1ULL, VK_LATENCY_MARKER_PRESENT_START_NV))
                     .IN_SEQUENCE(seq1)
-                    .TIMES(1)
+                    .TIMES(2)
                     .RETURN(S_OK);
                 REQUIRE_CALL(lowLatencyDevice, SetLatencyMarker(2ULL, VK_LATENCY_MARKER_PRESENT_START_NV))
                     .IN_SEQUENCE(seq2)
-                    .TIMES(1)
+                    .TIMES(2)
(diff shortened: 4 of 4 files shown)


wcampbell-nv:
Updated to get rid of duplicate PRESENT/RENDERSUBMIT/SIMULATION markers while allowing duplicate OUT_OF_BAND_PRESENT and other markers.

I've tested to confirm that The Witcher 3's duplicate PRESENT markers are removed and that we now receive multiple OUT_OF_BAND_PRESENT markers from the Streamline Sample app.

Saancreed (commented):
Thanks, I have just one small request, looks good otherwise.

jp7677 (commented):
Small nit, otherwise LGTM, thanks!
Would also be nice if you could add a second version of this test https://github.com/jp7677/dxvk-nvapi/blob/master/tests/nvapi_d3d.cpp#L416 but with markers that allow repeated frame IDs. As stated in the other PR, we can also take care of tests later if you are short on time.

adamdmoss:
Thinking aloud, is dxvk-nvapi really the right place to remove any markers at all?

Only one of these can be true (I think):
a. duplicate markers are benign but will never serve a purpose
b. duplicate markers are actually part of the app<->nvapi<->driver contract, i.e. OUT_OF_BAND_PRESENT
c. duplicate markers are an app bug which can cause real-world problems but nvidia's (windows) nvapi implementation filters them away
d. duplicate markers are an app bug which can cause real-world problems but nvidia's (windows) driver works around them
e. duplicate markers are an app bug but don't cause real-world problems

(I'm not talking about filtering non-monotonic frame ID markers, which neither #244 or this PR touch AFAICT, though I think the above classifications apply there also.)

The original PR #244 to remove duplicate markers strongly assumed _case a._ and was (IIRC) only designed to address the log spam of marker-spammy apps; within that assumption, it wasn't the wrong thing to do.  The initial version of *this* PR asserts _case b_ in basically being a revert of #244 and explaining why.

IMHO unless _case c._ is actually true (duplicate markers are an app bug which can cause real-world problems but nvidia's [windows] nvapi implementation filters them away) then the filtering belongs outside of dxvk-nvapi or shouldn't exist at all.

Saancreed:
@adamdmoss You raise a valid point, but we also need to take into consideration the following as well:

* We don't have the exact knowledge of what applications are in theory allowed or forbidden to do within the scope of D3D Reflex API, and what the Windows driver / nvapi is robust against. The Reflex SDK Integration Guide I have locally strongly suggests markers for given pair of frameID+type (`SIMULATION`, `RENDERSUBMIT` and `PRESENT`) should be sent only once, but it's also old enough to not be aware of async frame markers or `OUT_OF_BAND` markers existing, not to mention having documented how they are supposed to be used and whether duplicate `OUT_OF_BAND` markers are allowed.
* We also don't have the exact knowledge of what is allowed or forbidden within the scope of `VK_NV_ll2` (only what is written in VK spec) or whether this matches D3D Reflex. Notably, (while not exactly related to this PR) the Vulkan extension lacks a way to set async frame markers on queues at all… and over time the D3D Reflex had its API extended quite a few times (`OUT_OF_BAND_IGNORE` queue type, `NvAPI_D3D_SetReflexSync`, etc… not to mention [Streamline having PCL marker types that don't exist even in D3D Reflex](https://github.com/NVIDIA-RTX/Streamline/blob/3e78b8f4cecf4b7d83a58e9398f1edb192c74aae/include/sl_pcl.h#L73-L79)) while `VK_NV_ll2` is still at revision 2, so it's not unlikely that current D3D Reflex allows certain behaviors that were previously forbidden. The ties `VK_NV_ll2` has to swapchains and presentIDs already make the API different enough.
* We aren't programming against `VK_NV_ll2`, only DXVK's and vkd3d-proton's understanding and usage of it, and the contract of `ID3DLowLatencyDevice`, while mostly well defined, is _also_ not guaranteed to exactly match any of APIs mentioned above. I'm not sure if it's too late to change it to match what Reflex expects, but for that we'd probably have to know what _is_ expected of us.

So, going over possible cases:

> a. duplicate markers are benign but will never serve a purpose

Then we can safely filter them out if that's what vkd3d-proton wants us to do.

> b. duplicate markers are actually part of the app<->nvapi<->driver contract, i.e. OUT_OF_BAND_PRESENT

Looks like at least `OUT_OF_BAND` markers are designed to be harmless if repeated, if not actually expected to be repeatable; we didn't know that before but we know it now. The case of original `SIMULATION`, `RENDERSUBMIT` and `PRESENT` markers is trickier because at least vkd3d-proton considers repeating them to be a contract violation. But if this actually isn't misuse of the API, then vkd3d-proton (and possibly DXVK) would have to be altered to account for this.

> c. duplicate markers are an app bug which can cause real-world problems but nvidia's (windows) nvapi implementation filters them away

Then, as you observed, dxvk-nvapi is the correct place to workaround this.

> d. duplicate markers are an app bug which can cause real-world problems but nvidia's (windows) driver works around them

Isn't this case the same as C from external point of view? As far as (Windows) applications are concerned, nvapi _is_ part of the driver.

And even if we considered nvapi to be a separate component from the driver, I think it's still preferable to have the workaround here rather than in the next "driver" layer we're talking to, otherwise we'd have to replicate it in both libraries that implement `ID3DLowLatencyDevice` interface: dxvk and vkd3d-proton.

> e. duplicate markers are an app bug but don't cause real-world problems

In which case it shouldn't matter whether we filter them out or not.

I'm not saying it's objectively correct to do this filtering here, but it does seem to be a pragmatic one. Which is why I don't mind doing it here, or changing the behavior when requested by someone who knows something we don't (like NV engineers); it's probably the best we can do, given the limited information we have.

wcampbell-nv:
We've discussed this within Nvidia and have a few interesting points to add:

We're seeing the duplicate present markers because of a difference between the DX NvAPIs and VK_NV_low_latency2. In DX we expect the app to add PRESENT markers using NvAPI_D3D_SetLatencyMarker. The app then presents through Streamline which submits its own PRESENT markers through NvAPI_D3D12_SetAsyncFrameMarker. The DX driver handles these markers differently whereas Vulkan only has one vkSetLatencyMarkerNV method. dxvk-nvapi translates these different NvAPI calls into the same doubled PRESENT marker.

Vulkan Streamline is implemented the same way as DX, so it ends up submitting doubled PRESENT markers as well. I see doubled PRESENT markers in Doom The Dark Ages. The Nvidia Vulkan driver handles this and Reflex works fine. dxvk-nvapi does not strip these out.

Based on this I now agree with Adam, it would be better to pass these doubled PRESENT markers through and let the driver handle them like it does for native Vulkan apps. I'll take a look at vkd3d-proton to make sure it does not somehow rely on non-doubled markers, and revert back to the original change if @jp7677 and @Saancreed agree.

jp7677:
> We've discussed this within Nvidia and have a few interesting points to add:
> …

Thanks a lot for sharing those insights! Also thanks @adamdmoss for thinking aloud!

> Based on this I now agree with Adam, it would be better to pass these doubled PRESENT markers through and let the driver handle them like it does for native Vulkan apps. I'll take a look at vkd3d-proton to make sure it does not somehow rely on non-doubled markers, and revert back to the original change if @jp7677 and @Saancreed agree.

I’m good with removing the filtering here (so reverting above mentioned PR) with the assumption that vkd3d-proton can indeed be tweaked to handle this gracefully, thus also without log spam since as Adam pointed out, this was what got the filtering started. Very nice!

May be it’s also wise to tweak the VK_NV_ll2 documentation in the long term with those behavior expectations.

Saancreed:
On top of what Jens already said, I can also suggest merging this PR as a partial solution and fully revert the filtering later, when vkd3d-proton is prepared for such a change.

wcampbell-nv:
I sent out a PR to remove the warning messages on the vkd3d side. If that is approved I'll update this PR to remove the filtering entirely, it'll be less churn than submitting the partial solution now. Thanks for the reviews and test cases!

wcampbell-nv:
Updated to allow all types of markers to repeat, as my change removing the warnings in vkd3d was merged (https://github.com/HansKristian-Work/vkd3d-proton/pull/2555).

I also slightly updated the test to reflect the new behavior. I'm not sure how to run the test locally, let's hope it passes in CI.

jp7677 (approved):
LGTM. Just deleting the test case would also be fine.

Saancreed (approved):

jp7677:
Thanks @wcampbell-nv !

PR #318, opened 2025-08-29
Fake Reflex not supported for Stormgate

jp7677:
Not super cool, but better than crashing.

### DIFF
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -163,6 +163,20 @@ namespace dxvk::env {
         return false;
     }
 
+    bool needsUnsupportedLowLatencyDevice() {
+        static bool alreadyLogged = false;
+
+        static bool isStormgate = getExecutableName() == std::string("Stormgate-Win64-Shipping.exe");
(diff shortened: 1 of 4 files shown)


src/util/util_env.cpp
```diff
@@ -130,62 +130,52 @@ namespace dxvk::env {
...
             return true;
         }
 
         return false;
     }
 
+    bool needsUnsupportedLowLatencyDevice() {
+        static bool alreadyLogged = false;
+
+        if (getExecutableName() == std::string("Stormgate-Win64-Shipping.exe")) {
```
Saancreed, line 169: Since this is on a hot path, I'd prefer we evaluated it only once and cached the result. Maybe something like

```suggestion
        static bool isStormgate = getExecutableName() == std::string("Stormgate-Win64-Shipping.exe");

        if (isStormgate) {
```

would work?

jp7677: I went for the `static bool`, looks fine on paper.

Saancreed (approved):

PR #344, opened 2026-02-18
Add tests for Cubin 64-bit functions

jp7677:
and, well, some stuff surrounding that ...


src/nvapi/nvapi_d3d12_device.cpp
```diff
@@ -18,19 +18,23 @@ namespace dxvk {
...
 
-        if (!registry)
+        if (!factory || !registry)
             return false;
 
         uint32_t adapterCount = registry->GetAdapterCount();
         for (uint32_t i = 0; i < adapterCount; ++i) {
             auto dxgiAdapter = registry->GetAdapter(i)->GetDxgiAdapter();
 
+            Com<ID3D12Device> d3d12Device = factory->CreateD3D12Device(dxgiAdapter, D3D_FEATURE_LEVEL_12_0);
```
Saancreed: `auto d3d12Device`?

That said, it would be nice if we could ask for `ID3D12DeviceExt2` directly and skip the `QueryInterface` step below. Not too important though, it's pretty much a µ-opt.

PR #352, opened 2026-04-04
nvapi-disp: Sanitize hdr metadata input

jp7677:
~Fallback to the monitor's actual EDID capabilities if the application fails to provide valid luminance boundaries. This mimics Windows NVIDIA driver behavior.~ Amended to just filter out invalid values.

This is not on a hot path and seems to fix a real word issue, so I guess being pragmatic here is fine. Based on https://forums.developer.nvidia.com/t/595-release-feedback-discussion/362561/296 and thus https://github.com/zaxen/dxvk-nvapi/commit/79c4bea12df7f499f8844a38cf48907a07bc9e54 

cc: @zaxen

### DIFF
--- a/src/nvapi_disp.cpp
+++ b/src/nvapi_disp.cpp
@@ -187,6 +187,13 @@ extern "C" {
                     .MaxContentLightLevel = pHDRColorDataV1->mastering_display_data.max_content_light_level,
                     .MaxFrameAverageLightLevel = pHDRColorDataV1->mastering_display_data.max_frame_average_light_level,
                 };
+
+                // Filter invalid input data
+                if (metadata.MinMasteringLuminance != 0 && metadata.MinMasteringLuminance >= metadata.MaxMasteringLuminance * 10000) {
+                    metadata.MaxMasteringLuminance = 0;
+                    metadata.MinMasteringLuminance = 0;
+                }
+
                 if (FAILED(interop->SetGlobalHDRState(colorspace, &metadata)))
                     return InvalidArgument(n);
             }
--- a/tests/nvapi_sysinfo_hdr.cpp
+++ b/tests/nvapi_sysinfo_hdr.cpp
@@ -261,6 +261,19 @@ TEST_CASE("HDR related sysinfo methods succeed", "[.sysinfo-hdr]") {
             REQUIRE(inMetadata.MaxFrameAverageLightLevel == 799);
         }
 
+        SECTION("HdrColorControl (V2) with set command filters invalid input and returns OK") {
+            NV_HDR_COLOR_DATA colorData{};
+            colorData.version = NV_HDR_COLOR_DATA_VER2;
+            colorData.cmd = NV_HDR_CMD_SET;
+            colorData.hdrMode = NV_HDR_MODE_UHDA_PASSTHROUGH;
+            colorData.mastering_display_data.min_display_mastering_luminance = 100;
+            colorData.mastering_display_data.max_display_mastering_luminance = 0;
+
+            REQUIRE(NvAPI_Disp_HdrColorControl(primaryDisplayId, &colorData) == NVAPI_OK);
+            REQUIRE(inMetadata.MinMasteringLuminance == 0);
+            REQUIRE(inMetadata.MaxMasteringLuminance == 0);
+        }
+
         SECTION("HdrColorControl with unknown struct version returns incompatible-struct-version") {
             NV_HDR_COLOR_DATA colorData;
             colorData.version = NV_HDR_COLOR_DATA_VER2 + 1;


jp7677:
Alternative with the spec in mind, but not sure if that is sufficient wrt https://forums.developer.nvidia.com/t/595-release-feedback-discussion/362561/296

```c++
if (metadata.MinMasteringLuminance != 0 && metadata.MinMasteringLuminance >= metadata.MaxMasteringLuminance * 10000) {
    metadata.MaxMasteringLuminance = data.MaxLuminance;
    metadata.MinMasteringLuminance = data.MinLuminance;
}
```

Reset to zero instead of ` data.MaxLuminance`/` data.MinLuminance` is may be even better.

Edit: Did that with the last force-push.

zaxen:
Hey @jp7677, thanks for picking this up!

First of all, unfortunately the change you suggest (`metadata.MinMasteringLuminance != 0 &&...`) doesn't fix the crash in Elden Ring because...

I was doing some [extra debugging](https://github.com/zaxen/dxvk-nvapi/tree/fix/sanitize-hdr-metadata) to see whether I could figure out why enabling HDR in game made everything look washed out, so I added some extra trace logging which spits out the following when enabling HDR:

```
712.270:01a8:01ac:trace:nvapi64:NvAPI_Disp_HdrColorControl (65537, ptr=0x11f838)
712.270:01a8:01ac:trace:nvapi64:NVAPI: NvAPI_Disp_HdrColorControl SET cmd:
712.270:01a8:01ac:trace:nvapi64:  hdrMode: UHDA_PASSTHROUGH (5)
712.270:01a8:01ac:trace:nvapi64:  static_metadata_descriptor_id: 0
712.270:01a8:01ac:trace:nvapi64:  mastering_display_data:
712.270:01a8:01ac:trace:nvapi64:    displayPrimary_x0: 0
712.270:01a8:01ac:trace:nvapi64:    displayPrimary_y0: 0
712.270:01a8:01ac:trace:nvapi64:    displayPrimary_x1: 0
712.270:01a8:01ac:trace:nvapi64:    displayPrimary_y1: 0
712.270:01a8:01ac:trace:nvapi64:    displayPrimary_x2: 0
712.270:01a8:01ac:trace:nvapi64:    displayPrimary_y2: 0
712.270:01a8:01ac:trace:nvapi64:    displayWhitePoint_x: 0
712.270:01a8:01ac:trace:nvapi64:    displayWhitePoint_y: 0
712.270:01a8:01ac:trace:nvapi64:    max_display_mastering_luminance: 0
712.270:01a8:01ac:trace:nvapi64:    min_display_mastering_luminance: 0
712.270:01a8:01ac:trace:nvapi64:    max_content_light_level: 0
712.270:01a8:01ac:trace:nvapi64:    max_frame_average_light_level: 0
```

and, obviously, the min luminance of 0 will shortcircuit your new check.

(Apologies if you already know all this)

Which led me down a rabbit hole, till I found [this document](https://developer.nvidia.com/sites/default/files/akamai/gameworks/hdr/UHDColorForGames.pdf) which states:

> Setting the mastering data (primaries, white point, max luminance, etc) to 0 has the
special meaning that the content is mastered for the default with 1000 nit max
luminance and DCI primaries. For all gaming content today, DCI primaries or sRGB
primaries and 1000 nits maximum specified should produce good results.

So I then added extra sanitisation checks to [my branch](https://github.com/zaxen/dxvk-nvapi/tree/fix/sanitize-hdr-metadata) for the primaries and white point, etc.  I tried max luminance of 1000 and some values for DCI primaries that I got off wikipedia, then used the data coming from `wsi_edid.h` from DXVK and even changed the wsi_edid.h values to the values the EDID of my monitor gives. But all of this was to no avail, none of this fixed the washed out look in Elden Ring and it's where I've given up due to lack of knowledge.

Anyway, if you want me to submit that as a PR (with all the extra logging removed), please let me know or feel free to just cherry pick the bits you want!

jp7677:
Thanks for stepping by @zaxen and for providing more data and context. I didn't know that the input data where just zeros, that indeed doesn't work with what I had in mind. 

Based on your input, I get that there are two issues? Elden Ring crashing and once it no longer crashes, the output is washed out? Do you know which of your additions fixes the crash? Am I correct that even with all your additions, the colors are still washed out?

zaxen:
> there are two issues? Elden Ring crashing and once it no longer crashes, the output is washed out?

Exactly correct.  Based on the nvidia doc, probably something like this is sufficient to guard against a crash:

```
if (metadata.MinMasteringLuminance == 0 && metadata.MaxMasteringLuminance  == 0) {
    metadata.MaxMasteringLuminance = data.MaxLuminance; // or 1000 to match the doc
    metadata.MinMasteringLuminance = data.MinLuminance; // or 0
}
```

Adding guards to set the primaries if they're all 0 may also be useful for interop but I didn't see any change in Elden Ring by adding it.

For the washed out colours, I've just discovered the running through gamescope solves the washed out colours, so presumably this issue is out of scope for dxvk-nvapi and related dvxk stack parts and exists further down in kwin or something.

I was going to try short circuiting the kwin guard that causes the crash if both min lum and max lum = 0 and see if the nvidia driver or something else handles it like Windows does but I haven't got around to that yet and probably is overkill.

jp7677:
> > there are two issues? Elden Ring crashing and once it no longer crashes, the output is washed out?
> 
> Exactly correct.

Thanks for the clarification. Initially I thought that the sanitation would completely fix HDR output, but since it doesn't, I'm hesitant to do anything on our our side. Reading that it fully works with gamescope and hearing from others that passing zero should actually be fine (https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/VkHdrMetadataEXT.html#_description) further empathizes this. 

I guess `DXVK_NVAPI_DISABLE_ENTRYPOINTS=NvAPI_Disp_HdrColorControl` hides HDR options for that title?

> I was going to try short circuiting the kwin guard that causes the crash if both min lum and max lum = 0 and see if the nvidia driver or something else handles it like Windows does but I haven't got around to that yet and probably is overkill.

It would be interesting to know what data is passed to the gamescope WSI layer and where it comes from. Not really familiar with it, but by briefly looking into it, it discards when e.g. `max_content_light_level` or `max_frame_average_light_level` are zero (like in your log statements) at https://github.com/ValveSoftware/gamescope/blob/master/src/wlserver.cpp#L933

sbearcsiro:
Yeah, that Vulkan spec is pretty definitive, it's a mismatch between the vulkan spec and wayland protocol basically?  So I guess it's on nvidia to fix their vulkan WSI to not invoke the wayland protocol when maxlum <= minlum (like the [VK_hdr_layer we had to use before 595](https://github.com/Zamundaaa/VK_hdr_layer/blob/main/src/VkLayer_hdr_wsi.cpp#L744) did).

PR #353, opened 2026-04-04
nvapi-vulkan: Prevent log spam without layer

jp7677:
Since we don't fake vk reflex by default anymore, log file fills up when logging is enabled.


Saancreed (approved):
And here I was hoping that no such calls would be made if `InitLowLatencyDevice` failed…

jp7677:
> And here I was hoping that no such calls would be made if `InitLowLatencyDevice` failed…

Yeah. Seems that Indy is somewhat stubborn wrt this..

PR #355, opened 2026-04-11
README: Add new R595 Overrides

### LINKED ISSUE #354: Plans on implementing preset selection for DLSS FG
Hi! First off I would like to thank you for your awesome work!

Along with DLSS DMFG, Nvidia released a new preset for frame generation called preset "B" so frame generation can now be either preset "A" or "B". The upsides of preset B is that it has the potential to handle UI elements much better.

I read the wiki and could not find any information regarding forcing a different FG preset, which is expected since up until now there has been no preset selection. Is preset selection for FG already supported under the hood or is this something that you plan on implementing?

Thanks!

ptr1337:
Fixes: https://github.com/jp7677/dxvk-nvapi/issues/354


Saancreed (approved):
Right, thanks.

We should probably update the wiki too…

PR #360, opened 2026-05-03
Update catch2 and some minor things

jp7677:
Just some trivial things.

### DIFF
--- a/src/nvapi_d3d12.cpp
+++ b/src/nvapi_d3d12.cpp
@@ -12,34 +12,6 @@
 #include "util/util_string.h"
 #include "util/util_env.h"
 
-static inline std::optional<NVAPI_D3D12_RAYTRACING_THREAD_REORDERING_CAPS> GetThreadReorderingCaps(ID3D12Device* pDevice, dxvk::NvapiD3d12Device* device) {
-    if (!dxvk::env::isD3d12NvShaderExtnEnabled())
-        return {};
-
-    if (!device || !device->IsNvShaderExtnOpCodeSupported(NV_EXTN_OP_HIT_OBJECT_REORDER_THREAD))
-        return {};
-
-#if defined(_MSC_VER)
-    LUID luid = pDevice->GetAdapterLuid();
-#else
-    LUID luid;
-    pDevice->GetAdapterLuid(&luid);
-#endif
-
-    auto adapter = nvapiAdapterRegistry->FindAdapter(luid);
-    if (!adapter)
-        return {};
-
-    switch (adapter->GetReorderingHint()) {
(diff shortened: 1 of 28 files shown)


src/nvapi_d3d12.cpp
```diff
@@ -735,7 +735,7 @@ extern "C" {
         return Ok(str::format(n, " (", type, "/", fromRaytracingCaps(type), ")"));
     }
 
-    static bool ConvertBuildRaytracingAccelerationStructureInputs(const NVAPI_D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS_EX* nvDesc, std::vector<D3D12_RAYTRACING_GEOMETRY_DESC>& geometryDescs, D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS* d3dDesc) {
+    extern "C++" inline bool ConvertBuildRaytracingAccelerationStructureInputs(const NVAPI_D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS_EX* nvDesc, std::vector<D3D12_RAYTRACING_GEOMETRY_DESC>& geometryDescs, D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS* d3dDesc) {
```
Saancreed: I believe this somewhat changes the meaning, as without `static` the symbol will now be visible to other TUs for little reason (and `inline` is just a hint iiuc).

That said, I don't see any other way of "fixing" this other than removing the `extern "C" {` block from the file. Ideally we'd use explicit `extern "C"` on every function that can be returned via `nvapi_QueryInterface`, which would let us have `static` functions without C linkage, but that would be a huge diff.

Saancreed (approved):

src/nvapi_d3d12.cpp
jp7677: Well, I did just that in the last three commits (including restoring `static`). The diff is actually doable when hiding white spaces, but `git blame` gets kind-of unusable.

Saancreed: :sweat_smile:

Yeah, but overall I like this a bit more.

(Maybe we should consider a macro for `extern "C" NvAPI_Status __cdecl` in the future?)

Saancreed (approved):

src/nvapi_d3d12.cpp
jp7677: did that as well while at it ;), ok with the naming?

Saancreed: Yup, looks good.

PR #361, opened 2026-05-04
tests: Add tests for D3D12 NV shader support

jp7677:


tests/nvapi_d3d12.cpp
```diff
@@ -181,9 +184,34 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
...
+            ALLOW_CALL(device, IsNvShaderExtnOpCodeSupported(_))
+                .RETURN(false);
+
+            auto supported = true;
+            REQUIRE(NvAPI_D3D12_IsNvShaderExtnOpCodeSupported(&device, NV_EXTN_OP_HIT_OBJECT_REORDER_THREAD, &supported) == NVAPI_OK);
+            REQUIRE(supported == false);
+        }
+
+        SECTION("IsNvShaderExtnOpCodeSupported with VKD3D-Proton shader support") {
+            ALLOW_CALL(device, IsNvShaderExtnOpCodeSupported(static_cast<UINT32>(NV_EXTN_OP_HIT_OBJECT_REORDER_THREAD)))
```
Saancreed, line 208: Is the `static_cast` needed? Does the macro'd literal not match the call otherwise?

tests/nvapi_d3d12.cpp
```diff
@@ -651,42 +703,108 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
...
+                                vkProps.rayTracingInvocationReorderProps->rayTracingInvocationReorderReorderingHint = VK_RAY_TRACING_INVOCATION_REORDER_MODE_REORDER_NV;
+                            }));
+
+                SECTION("GetRaytracingCaps with reorder support and with NV_EXTN_OP_HIT_OBJECT_REORDER_THREAD support returns OK and claims that thread reordering is supported") {
+                    ALLOW_CALL(device, IsNvShaderExtnOpCodeSupported(static_cast<UINT32>(NV_EXTN_OP_HIT_OBJECT_REORDER_THREAD)))
+                        .RETURN(true);
+
+                    REQUIRE(NvAPI_Initialize() == NVAPI_OK);
+
+                    NVAPI_D3D12_RAYTRACING_THREAD_REORDERING_CAPS caps;
```
Saancreed: Maybe initialize this to the other value to make sure it's actually written to by the call?

tests/nvapi_d3d12.cpp
```diff
@@ -651,42 +703,108 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
...
+                    REQUIRE(NvAPI_D3D12_GetRaytracingCaps(static_cast<ID3D12Device*>(&device), NVAPI_D3D12_RAYTRACING_CAPS_TYPE_THREAD_REORDERING, &caps, sizeof(caps)) == NVAPI_OK);
+                    REQUIRE(caps == NVAPI_D3D12_RAYTRACING_THREAD_REORDERING_CAP_STANDARD);
+                }
+
+                SECTION("GetRaytracingCaps with reorder support but without NV_EXTN_OP_HIT_OBJECT_REORDER_THREAD support returns OK and claims that thread reordering is not supported") {
+                    ALLOW_CALL(device, IsNvShaderExtnOpCodeSupported(static_cast<UINT32>(NV_EXTN_OP_HIT_OBJECT_REORDER_THREAD)))
+                        .RETURN(false);
+
+                    REQUIRE(NvAPI_Initialize() == NVAPI_OK);
```
Saancreed, line 734: Can this call be moved to the outer `SECTION`?

I'd ask the same about other instances of `NvAPI_Initialize`, but I imagine Vk props have to be configured before them.

tests/nvapi_d3d12.cpp
```diff
@@ -181,9 +184,34 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
...
-        REQUIRE(supported == false);
+        // DXVK_NVAPI_D3D12_NV_SHADER_EXTN = 1 is set in section starting listener
+
+        SECTION("IsNvShaderExtnOpCodeSupported without VKD3D-Proton support") {
+            ALLOW_CALL(device, QueryInterface(__uuidof(ID3D12DeviceExt4), _))
+                .RETURN(E_NOTIMPL);
+
+            auto supported = true;
+            REQUIRE(NvAPI_D3D12_IsNvShaderExtnOpCodeSupported(&device, NV_EXTN_OP_HIT_OBJECT_REORDER_THREAD, &supported) == NVAPI_OK);
+            REQUIRE(supported == false);
```
Saancreed: `REQUIRE_FALSE`?

Though it might be too big of a change if the existing code prefers `== false`, in which case consistency wins, I guess.

Saancreed (approved):
Just some questions and small suggestions.

tests/nvapi_d3d12.cpp
jp7677: Unfortunately yes, otherwise compiler hits `warning: comparison of integer expressions of different signedness: 'const int' and 'unsigned int'`

tests/nvapi_d3d12.cpp
jp7677: Mock configuration has to happen indeed before because of `std::move` inside  `NvAPI_Initialize()`

tests/nvapi_d3d12.cpp
jp7677: good idea to use this one more often (we already do). addressed at a couple of places separately,

PR #362, opened 2026-05-08
includes: Update vkroots to ee76e62

jp7677:
This addresses https://github.com/misyltoad/vkroots/issues/15 and allows our Reflex layer to be used together with the gamescope WSI layer. Reflex and HDR for Indy!


Saancreed (approved):
🎉

PR #367, opened 2026-05-25
Implement  NvAPI_DISP_GetDisplayConfig

### LINKED ISSUE #366: [Feature Request] Implement `NvAPI_DISP_GetDisplayConfig`
Sound Voltex (Konami arcade game) calls `NvAPI_DISP_GetDisplayConfig` to retrieve refresh rate info to set the gameplay frame rate. If this call fails it falls back to a dummy value of 60Hz, resulting in a weird mix of high FPS UI elements but choppy gameplay when playing on 120Hz.

Currently (on Linux) we have to apply a hex edit to workaround this issue by hardcoding the refresh rate value to 60/120 FPS.

It also calls some other APIs:
* `0xad298d3f` (private?)
* `NvAPI_Initialize` (obviously)
* `0x33c7358c`
* `0x593e8644`
* `NvAPI_EnumPhysicalGPUs`
* `NvAPI_GPU_GetConnectedDisplayIds`
* `NvAPI_DISP_SetDisplayConfig` (fine to ignore)
* `NvAPI_DISP_GetGDIPrimaryDisplayId`

Some more info here when I had a similar issue in Windows land: https://github.com/spice2x/spice2x.github.io/issues/107

jp7677:
~Very (very) early not tested draft.~ still needs some testing but it's getting somewhere.

With inspiration from on https://github.com/NVIDIA/nvapi/blob/main/Sample_Code/DisplayConfiguration/DisplayConfiguration.cpp


src/nvapi/nvapi_output.cpp
```diff
@@ -12,41 +12,39 @@ namespace dxvk {
...
-                    m_colorData.BitsPerColor = NV_BPC_10;
-                    break;
-                case 12:
-                    m_colorData.BitsPerColor = NV_BPC_12;
-                    break;
-                case 16:
-                    m_colorData.BitsPerColor = NV_BPC_16;
-                    break;
-            }
+            m_colorData.BitsPerColor = desc1.BitsPerColor;
```
jp7677, line 47: Turns out this is a hardcoded value in DXVK https://github.com/doitsujin/dxvk/blob/master/src/dxgi/dxgi_output.cpp#L238

src/nvapi/nvapi_output.cpp
```diff
@@ -12,41 +12,39 @@ namespace dxvk {
...
         DXGI_OUTPUT_DESC desc{};
-        dxgiOutput->GetDesc(&desc);
+        dxgiOutput->GetDesc(&desc); // TODO: Check return code
 
         m_deviceName = str::fromws(desc.DeviceName);
         log::info(str::format("NvAPI Output: ", m_deviceName));
 
+        m_displayMode.Top = desc.DesktopCoordinates.top;
+        m_displayMode.Left = desc.DesktopCoordinates.left;
+        // DXVK reports no rotation
```
jp7677, line 22: We could add support for rotation on our side just in case that DXVK correctly reports this in the future. But i guess for our use cases that's just a setup we can safely ignore.

src/nvapi_disp.cpp
```diff
@@ -252,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+        auto output = outputs[i];
+
+        switch (pathInfo[0].version) {
+            case NV_DISPLAYCONFIG_PATH_INFO_VER1: {
+                auto& pathInfoV1 = reinterpret_cast<NV_DISPLAYCONFIG_PATH_INFO_V1*>(pathInfo)[i];
+                pathInfoV1.version = NV_DISPLAYCONFIG_PATH_INFO_VER1;
+                pathInfoV1.reserved_sourceId = 0;
+
+                if (!pathInfoV1.targetInfo)
+                    pathInfoV1.targetInfoCount = 1; // Just single or extended mode, ignore clone/mirror display setups
```
jp7677, line 345: So yeah, no valid reporting for clone setups.

src/nvapi_disp.cpp
```diff
@@ -252,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+
+    if (!pathInfo) {
+        *pathInfoCount = outputs.size();
+        return Ok(n);
+    }
+
+    for (auto i = 0U; i < *pathInfoCount; i++) {
+        auto output = outputs[i];
+
+        switch (pathInfo[0].version) {
```
jp7677, line 324: that's a bit lazy, but avoids to hassle with different array types.

src/nvapi_disp.cpp
```diff
@@ -268,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+
+        for (auto j = 0U; j < nvapiAdapterRegistry->GetOutputCount(adapter); j++) {
+            outputs.push_back(nvapiAdapterRegistry->GetOutput(adapter, j));
+        }
+    }
+
+    if (!pathInfo) {
+        *pathInfoCount = outputs.size();
+        return Ok(n);
+    }
```
Saancreed, line 312:
```suggestion
    } else if (*pathInfoCount < outputs.size()) {
        *pathInfoCount = outputs.size();
        return InsufficientBuffer(n);
    }
```

Worth checking how native behaves if given a buffer that's too small. In theory the case of `count == 0 && pathInfo != nullptr` should result in `NVAPI_INVALID_ARGUMENT` but that sounds a little weird.

src/nvapi_disp.cpp
```diff
@@ -268,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+
+    for (auto i = 0U; i < *pathInfoCount; i++) {
+        auto output = outputs[i];
+
```
Saancreed, line 323:
```suggestion
        auto& displayMode = output->GetDisplayMode();

```

To simplify assignments below?

src/nvapi_disp.cpp
```diff
@@ -268,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+                        *pathInfoV1.sourceModeInfo = {};
+                        pathInfoV1.sourceModeInfo->resolution.width = output->GetDisplayMode().Width;
+                        pathInfoV1.sourceModeInfo->resolution.height = output->GetDisplayMode().Height;
+                        pathInfoV1.sourceModeInfo->resolution.colorDepth = BitsPerColorToColorDepth(output->GetColorData().BitsPerColor);
+                        pathInfoV1.sourceModeInfo->colorFormat = NV_FORMAT_UNKNOWN;
+                        pathInfoV1.sourceModeInfo->position.x = output->GetDisplayMode().Left;
+                        pathInfoV1.sourceModeInfo->position.y = output->GetDisplayMode().Top;
+                        pathInfoV1.sourceModeInfo->spanningOrientation = NV_DISPLAYCONFIG_SPAN_NONE;
+                        pathInfoV1.sourceModeInfo->bGDIPrimary = output->IsPrimary();
+                    }
```
Saancreed: Why is `sourceModeInfo` assigned only if `targetInfo` was provided? The docs explicitly call out that the third pass for `targetInfo` is optional, which presumably means that it should be possible to fetch `sourceModeInfo` on the second call.

Same below.

src/nvapi_disp.cpp
```diff
@@ -268,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+
+                    pathInfoV1.targetInfo[0].displayId = output->GetId();
+                    if (pathInfoV1.targetInfo[0].details) {
+                        *pathInfoV1.targetInfo[0].details = {};
+                        pathInfoV1.targetInfo[0].details->version = NV_DISPLAYCONFIG_PATH_ADVANCED_TARGET_INFO_VER1;
+                        pathInfoV1.targetInfo[0].details->refreshRate1K = output->GetDisplayMode().RefreshRate * 1000;
+                        pathInfoV1.targetInfo[0].details->rotation = NV_ROTATE_0;
+                        pathInfoV1.targetInfo[0].details->scaling = NV_SCALING_DEFAULT;
```
Saancreed: Interestingly, most members of this struct are marked as `(IN)` in the headers, so I'm not totally sure if we're supposed to set them.

src/nvapi_disp.cpp
```diff
@@ -268,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+                        pathInfoV1.sourceModeInfo->position.x = output->GetDisplayMode().Left;
+                        pathInfoV1.sourceModeInfo->position.y = output->GetDisplayMode().Top;
+                        pathInfoV1.sourceModeInfo->spanningOrientation = NV_DISPLAYCONFIG_SPAN_NONE;
+                        pathInfoV1.sourceModeInfo->bGDIPrimary = output->IsPrimary();
+                    }
+
+                    pathInfoV1.targetInfo[0].displayId = output->GetId();
+                    if (pathInfoV1.targetInfo[0].details) {
+                        *pathInfoV1.targetInfo[0].details = {};
+                        pathInfoV1.targetInfo[0].details->version = NV_DISPLAYCONFIG_PATH_ADVANCED_TARGET_INFO_VER1;
```
Saancreed, line 357: Check the input version first? Or are we not supposed to?

Same below.

src/nvapi_disp.cpp
```diff
@@ -268,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+                }
+
+                break;
+            }
+            default:
+                return IncompatibleStructVersion(n, pathInfo->version);
+        }
+    }
+
+    return Ok(n);
```
Saancreed, line 417:
```suggestion
    *pathInfoCount = outputs.size();
    return Ok(n);
```

Or are we not supposed to set this to the final number of path infos we filled, in case we were given a larger than necessary buffer?

src/nvapi_disp.cpp
```diff
@@ -268,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+                        pathInfoV1.sourceModeInfo->resolution.height = output->GetDisplayMode().Height;
+                        pathInfoV1.sourceModeInfo->resolution.colorDepth = BitsPerColorToColorDepth(output->GetColorData().BitsPerColor);
+                        pathInfoV1.sourceModeInfo->colorFormat = NV_FORMAT_UNKNOWN;
+                        pathInfoV1.sourceModeInfo->position.x = output->GetDisplayMode().Left;
+                        pathInfoV1.sourceModeInfo->position.y = output->GetDisplayMode().Top;
+                        pathInfoV1.sourceModeInfo->spanningOrientation = NV_DISPLAYCONFIG_SPAN_NONE;
+                        pathInfoV1.sourceModeInfo->bGDIPrimary = output->IsPrimary();
+                    }
+
+                    pathInfoV1.targetInfo[0].displayId = output->GetId();
```
Saancreed, line 351: Shouldn't we check `targetInfoCount` first? Or set it to `1` afterwards?

src/nvapi_disp.cpp
Saancreed: I strongly doubt that it's legal to pass a pointer to an area of memory containing multiple path infos of different versions. Still, I'd check that `pathInfo[i].version == pathInfo[0].version` and error out if it's not.

src/nvapi_disp.cpp
```diff
@@ -268,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+                }
+
+                break;
+            }
+            case NV_DISPLAYCONFIG_PATH_INFO_VER2: {
+                pathInfo[i].version = NV_DISPLAYCONFIG_PATH_INFO_VER2;
+                pathInfo[i].sourceId = 0;
+                pathInfo[i].IsNonNVIDIAAdapter = !output->GetParent()->HasNvProprietaryDriver();
+                pathInfo[i].reserved = 0;
+                pathInfo[i].pOSAdapterID = nullptr;
```
Saancreed: In theory we should set it for non-NV adapters… or write to it if not `nullptr`? Idk, it's all pretty confusing.

src/nvapi_disp.cpp
```diff
@@ -268,3 +284,111 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+                        pathInfoV1.targetInfo[0].details->connector = NVAPI_GPU_CONNECTOR_UNKNOWN;
+                    }
+                }
+
+                break;
+            }
+            case NV_DISPLAYCONFIG_PATH_INFO_VER2: {
+                pathInfo[i].version = NV_DISPLAYCONFIG_PATH_INFO_VER2;
+                pathInfo[i].sourceId = 0;
+                pathInfo[i].IsNonNVIDIAAdapter = !output->GetParent()->HasNvProprietaryDriver();
```
Saancreed: `&& !HasNvkDriver()`?

tests/nvapi_system.cpp
```diff
@@ -456,6 +458,48 @@ TEST_CASE("Sysinfo methods succeed against local system", "[system]") {
...
+
+    NvU32 pathInfoCount{};
+    CHECK(nvAPI_DISP_GetDisplayConfig(&pathInfoCount, nullptr) == NVAPI_OK);
+
+    if (pathInfoCount > 0) {
+        auto pathInfo = std::vector<NV_DISPLAYCONFIG_PATH_INFO>(pathInfoCount);
+        for (auto& info : pathInfo)
+            info.version = NV_DISPLAYCONFIG_PATH_INFO_VER;
+
+        CHECK(nvAPI_DISP_GetDisplayConfig(&pathInfoCount, pathInfo.data()) == NVAPI_OK);
```
Saancreed: If `pathInfoCount` could be modified by this call (something to check…) then we should do a `i = 0; i < pathInfoCount; i++` loop instead. Or just resize the vector after this call to `pathInfoCount`.

tests/nvapi_system.cpp
```diff
@@ -456,6 +458,48 @@ TEST_CASE("Sysinfo methods succeed against local system", "[system]") {
...
+
+        for (auto& info : pathInfo) {
+            info.sourceModeInfo = new NV_DISPLAYCONFIG_SOURCE_MODE_INFO_V1;
+            info.targetInfo = new NV_DISPLAYCONFIG_PATH_TARGET_INFO_V2[info.targetInfoCount];
+            for (auto j = 0U; j < info.targetInfoCount; j++) {
+                info.targetInfo[j].details = new NV_DISPLAYCONFIG_PATH_ADVANCED_TARGET_INFO;
+                info.targetInfo[j].details->version = NV_DISPLAYCONFIG_PATH_ADVANCED_TARGET_INFO_VER;
+            }
```
Saancreed: Can we use vectors instead? Sure, `targetInfo`s make it a bit awkward, but I think it's worth saving on explicit cleanup with `delete`s later.

Saancreed (commented):
More than I expected, but this function has quite a lot going on so I'm not too surprised.

src/nvapi_disp.cpp
jp7677: I've added both, it's also consistent now with `pathInfoCount`

tests/nvapi_system.cpp
jp7677: I've added an explicit REQUIRE.

src/nvapi_disp.cpp
jp7677: ah, thanks. missed that one.

src/nvapi_disp.cpp
jp7677: On the other hand, fetch the refresh rate (which is also `IN`) is the initial motivation to add this. pretty confusing indeed.

src/nvapi_disp.cpp
jp7677: That is quite a tricky one. The last commit just ignores this field. If it is `nullptr`, if there really is a valid LUID behind it, it just stays default. 
I did a very brief test on Windows, NVAPI from nv didn't even return a `pathInfo` for a non-nv GPU, only the display connected to the nv GPU is reported (intel/nv laptop). So i wonder if this is only relevant for `NvAPI_DISP_SetDisplayConfig` but not for `NvAPI_DISP_GetDisplayConfig`.

src/nvapi_disp.cpp
```diff
@@ -252,3 +284,145 @@ NVAPI_FUNCTION NvAPI_DISP_GetGDIPrimaryDisplayId(NvU32* displayId) {
...
+                // Ignore pOSAdapterID
+
+                if (pathInfo[i].sourceModeInfo) {
+                    *pathInfo[i].sourceModeInfo = {};
+                    pathInfo[i].sourceModeInfo->resolution.width = displayMode.Width;
+                    pathInfo[i].sourceModeInfo->resolution.height = displayMode.Height;
+                    pathInfo[i].sourceModeInfo->resolution.colorDepth = BitsPerColorToColorDepth(colorData.BitsPerColor);
+                    pathInfo[i].sourceModeInfo->colorFormat = NV_FORMAT_UNKNOWN;
+                    pathInfo[i].sourceModeInfo->position.x = displayMode.Left;
+                    pathInfo[i].sourceModeInfo->position.x = displayMode.Left;
```
Saancreed: Unneecessary double assignment?

tests/nvapi_sysinfo_topo.cpp
```diff
@@ -337,6 +337,149 @@ TEST_CASE("Topology methods succeed", "[.sysinfo-topo]") {
...
+    SECTION("GetDisplayConfig succeeds") {
+        SECTION("GetDisplayConfig (V1) returns OK") {
+            NvU32 pathInfoCount{};
+            REQUIRE(NvAPI_DISP_GetDisplayConfig(&pathInfoCount, nullptr) == NVAPI_OK);
+            REQUIRE(pathInfoCount == 3);
+
+            auto pathInfo = std::vector<NV_DISPLAYCONFIG_PATH_INFO_V1>(pathInfoCount);
+            for (auto& info : pathInfo) {
+                info.version = NV_DISPLAYCONFIG_PATH_INFO_VER1;
+                info.sourceModeInfo = new NV_DISPLAYCONFIG_SOURCE_MODE_INFO_V1;
```
Saancreed: Quite a few `new`s in this file that don't appear to have corresponding `delete`s.

Saancreed (commented):
Last few things I think, LGTM otherwise. Still not a fan of `new` in test code though.

tests/nvapi_sysinfo_topo.cpp
jp7677: Sry, time run out, still want to look at avoiding `new`. That said, I'm counting 3 new and 3 delete?

Saancreed: In the other file, yes. In this one, I see 6 `new`s and no `delete`s. Unless they aren't part of the diff?

jp7677: derp :). Fixed for now and will still look into a (hopefully) more elegant solution.

jp7677: Added a commit that avoids new/delete and hides some of the awkwardness.  I guess there is something to say for both versions.

jp7677: mmh, there are some deconstruction issues that I don't see yet...

Edit: turns out using pointer of elements in a vector is a bad idea, so nice thought, but just doesn't work. I suggest I just drop the last commit and keep the new/delete's. May be not that nice, but closer to the example from the NVAPI headers repo.

jp7677: I think I got it now. Still not absolutely sure if we should keep the last commit.

Saancreed (approved):

Issue #171, opened 2024-04-05
War Thunder enabling DLSS = freeze


brokeDude2901:
Hello I'm reporting War Thunder problem with DLSS enabled ( will play fine without DLSS ):
`Proton Experimental / X11 / Nvidia RTX 2080 / Ubuntu 23.10`
Launch command:
`PROTON_LOG=1 PROTON_FORCE_NVAPI=1 DXVK_NVAPI_LOG_LEVEL=info DXVK_NVAPI_LOG_PATH=/home/user/ %command%`
[dxvk-nvapi.log](https://github.com/jp7677/dxvk-nvapi/files/14885780/dxvk-nvapi.log)

jp7677:
Thanks for the report and sorry for the late response. I'm awaiting feedback from another source regarding your report, that takes a bit longer, hence the delay.
The nvapi log file looks quite normal, I don't think the unimplemented functions would cause a hang, but not sure though. Do you also have a proton log available?

jp7677:
> Not sure if other people are having the same issue

Unfortunately yes, also trying a few obvious things didn't mitigate this. What we can do, is preventing DLSS from getting enabled, so that it at least no longer hangs/crashes. https://github.com/jp7677/dxvk-nvapi/pull/172 should do this.

jp7677:
Just for reference, I've looked again from the `nvapi` side of things. Installing `nvml` doesn't help and also faking the Reflex calls (this is a D3D11 title, we only have Reflex for D3D12/VKD3D-P) also didn't made any difference, the title still crashes.
Adding the remaining missing functions (https://github.com/jp7677/dxvk-nvapi/pull/176), as far as known to us, also made no difference.

SveSop:
I did some tests with the WIP/Test PR for reflex in DXVK [https://github.com/doitsujin/dxvk/pull/3690](https://github.com/doitsujin/dxvk/pull/3690) and it seems to work fine for this game - meaning no errors or crashes. Not really possible (for me) to determine if it improves latency.

Enabling DLSS still crashes with the same D3D11 texture error, so i would say it has nothing to do with the Reflex settings.

```
err:   D3D11: Cannot create texture:
err:     Format:  39
err:     Extent:  1280x800x1
err:     Samples: 1
err:     Layers:  1
err:     Levels:  1
err:     Usage:   e8
err:     Flags:   0
```
[...]

PR #370, opened 2026-06-03
util: Remove Pascal spoofing for War Thunder

jp7677:
Looks like the earlier crashes with D3D11 i.c.w. DLSS have been fixed in the meantime.

Screenshot From 2026-06-03 20-20-52

Originally added in https://github.com/jp7677/dxvk-nvapi/pull/172 based on https://github.com/jp7677/dxvk-nvapi/issues/171


Saancreed (approved):
Nice!

PR #371, opened 2026-06-15
NVAPI support for OMM

### LINKED ISSUE #123: Support Nvidia new OMM sample..

chrisd-work:
This adds OMM support for NVAPI. This takes a different approach than https://github.com/jp7677/dxvk-nvapi/pull/125. However 125 was a major source of inspiration. Notably this approach put most of the burden on dxvk-nvapi to convert the data, and then pass it along into vkd3d-proton.

There is still two call paths 1 for DXR1.2 which requires SER as well. Whereas NVAPI is more permissive.

Please look at https://github.com/HansKristian-Work/vkd3d-proton/pull/3005 for the VKD3D side of these changes. I have tried to keep the changes atomic. I am willing to split the PR into phases if needed.

### DIFF
--- a/src/nvapi/nvapi_as_convert.cpp
+++ b/src/nvapi/nvapi_as_convert.cpp
@@ -1,0 +1,246 @@
+#include "nvapi_as_convert.h"
+#include "../util/util_log.h"
+#include "../util/util_statuscode.h"
+
+namespace dxvk {
+
+    NvAPI_Status BuildFlagsToD3d12(D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS& d3d12Flags,
+        NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS_EX nvapiFlags,
+        bool supportsOmm) {
+        constexpr auto n = __func__;
+        /* All currently known flags */
+        const NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS_EX allFlags =
+            static_cast<NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS_EX>(
+                NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_UPDATE_EX | NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_COMPACTION_EX | NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE_EX | NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_BUILD_EX | NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_MINIMIZE_MEMORY_EX | NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE_EX | NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_UPDATE_EX | NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DISABLE_OMMS_EX | NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_OPACITY_STATES_UPDATE_EX | NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DATA_ACCESS_EX);
+
+        // D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DATA_ACCESS is only
+        // defined in DirectX-Headers preview-tag releases (added in v1.717.0-preview).
+        // The vendored submodule tracks stable tags only, so we don't pull from preview,
+        // and we carry this provisional 0x100 (the value the preview header defines) until
+        // the flag ships in a stable DirectX-Headers tag and the submodule bump pulls it in.
+        static constexpr D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS
+            D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DATA_ACCESS_PROVISIONAL =
+                static_cast<D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS>(0x100);
+
+        if (nvapiFlags & ~allFlags) {
+            log::info(str::format("Unsupported NVAPI build flags: ", nvapiFlags));
+            return InvalidArgument(n);
+        }
+
+        d3d12Flags = D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_NONE;
+
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_UPDATE_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_UPDATE;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_COMPACTION_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_COMPACTION;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_BUILD_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_BUILD;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_MINIMIZE_MEMORY_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_MINIMIZE_MEMORY;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_UPDATE_EX) {
+            if (!supportsOmm)
+                log::info("Ignoring NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_UPDATE_EX OMM not supported");
+            else
+                d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_UPDATE;
+        }
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DISABLE_OMMS_EX) {
+            if (!supportsOmm)
+                log::info("Ignoring NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DISABLE_OMMS_EX OMM not supported");
+            else
+                d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DISABLE_OMMS;
+        }
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_OPACITY_STATES_UPDATE_EX) {
+            /* OPACITY_STATES_UPDATE_EX is a narrower data-only variant of ALLOW_OMM_UPDATE_EX. D3D12 only has the wide
+             * ALLOW_OMM_UPDATE, so widen rather than drop. */
+            if (!supportsOmm)
+                log::info("Ignoring NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_OPACITY_STATES_UPDATE_EX OMM not supported");
+            else
+                d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_UPDATE;
+        }
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DATA_ACCESS_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DATA_ACCESS_PROVISIONAL;
+
+        return Ok(n);
+    }
+
+    NvAPI_Status GeomDescTypeToD3d12(D3D12_RAYTRACING_GEOMETRY_TYPE& d3d12Type,
+        NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_EX nvapiType) {
+        constexpr auto n = __func__;
+        switch (nvapiType) {
+            case NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_TRIANGLES_EX:
+                d3d12Type = D3D12_RAYTRACING_GEOMETRY_TYPE_TRIANGLES;
+                return Ok(n);
+            case NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_PROCEDURAL_PRIMITIVE_AABBS_EX:
+                d3d12Type = D3D12_RAYTRACING_GEOMETRY_TYPE_PROCEDURAL_PRIMITIVE_AABBS;
+                return Ok(n);
+            case NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_OMM_TRIANGLES_EX:
+                d3d12Type = D3D12_RAYTRACING_GEOMETRY_TYPE_OMM_TRIANGLES;
+                return Ok(n);
+            case NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_DMM_TRIANGLES_EX:
+                log::info("Unsupported NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_DMM_TRIANGLES_EX");
+                return NotSupported(n);
+            case NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_SPHERES_EX:
+                log::info("Unsupported NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_SPHERES_EX");
+                return NotSupported(n);
+            case NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_LSS_EX:
+                log::info("Unsupported NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_LSS_EX");
+                return NotSupported(n);
+            default:
+                log::info(str::format("Unsupported NVAPI_D3D12_RAYTRACING_GEOMETRY_TYPE_ ", nvapiType));
+                return InvalidArgument(n);
+        }
+    }
+
+    const NVAPI_D3D12_RAYTRACING_GEOMETRY_DESC_EX& GeomDesc(
+        const NVAPI_D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS_EX& nvapiDesc, NvU32 index) {
+        if (nvapiDesc.descsLayout == D3D12_ELEMENTS_LAYOUT_ARRAY_OF_POINTERS)
+            return *nvapiDesc.ppGeometryDescs[index];
+        return *reinterpret_cast<const NVAPI_D3D12_RAYTRACING_GEOMETRY_DESC_EX*>(
+            reinterpret_cast<const uint8_t*>(nvapiDesc.pGeometryDescs) + nvapiDesc.geometryDescStrideInBytes * index);
+    }
+
+    NvAPI_Status OmmArrayBuildFlagsToD3d12(D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS& d3d12Flags,
+        NVAPI_D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_BUILD_FLAGS nvapiFlags) {
+        constexpr auto n = __func__;
+        const NVAPI_D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_BUILD_FLAGS allFlags =
+            static_cast<NVAPI_D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_BUILD_FLAGS>(
+                NVAPI_D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_BUILD_FLAG_PREFER_FAST_TRACE | NVAPI_D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_BUILD_FLAG_PREFER_FAST_BUILD);
+
+        if (nvapiFlags & ~allFlags) {
+            log::info(str::format("Unsupported NVAPI OMM-Array build flags: ", nvapiFlags));
+            return InvalidArgument(n);
+        }
+
+        d3d12Flags = D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_NONE;
+
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_BUILD_FLAG_PREFER_FAST_TRACE)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_BUILD_FLAG_PREFER_FAST_BUILD)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_BUILD;
+
+        return Ok(n);
+    }
+
+    NvAPI_Status OmmArrayInputsToD3d12(D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS& d3d12Inputs,
+        D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_DESC& arrayDesc,
+        const NVAPI_D3D12_BUILD_RAYTRACING_OPACITY_MICROMAP_ARRAY_INPUTS& nvapiInputs) {
+        constexpr auto n = __func__;
+
+        memset(&d3d12Inputs, 0x00, sizeof(d3d12Inputs));
+        memset(&arrayDesc, 0x00, sizeof(arrayDesc));
+
+        /* NVAPI_..._USAGE_COUNT and D3D12_..._HISTOGRAM_ENTRY are byte-compatible:
+         * { UINT count, UINT subdivisionLevel, OMM_FORMAT format }. Field sizes
+         * match (NvU32 == UINT == 4 bytes) and OMM_FORMAT enum values match
+         * (OC1_2_STATE=0x1, OC1_4_STATE=0x2 in both). Cast directly. */
+        arrayDesc.NumOmmHistogramEntries = nvapiInputs.numOMMUsageCounts;
+        arrayDesc.pOmmHistogram = reinterpret_cast<const D3D12_RAYTRACING_OPACITY_MICROMAP_HISTOGRAM_ENTRY*>(
+            nvapiInputs.pOMMUsageCounts);
+        arrayDesc.InputBuffer = nvapiInputs.inputBuffer;
+        arrayDesc.PerOmmDescs = nvapiInputs.perOMMDescs;
+
+        d3d12Inputs.Type = D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_OPACITY_MICROMAP_ARRAY;
+        d3d12Inputs.NumDescs = 1;
+        d3d12Inputs.DescsLayout = D3D12_ELEMENTS_LAYOUT_ARRAY;
+        d3d12Inputs.pOpacityMicromapArrayDesc = &arrayDesc;
+
+        if (NvAPI_Status status = OmmArrayBuildFlagsToD3d12(d3d12Inputs.Flags, nvapiInputs.flags); status != NVAPI_OK)
+            return status;
+
+        return Ok(n);
+    }
+
+    NvAPI_Status NvapiAsConverter::Convert(D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS& d3d12Desc,
+        const NVAPI_D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS_EX& nvapiDesc,
+        bool supportsOmm) {
+        constexpr auto n = __func__;
+        memset(&d3d12Desc, 0x00, sizeof(d3d12Desc));
+
+        d3d12Desc.Type = nvapiDesc.type;
+        d3d12Desc.NumDescs = nvapiDesc.numDescs;
+
+        if (NvAPI_Status status = BuildFlagsToD3d12(d3d12Desc.Flags, nvapiDesc.flags, supportsOmm); status != NVAPI_OK)
+            return status;
+
+        if (nvapiDesc.type == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_TYPE_TOP_LEVEL) {
+            d3d12Desc.DescsLayout = nvapiDesc.descsLayout;
+            d3d12Desc.InstanceDescs = nvapiDesc.instanceDescs;
+        } else {
+            const bool usesStride = nvapiDesc.descsLayout != D3D12_ELEMENTS_LAYOUT_ARRAY_OF_POINTERS;
+
+            Reserve(nvapiDesc.numDescs);
+
+            d3d12Desc.DescsLayout = D3D12_ELEMENTS_LAYOUT_ARRAY;
+            d3d12Desc.pGeometryDescs = m_geometryDescs;
+
+            for (auto i = 0U; i < nvapiDesc.numDescs; i++) {
+                const NVAPI_D3D12_RAYTRACING_GEOMETRY_DESC_EX& nvapiGeomDesc = GeomDesc(nvapiDesc, i);
+                D3D12_RAYTRACING_GEOMETRY_DESC& d3d12GeomDesc = m_geometryDescs[i];
+                D3D12_RAYTRACING_GEOMETRY_OMM_LINKAGE_DESC& d3d12OpacityMicromapLinkageDesc =
+                    m_opacityMicromapLinkageDescs[i];
+
+                d3d12GeomDesc.Flags = nvapiGeomDesc.flags;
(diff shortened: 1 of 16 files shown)


Saancreed:
While I like the general direction of this, my comments from linked vkd3d-proton PR still apply. Notably:

* I'd strongly prefer if `SetOpacityMicromapOverrideNVAPI` was `SetCreatePipelineStateFlags(NVAPI)`.
* I'm not yet sure if `GetRaytracingAccelerationStructurePrebuildInfoNVAPI`, `BuildRaytracingAccelerationStructureNVAPI` and `EmitRaytracingAccelerationStructurePostbuildInfoNVAPI` will stay or get dropped, pending Hans-Kristian's review.
* `D3D12_VK_EXT_OPACITY_MICROMAP` is technically less correct than `D3D12_VK_KHR_OPACITY_MICROMAP` would be.

And because we're the ones depending on vkd3d-proton and not the other way around, this should wait until vkd3d-proton's side is in place.

Other than that, your changes seem to have caused some assertion failures in tests, mostly around QueryInterface usage, see https://github.com/jp7677/dxvk-nvapi/actions/runs/27529933594/job/81365310936 (and expand "Present test results").

chrisd-work:
> I'd strongly prefer if SetOpacityMicromapOverrideNVAPI was SetCreatePipelineStateFlags(NVAPI).

Will rename.

> I'm not yet sure if GetRaytracingAccelerationStructurePrebuildInfoNVAPI, BuildRaytracingAccelerationStructureNVAPI and EmitRaytracingAccelerationStructurePostbuildInfoNVAPI will stay or get dropped, pending Hans-Kristian's review.

The historical contract is that NVAPI OMM and DXR 1.2 are decoupled: you can ship NVAPI OMM without DXR 1.2. Treating DXR 1.2 as a precondition for the NVAPI path would collapse a lot of code in both repos and is tempting, but full DXR 1.2 hasn't landed in vkd3d-proton yet, so adopting that simplification today means gating the NVAPI path on something that doesn't exist. Worth pinning down with Hans before either side merges.

The counter argument to using DXR 1.2 as a gate for NVAPI OMM is that it simplifies the code mainly on VKD3D-proton, but a bit on dxvk-nvapi as well.

> D3D12_VK_EXT_OPACITY_MICROMAP is technically less correct than D3D12_VK_KHR_OPACITY_MICROMAP would be.

Will rename. Better still: drop the VK_EXT/VK_KHR prefix entirely. We've already removed every other mention of Vulkan extension names from the user-facing surface; keeping it on this identifier is inconsistent.

> And because we're the ones depending on vkd3d-proton and not the other way around, this should wait until vkd3d-proton's side is in place.

Both codebases are designed to work independently of the other, so I don't have a hard ordering preference; that's the maintainers' call. What I do feel strongly about is reviewing both sides jointly before either merges. If vkd3d-proton lands first and this review then surfaces a structural issue here, reverting is more expensive than the wait.

> Other than that, your changes seem to have caused some assertion failures in tests, mostly around QueryInterface usage, see https://github.com/jp7677/dxvk-nvapi/actions/runs/27529933594/job/81365310936 (and expand "Present test results").

I will take a look.

Saancreed:
> > I'd strongly prefer if SetOpacityMicromapOverrideNVAPI was SetCreatePipelineStateFlags(NVAPI).
> 
> Will rename.

It's less about rename and more about changing `BOOL` to some flags type (even if it would be with only one defined flag value for now). See my comment on the other PR for a more complete suggestion.

> The historical contract is that NVAPI OMM and DXR 1.2 are decoupled: you can ship NVAPI OMM without DXR 1.2. Treating DXR 1.2 as a precondition for the NVAPI path would collapse a lot of code in both repos and is tempting, but full DXR 1.2 hasn't landed in vkd3d-proton yet, so adopting that simplification today means gating the NVAPI path on something that doesn't exist. Worth pinning down with Hans before either side merges.
> 
> The counter argument to using DXR 1.2 as a gate for NVAPI OMM is that it simplifies the code mainly on VKD3D-proton, but a bit on dxvk-nvapi as well.

Right, makes sense, which is why I don't mind them staying. But there's a chance we could sacrifice a bit of correctness for simplicity by actually removing DXR 1.2 requirement from DXR OMM. One case where it would break are theoretical future games that rely on DXR OMM calls actually failing if other parts of DXR 1.2 are missing, although maybe it's a fine tradeoff. Up to Hans-Kristian to decide though, if he prefers duplicated methods with stricter validation on DXR ones, that's fine too.

> Will rename. Better still: drop the VK_EXT/VK_KHR prefix entirely. We've already removed every other mention of Vulkan extension names from the user-facing surface; keeping it on this identifier is inconsistent.

Sounds good.

> Both codebases are designed to work independently of the other, so I don't have a hard ordering preference; that's the maintainers' call. What I do feel strongly about is reviewing both sides jointly before either merges. If vkd3d-proton lands first and this review then surfaces a structural issue here, reverting is more expensive than the wait.

:+1:

By the way, thanks for pushing this through :heart:

chrisd-work:
> >> The historical contract is that NVAPI OMM and DXR 1.2 are decoupled: you can ship NVAPI OMM without DXR 1.2. Treating DXR 1.2 as a precondition for the NVAPI path would collapse a lot of code in both repos and is tempting, but full DXR 1.2 hasn't landed in vkd3d-proton yet, so adopting that simplification today means gating the NVAPI path on something that doesn't exist. Worth pinning down with Hans before either side merges.

>>    The counter argument to using DXR 1.2 as a gate for NVAPI OMM is that it simplifies the code mainly on VKD3D-proton, but a bit on dxvk-nvapi as well.

> Right, makes sense, which is why I don't mind them staying. But there's a chance we could sacrifice a bit of correctness for simplicity by actually removing DXR 1.2 requirement from DXR OMM. One case where it would break are theoretical future games that rely on DXR OMM calls actually failing if other parts of DXR 1.2 are missing, although maybe it's a fine tradeoff. Up to Hans-Kristian to decide though, if he prefers duplicated methods with stricter validation on DXR ones, that's fine too.

Alternative path: key OMM-on off SetCreatePipelineStateFlags(NVAPI). The flag is the explicit NVAPI signal from the app, so we can use it as the override that forces OMM on even when full DXR 1.2 isn't available. That lets us drop GetRaytracingAccelerationStructurePrebuildInfoNVAPI, BuildRaytracingAccelerationStructureNVAPI, and EmitRaytracingAccelerationStructurePostbuildInfoNVAPI entirely.

Downside: behavior for D3D12 apps changes subtly (DXR OMM calls no longer fail when other DXR 1.2 features are missing, provided the app set the NVAPI flag). The case where that matters is fringe; very unlikely a production app would hit it.

HansKristian-Work:
Discussed offline, I'd be happy to check for just OMM being supported rather than full DXR 1.2 being exposed. If it helps avoid code duplication in vkd3d-proton that seems like the best option.

Saancreed:
> Alternative path: key OMM-on off SetCreatePipelineStateFlags(NVAPI). The flag is the explicit NVAPI signal from the app, so we can use it as the override that forces OMM on even when full DXR 1.2 isn't available.

Couldn't applications in theory call `NvAPI_D3D12_GetRaytracingOpacityMicromapArrayPrebuildInfo` with OMM input before calling `SetCreatePipelineStateFlags`? It would fall apart then, I think…

> Downside: behavior for D3D12 apps changes subtly (DXR OMM calls no longer fail when other DXR 1.2 features are missing, provided the app set the NVAPI flag). The case where that matters is fringe; very unlikely a production app would hit it.

I'd consider this acceptable even if NVAPI flag wasn't set. ~~But again, let's wait and see what Hans-Kristian prefers.~~ Oh, ninja'ed :sweat_smile:

src/nvapi/nvapi_as_convert.cpp
```diff
@@ -0,0 +1,248 @@
...
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_BUILD_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_BUILD;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_MINIMIZE_MEMORY_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_MINIMIZE_MEMORY;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PERFORM_UPDATE;
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_UPDATE_EX) {
+            if (!supportsOmm)
+                log::info("Ignoring NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_UPDATE_EX OMM not supported");
```
Saancreed, line 46: If the implementation doesn't support OMM, isn't it invalid to use these flags anyway? I'd just return something like `NotSupported` immediately if so.

src/nvapi/nvapi_as_convert.cpp
```diff
@@ -0,0 +1,248 @@
...
+    const NVAPI_D3D12_BUILD_RAYTRACING_OPACITY_MICROMAP_ARRAY_INPUTS& nvapiInputs) {
+    constexpr auto n = __func__;
+
+    memset(&d3d12Inputs, 0x00, sizeof(d3d12Inputs));
+    memset(&arrayDesc, 0x00, sizeof(arrayDesc));
+
+    /* NVAPI_..._USAGE_COUNT and D3D12_..._HISTOGRAM_ENTRY are byte-compatible:
+     * { UINT count, UINT subdivisionLevel, OMM_FORMAT format }. Field sizes
+     * match (NvU32 == UINT == 4 bytes) and OMM_FORMAT enum values match
+     * (OC1_2_STATE=0x1, OC1_4_STATE=0x2 in both). Cast directly. */
```
Saancreed: Maybe demonstrate this claim with a few `static_assert`s? On the size of each structure, then size and offset of each member.

src/nvapi_d3d12.cpp
```diff
@@ -847,31 +774,41 @@ NVAPI_FUNCTION NvAPI_D3D12_GetRaytracingAccelerationStructurePrebuildInfoEx(ID3D
...
+    D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_DESC buildDesc;
+    buildDesc.DestAccelerationStructureData = pParams->pDesc->destAccelerationStructureData;
+    if (auto status = commandList->GetAsConverter().convert(buildDesc.Inputs, pParams->pDesc->inputs, supportsOmm);
+        status != NVAPI_OK)
+        return status;
 
-    if (!ConvertBuildRaytracingAccelerationStructureInputs(&pParams->pDesc->inputs, geometryDescs, &desc.Inputs))
-        return NotSupported(n);
+    buildDesc.SourceAccelerationStructureData = pParams->pDesc->sourceAccelerationStructureData;
+    buildDesc.ScratchAccelerationStructureData = pParams->pDesc->scratchAccelerationStructureData;
```
Saancreed: Nitpicking, but we don't have to do the conversion before assigning `SourceAccelerationStructureData` and `ScratchAccelerationStructureData`, the order that we had before was only such because tools like clangd liked to complain about designated initializer initializing members out of order. But if we drop the initializer, then we don't have to keep this split, it looks kind of weird now.

src/nvapi_d3d12.cpp
```diff
@@ -879,6 +879,32 @@ NVAPI_FUNCTION NvAPI_D3D12_SetCreatePipelineStateOptions(ID3D12Device5* pDevice,
...
+        return IncompatibleStructVersion(n, pParams->version);
+
+    auto device = NvapiD3d12Device::GetOrCreate(pDevice);
+    if (!device)
+        return NoImplementation(n, alreadyLoggedNoImplementation);
+
+    if (pParams->serializedDataType == NVAPI_D3D12_SERIALIZED_DATA_RAYTRACING_ACCELERATION_STRUCTURE_EX || (pParams->serializedDataType == NVAPI_D3D12_SERIALIZED_DATA_RAYTRACING_OPACITY_MICROMAP_ARRAY_EX && device->IsOpacityMicromapSupported()))
+        pParams->checkStatus = D3D12_DRIVER_MATCHING_IDENTIFIER_UNRECOGNIZED;
+    else
+        pParams->checkStatus = D3D12_DRIVER_MATCHING_IDENTIFIER_UNSUPPORTED_TYPE;
```
Saancreed: Couldn't we at this point just forward to vkd3d-proton via `ID3D12Device5::CheckDriverMatchingIdentifier`? I know it will return the same thing anyway, but in case serialization is ever implemented (not yet sure how feasible is that), we won't have to update this to support it; instead we'll get it "for free".

src/nvapi_d3d12.cpp
```diff
@@ -905,6 +905,50 @@ NVAPI_FUNCTION NvAPI_D3D12_CheckDriverMatchingIdentifierEx(ID3D12Device5* pDevic
...
+    if (!commandList->IsOpacityMicromapSupported())
+        return NotSupported(n);
+
+    D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_DESC buildDesc{};
+    D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_DESC arrayDesc;
+    buildDesc.DestAccelerationStructureData = pParams->pDesc->destOpacityMicromapArrayData;
+    buildDesc.ScratchAccelerationStructureData = pParams->pDesc->scratchOpacityMicromapArrayData;
+
+    auto postbuildDescs = reinterpret_cast<const D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_DESC*>(
+        pParams->pPostbuildInfoDescs);
```
Saancreed: I don't think this is safe. For `NVAPI_D3D12_RAYTRACING_OPACITY_MICROMAP_ARRAY_POSTBUILD_INFO_TYPE`, `0x0` means `CURRENT_SIZE`, but for `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_TYPE` `0x0` means `COMPACTED_SIZE`, so afaict reinterpret-casting between them will change the meaning.

src/nvapi/nvapi_as_convert.cpp
chrisd-work: The hard rejection is at the geometry-desc level: OMM_TRIANGLES_EX on a non-OMM device returns NotSupported, so any build that actually needs OMM fails there. These flag-level branches only fire on builds with no OMM geometry, where the bits have no effect to begin with; dropping them with an info log is safe. 

This is typical driver behavior: be maximally permissive on hint flags that have no observable effect, and reject only at the point where a feature is actually exercised.

src/nvapi_d3d12.cpp
chrisd-work: will fix.

src/nvapi_d3d12.cpp
chrisd-work: Yeah it make sense. I agree that it may be next to impossible to actually support serialization. OMM actually makes this gap even harder today, not easier.

src/nvapi_d3d12.cpp
chrisd-work: Will fix.

src/nvapi/nvapi_as_convert.cpp
chrisd-work: You can only prove some things with asserts, and you cannot future proof the statement. This is a problem in general. Since it is standard headers they will not change. This is why I avoided added it. It at worse gives you a false sense of assurance. This false sense of assurance exists no matter the shape. You could test that it changes in the future, but then it is O(n). In the end though I am not strongly attached to this approach, and willing to change it.

Saancreed: No need, I'm good with a comment too. As you pointed out, it's only best-effort check anyway.

src/nvapi_d3d12.cpp
chrisd-work: Fixed.

src/nvapi_d3d12.cpp
chrisd-work: Fixed.

src/nvapi_d3d12.cpp
chrisd-work: Fixed.

src/nvapi/nvapi_as_convert.cpp
```diff
@@ -0,0 +1,248 @@
+#include "nvapi_as_convert.h"
+#include "../util/util_log.h"
+#include "../util/util_statuscode.h"
+
+using namespace dxvk;
+
+namespace {
```
jp7677: I would use `namespace dxvk { }` and put everything in it.

src/nvapi/nvapi_as_convert.h
```diff
@@ -0,0 +1,52 @@
...
+        ~NvapiAsConverter() {
+            if (m_geometryDescs != m_geometryDescsStack)
+                delete[] m_geometryDescs;
+            if (m_opacityMicromapLinkageDescs != m_opacityMicromapLinkageDescsStack)
+                delete[] m_opacityMicromapLinkageDescs;
+        }
+        NvapiAsConverter(const NvapiAsConverter&) = delete;
+        NvapiAsConverter& operator=(const NvapiAsConverter&) = delete;
+
+        [[nodiscard]] NvAPI_Status convert(D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS& d3d12Desc,
```
jp7677: `Convert`

src/nvapi/nvapi_as_convert.h
```diff
@@ -0,0 +1,52 @@
...
+            const NVAPI_D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS_EX& nvapiDesc,
+            bool supportsOmm);
+
+      private:
+        static const size_t StackReserve = 16;
+
+        // Convert is destructive.
+        void reserve(size_t reserveSize) {
```
jp7677: `Reserve`

src/nvapi/nvapi_d3d12_device.cpp
```diff
@@ -1,6 +1,6 @@
 #include "nvapi_d3d12_device.h"
-#include "../util/com_pointer.h"
 #include "../util/util_log.h"
+#include "nvapi_as_convert.h"
```
jp7677: not needed here as far as I can see

src/nvapi/nvapi_as_convert.cpp
```diff
@@ -0,0 +1,248 @@
...
+             * ALLOW_OMM_UPDATE, so widen rather than drop. */
+            if (!supportsOmm)
+                log::info("Ignoring NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_OPACITY_STATES_UPDATE_EX OMM not supported");
+            else
+                d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_UPDATE;
+        }
+        if (nvapiFlags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DATA_ACCESS_EX)
+            d3d12Flags |= D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_DATA_ACCESS_PROVISIONAL;
+
+        return Ok(n);
```
jp7677, line 67: We use those statuscode function explicitly for entrypoint functions. To be consistent with other classes/fucntions. returning an `HRESULT` here is more suitable. This applies to all functions here.

src/nvapi/nvapi_as_convert.cpp
```diff
@@ -0,0 +1,248 @@
+#include "nvapi_as_convert.h"
+#include "../util/util_log.h"
+#include "../util/util_statuscode.h"
+
+using namespace dxvk;
+
+namespace {
+
+    NvAPI_Status nvapiBuildFlagsToD3d12(D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS& d3d12Flags,
```
jp7677: Since we are already in nvapi land, `BuildFlagsToD3d12` would be fine too, same for other functions. Within classes we prefer PascalCase.

src/nvapi/nvapi_as_convert.cpp
```diff
@@ -0,0 +1,248 @@
...
+        d3d12Desc.InstanceDescs = nvapiDesc.instanceDescs;
+    } else {
+        const bool usesStride = nvapiDesc.descsLayout != D3D12_ELEMENTS_LAYOUT_ARRAY_OF_POINTERS;
+
+        reserve(nvapiDesc.numDescs);
+
+        d3d12Desc.DescsLayout = D3D12_ELEMENTS_LAYOUT_ARRAY;
+        d3d12Desc.pGeometryDescs = m_geometryDescs;
+
+        for (NvU32 i = 0; i < nvapiDesc.numDescs; i++) {
```
jp7677: Usually we use `auto i = 0U`. no strong preference though.

src/nvapi/nvapi_d3d12_device.h
```diff
@@ -40,10 +49,12 @@ namespace dxvk {
 
         static std::optional<bool> m_cubin64bitSupportAvailable;
 
-        ID3D12DeviceExt4* m_vkd3dDevice{};
+        ID3D12DeviceExt5* m_vkd3dDevice{};
+        uint32_t m_deviceExtTier = 0;
```
jp7677: `m_deviceExtTier` can be a local var in the constructor.

src/nvapi/nvapi_d3d12_graphics_command_list.cpp
```diff
@@ -26,18 +26,28 @@ namespace dxvk {
...
-        m_supportsExtGraphicsCommandList1 = SUCCEEDED(vkd3dCommandList->QueryInterface(IID_PPV_ARGS(&commandListExt1)));
+    NvapiD3d12GraphicsCommandList::NvapiD3d12GraphicsCommandList(ID3D12GraphicsCommandList* commandList,
+        ID3D12GraphicsCommandListExt* vkd3dCommandList)
+        : m_vkd3dGraphicsCommandList(static_cast<ID3D12GraphicsCommandListExt2*>(vkd3dCommandList)) { // NOLINT(*-pro-type-static-cast-downcast)
+        m_commandListExtTier = probeInterfaceChain(vkd3dCommandList, {
+                                                                         __uuidof(ID3D12GraphicsCommandListExt1),
+                                                                         __uuidof(ID3D12GraphicsCommandListExt2),
+                                                                     });
+
+        ID3D12Device* device = nullptr;
```
jp7677: `Com<ID3D12Device> device`? Also I would be OK if we do a `QueryInterface` back to `ID3D12GraphicsCommandList` here instead of passing both `ID3D12GraphicsCommandList*` and         `ID3D12GraphicsCommandListExt*` into the constructor.

src/nvapi/nvapi_d3d12_graphics_command_list.h
```diff
@@ -2,23 +2,39 @@
...
+        [[nodiscard]] bool VerifyOpacityMicromapArrayNVAPI(D3D12_GPU_VIRTUAL_ADDRESS opacity_micromap_array) const;
+
       private:
         static std::unordered_map<ID3D12GraphicsCommandList*, NvapiD3d12GraphicsCommandList> m_nvapiDeviceMap;
         static std::mutex m_mutex;
 
-        ID3D12GraphicsCommandListExt1* m_vkd3dGraphicsCommandList{};
-        bool m_supportsExtGraphicsCommandList1 = false;
+        ID3D12GraphicsCommandListExt2* m_vkd3dGraphicsCommandList{};
+        uint32_t m_commandListExtTier = 0;
```
jp7677: We could also keep `m_supportsExtGraphicsCommandList1` and make `m_commandListExtTier` a local var in the constructor.

src/nvapi/nvapi_as_convert.h
```diff
@@ -0,0 +1,52 @@
...
+        }
+        NvapiAsConverter(const NvapiAsConverter&) = delete;
+        NvapiAsConverter& operator=(const NvapiAsConverter&) = delete;
+
+        [[nodiscard]] NvAPI_Status convert(D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS& d3d12Desc,
+            const NVAPI_D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS_EX& nvapiDesc,
+            bool supportsOmm);
+
+      private:
+        static const size_t StackReserve = 16;
```
jp7677: Could be `constexpr`

src/nvapi_d3d12.cpp
```diff
@@ -734,115 +737,313 @@ NVAPI_FUNCTION NvAPI_D3D12_GetRaytracingCaps(ID3D12Device* pDevice, NVAPI_D3D12_
...
+
+    constexpr NvU32 postbuildStackCapacity = 16;
+    D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_DESC postbuildStack[postbuildStackCapacity];
+    std::unique_ptr<D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_DESC[]> postbuildHeap;
+    D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_DESC* postbuildDescs = postbuildStack;
+    if (pParams->numPostbuildInfoDescs > postbuildStackCapacity) {
+        postbuildHeap = std::make_unique<D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_DESC[]>(pParams->numPostbuildInfoDescs);
+        postbuildDescs = postbuildHeap.get();
+    }
+    for (NvU32 i = 0; i < pParams->numPostbuildInfoDescs; ++i) {
```
jp7677: Could be `auto i = 0U`

src/nvapi_d3d12.cpp
```diff
@@ -734,115 +737,313 @@ NVAPI_FUNCTION NvAPI_D3D12_GetRaytracingCaps(ID3D12Device* pDevice, NVAPI_D3D12_
...
-    if (!ConvertBuildRaytracingAccelerationStructureInputs(&pParams->pDesc->inputs, geometryDescs, &desc.Inputs))
+    auto commandList = NvapiD3d12GraphicsCommandList::GetOrCreate(pCommandList);
+    if (!commandList)
+        return NoImplementation(n, alreadyLoggedNoImplementation);
+    if (!commandList->IsOpacityMicromapSupported())
+        return NotSupported(n);
+
+    if (!commandList->VerifyOpacityMicromapArrayNVAPI(pParams->opacityMicromapArray))
+        return Error(n, alreadyLoggedError);
+    return Ok(n, alreadyLoggedOk);
```
jp7677, line 1008: Empty line above please.

tests/nvapi_d3d11.cpp
```diff
@@ -59,6 +59,15 @@ TEST_CASE("D3D11 methods succeed", "[.d3d11]") {
         .LR_SIDE_EFFECT(*_1 = &device)
         .LR_SIDE_EFFECT(deviceRefCount++);
 
+    ALLOW_CALL(device, QueryInterface(__uuidof(IUnknown), _))
```
jp7677: For readability it might be better to place the `IUnknown` expectations first (so in this case before line 14).

tests/nvapi_d3d12.cpp
```diff
@@ -187,8 +257,12 @@ TEST_CASE("D3D12 methods succeed", "[.d3d12]") {
         // DXVK_NVAPI_D3D12_NV_SHADER_EXTN = 1 is set in section starting listener
 
         SECTION("IsNvShaderExtnOpCodeSupported without VKD3D-Proton support") {
+            ALLOW_CALL(device, QueryInterface(__uuidof(ID3D12DeviceExt5), _))
+                .RETURN(E_NOTIMPL);
             ALLOW_CALL(device, QueryInterface(__uuidof(ID3D12DeviceExt4), _))
                 .RETURN(E_NOTIMPL);
+            ALLOW_CALL(device, QueryInterface(__uuidof(ID3D12DeviceExt3), _))
```
jp7677: Looks like I missed to define

```
ALLOW_CALL(device, QueryInterface(__uuidof(ID3D12DeviceExt3), _))
        .LR_SIDE_EFFECT(*_2 = static_cast<ID3D12DeviceExt3*>(&device))
        .LR_SIDE_EFFECT(deviceRefCount++)
        .RETURN(S_OK);
```
 on the top level of those section. With that added, the expectation for `ID3D12DeviceExt3` here and below should not be needed.

jp7677 (commented):
Hi @chrisd-work thanks a lot for the PR. Very cool to see that things are coming together and falling into place. I did a review pass with focus om styling and consistency ( @Saancreed has a much better knowledge on the actual functionally ). Comments are mostly small things or suggestions.
Overall I think this looks very good and fits nicely. Also cool that you took care of the test suite!

jp7677:
Could you also please split the first commit (updating DX headers and introduction of `probeInterfaceChain`)?

src/nvapi/nvapi_as_convert.cpp
Saancreed: I'd suggest keeping `NvAPI_Status` type, but switching to just `NVAPI_OK` instead. Otherwise we'd have to remap it to `NvAPI_Status` anyway a moment later.

Saancreed (approved):
No further comments from my side, so I'll approve now, but I'll likely have another look once you adjust for feedback from Jens.

src/nvapi/nvapi_as_convert.cpp
jp7677: Yeah that's fine as well. We can later revisit the places where we use  `HRESULT` to see if it makes sense to make things consistent.

chrisd-work:
I will be working on the next pass today, I was on vacation for 5 days.

src/nvapi/nvapi_as_convert.cpp
chrisd-work: Yes, I agree lets keep it as `NvAPI_Status` for now. I avoided HRESULT for the reason that @Saancreed mentioned you would need a remap.

src/nvapi/nvapi_as_convert.cpp
chrisd-work: Fixed

src/nvapi/nvapi_as_convert.h
chrisd-work: Fixed

src/nvapi/nvapi_as_convert.h
chrisd-work: Fixed

src/nvapi/nvapi_d3d12_device.cpp
chrisd-work: Fixed

src/nvapi/nvapi_as_convert.cpp
chrisd-work: Fixed

src/nvapi/nvapi_as_convert.cpp
chrisd-work: Fixed

src/nvapi/nvapi_d3d12_device.h
chrisd-work: Fixed

src/nvapi/nvapi_d3d12_graphics_command_list.cpp
chrisd-work: Fixed

src/nvapi/nvapi_d3d12_graphics_command_list.h
chrisd-work: I used the name `m_supportsCubin64bit` to be consistent with the device. Otherwise changed as you suggested.

src/nvapi/nvapi_as_convert.h
chrisd-work: Fixed.

src/nvapi_d3d12.cpp
chrisd-work: Fixed.

src/nvapi_d3d12.cpp
chrisd-work: Fixed.

tests/nvapi_d3d11.cpp
chrisd-work: Fixed.

tests/nvapi_d3d12.cpp
chrisd-work: Fixed.

chrisd-work:
Header split as you asked. I think it got squashed on accident at some point during rebases. I have addressed all the issues you mentioned. Only one has a minor deviation from your suggestion which I noted.

src/nvapi/nvapi_as_convert.cpp
jp7677: What I meant (and I also think Sancreed wanted to suggest) was to return `NvAPI_Status` in all functions here directly, so `return NVAPI_OK` instead of `return Ok(n);` (and omit `constexpr auto n = __func__;`).

src/nvapi/nvapi_as_convert.cpp
jp7677: What about also putting `OmmArrayInputsToD3d12()` and `NvapiAsConverter::Convert()` into `namespace dxvk { }` (so that all of `nvapi_as_convert.cpp` is in it)?

src/nvapi_d3d12.cpp
```diff
@@ -734,62 +737,79 @@ NVAPI_FUNCTION NvAPI_D3D12_GetRaytracingCaps(ID3D12Device* pDevice, NVAPI_D3D12_
...
-    }
+    const bool supportsOmm = device->IsOpacityMicromapSupported();
 
-    return false;
+    D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS inputs;
+    /* Cannot cache the NvapiAsConverter per device, as device doesn't have to be thread safe, and NvapiAsConverter
+     * is destructive */
+    NvapiAsConverter asConverter;
+    if (auto status = asConverter.Convert(inputs, *pParams->pDesc, supportsOmm); status != NVAPI_OK)
+        return status;
```
jp7677, line 766: Looking again, when returning the status here directly, we do mis an ending log statement here. I guess that was the reason to use the status code functions in NvApiAsConverter to get the log statement from there?
Introducing HRESULT (or returning NVAPI_STATUS directly) and mappings here would resolve this (and also makes reducing log spam more consistent) but I get that the overhead isn’t that great. We could pass `n` into the function to “fix” the logging statements, but that feels messy. May be leave that for now and we address this later?

src/nvapi/nvapi_as_convert.cpp
chrisd-work: I will change this too my reluctance was born out of the fact that namespace indents the code as well. Typically I turn this of in clang-format. It just creates too much white space.

src/nvapi_d3d12.cpp
chrisd-work: I would want to keep it as is, at least with the current logging system. dxvk-nvapi could have an additional exit line, but to me it is more interesting to know that `NvapiAsConverter::Convert` failed than `NvAPI_D3D12_GetRaytracingAccelerationStructurePrebuildInfoEx`. In debugging I always want the minimal surface area. Logging at the function level expands the surface of code that can generate the error.

So the current implementation will reflect the function that fails as `NvapiAsConverter::Convert`, or one of its children. The entry point is less useful in general. If the entry point is important, I have a few ideas about how to do this. Mostly it is about using TLS with either `__func__` or `std::source_location` if C++20 is allowed. You can wrap this up with RAII.

Then when `log::info` or `log::trace` occurs, it can check the TLS; if it is enabled it can log two pieces of information: the NVAPI entry point, and the function + line + cpp in full logging that actually generated the error. If the TLS is missing, just the latter. This TLS approach would be more robust than passing `n` through call sites.

One counter proposal would be to log twice: once at the point of failure `NvapiAsConverter::Convert`, and a second time as the function leaves `NvAPI_D3D12_GetRaytracingAccelerationStructurePrebuildInfoEx`. However, this only works well in single-threaded applications. Since D3D12 applications are allowed to build command lists from multiple threads at the same time, a real possibility is that these lines will become interleaved. This makes the TLS solution the preferred one as it ensures that there is one line that contains all of the relevant information. The TLS solution does require a bit of discipline though: a programmer has to install the entry-point guard at the top of every NVAPI shim.

In the end this is your codebase. If you want me to change it I am happy to. As I stated in a PM, I plan to submit a later PR to cleanup the logging system, so taking a look at shoring up these things in that PR would be in scope.

chrisd-work: And I forgot to mention your idea of passing in the `n`. Yes that can work too at least as a stop gap, but the TLS would be more ideal long term.

src/nvapi/nvapi_as_convert.cpp
chrisd-work: Fixed.

chrisd-work:
Fixed one defect and responded to the other. Ready for another review pass.

jp7677 (approved):
LGTM, thanks for the fast response and let’s indeed defer the logging to another moment.

Saancreed (approved):

PR #375, opened 2026-06-25
Update to R610 headers

Saancreed:
Curiously enough, https://github.com/NVIDIA-RTX/Streamline/blob/e8aaa6eaac968711fb62473d4ae8256dde20919b/source/core/sl.ota/ota.h#L72-L74 still aren't in latest headers. Maybe we should manually add support?


jp7677 (approved):
LGTM, thanks! With regard to the Streamline OTA / update keys, I'm fine if we add those manually.

jp7677:
Note to self: update  https://github.com/jp7677/dxvk-nvapi/wiki/Passing-driver-settings once this is merged

Saancreed:
> With regard to the Streamline OTA / update keys, I'm fine if we add those manually.

Cool, I'll try to get this in shape soon, but in another PR. For now, the generic `DXVK_NVAPI_DRS_SETTINGS=0x10e41e06=1` should do.

PR #376, opened 2026-06-28
Add latency marker-only logging

jpietek:
## Summary

Add `DXVK_NVAPI_LATENCY_MARKER_LOG=1` to emit only latency marker events without enabling full `DXVK_NVAPI_LOG_LEVEL=trace`.

The marker-only log covers D3D, D3D12 async frame markers, and Vulkan latency markers. Each line includes the API, frame ID, marker type, marker value, and a QPC timestamp in microseconds.

Default behavior is unchanged.

## Rationale

Full trace logging records every NVAPI entry point and is too noisy when tooling only needs latency marker timing. This gives external tooling a smaller opt-in signal without parsing the full trace stream.

This metric is very useful for building a latency meter on Linux similar to the NVIDIA App latency meter. It could be used by applications with in-game overlays, such as MangoHUD. I am using this metric in my NVIDIA GPU tuning tool, PenguinBurner: https://github.com/jpietek/PenguinBurner

## Validation

- `clang-format`
- `git diff --check HEAD~1..HEAD`
- `meson setup --cross-file ./build-win64.txt --buildtype debugoptimized -Denable_tests=True build/64`
- `meson compile -C build/64`
- `wine build/64/tests/nvapi64-tests.exe '[@unit-tests]'`\n- `wine build/64/tests/nvofapi64-tests.exe '[@unit-tests]'`\n- `meson setup --cross-file ./build-win32.txt --buildtype debugoptimized -Denable_tests=True build/32`\n- `meson compile -C build/32`\n- `python3 validate-methods.py ...`


jp7677:
Thanks for the PR. I understand your motivation, but the DXVK-NVAPI logging mechanism is not the best place for that. The logging is really just meant for troubleshooting, not for IPC. Additionally we don't to commit to a fixed logging format.

You might want to look into Vulkan layers to extract the info you need.

PR #377, opened 2026-07-01
nvapi-vulkan: Refactor VkReflex implementation into separate backends

Saancreed:
Not sure if the extra level of indirection via `unique_ptr` is worth it, but I like having the backends separated.

(Let's not merge this just yet.)

### DIFF
--- a/src/nvapi/nvapi_vulkan_low_latency_device.cpp
+++ b/src/nvapi/nvapi_vulkan_low_latency_device.cpp
@@ -1,4 +1,6 @@
 #include "./nvapi_vulkan_low_latency_device.h"
+#include "../util/util_env.h"
+#include "../util/util_log.h"
 
 namespace dxvk {
     std::unique_ptr<Vk> NvapiVulkanLowLatencyDevice::m_vk = nullptr;
@@ -28,30 +30,24 @@ namespace dxvk {
         m_vk.reset();
     }
 
-    std::pair<NvapiVulkanLowLatencyDevice*, VkResult> NvapiVulkanLowLatencyDevice::GetOrCreate(VkDevice device, bool allowFake) {
+    std::pair<NvapiVulkanLowLatencyDevice*, VkResult> NvapiVulkanLowLatencyDevice::GetOrCreate(VkDevice device) {
         std::scoped_lock lock{m_mutex};
 
         if (auto lowLatencyDevice = Get(device))
             return {lowLatencyDevice, VK_SUCCESS};
 
-        if (!m_vk || !m_vk->IsAvailable())
+        if (!m_vk || !m_vk->IsAvailable()) {
+            log::info("Initializing Vulkan Low-Latency failed: could not find usable Vulkan loader (or winevulkan) module in the current process");
             return {nullptr, VK_ERROR_INITIALIZATION_FAILED};
-
-#define VK_GET_DEVICE_PROC_ADDR(proc) auto proc = reinterpret_cast<PFN_##proc>(m_vk->GetDeviceProcAddr(device, #proc))
-
-        VK_GET_DEVICE_PROC_ADDR(vkCreateSemaphore);
-        VK_GET_DEVICE_PROC_ADDR(vkDestroySemaphore);
-
-        if (!vkCreateSemaphore || !vkDestroySemaphore)
-            return {nullptr, VK_ERROR_INCOMPATIBLE_DRIVER};
+        }
 
         std::unique_ptr<NvapiVulkanLowLatencyDevice> lowLatencyDevice;
         VkResult vr;
 
-        std::tie(lowLatencyDevice, vr) = NvapiVulkanLowLatency2LayerDevice::TryCreate(device, vkDestroySemaphore, vkCreateSemaphore);
+        std::tie(lowLatencyDevice, vr) = NvapiVulkanLowLatency2LayerDevice::TryCreate(device);
 
-        if ((!lowLatencyDevice || vr != VK_SUCCESS) && allowFake)
-            std::tie(lowLatencyDevice, vr) = NvapiVulkanLowLatencyFakeDevice::TryCreate(device, vkDestroySemaphore, vkCreateSemaphore);
+        if (!lowLatencyDevice || vr != VK_SUCCESS)
+            std::tie(lowLatencyDevice, vr) = NvapiVulkanLowLatencyFakeDevice::TryCreate(device);
 
         if (!lowLatencyDevice || vr != VK_SUCCESS)
             return {nullptr, vr};
@@ -124,6 +120,8 @@ namespace dxvk {
         m_vkDestroySemaphore(m_device, m_semaphore, nullptr);
     }
 
+#define VK_GET_DEVICE_PROC_ADDR(proc) auto proc = reinterpret_cast<PFN_##proc>(m_vk->GetDeviceProcAddr(device, #proc))
+
     NvapiVulkanLowLatencyFakeDevice::NvapiVulkanLowLatencyFakeDevice(
         VkDevice device,
         VkSemaphore semaphore,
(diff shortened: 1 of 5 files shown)


src/nvapi_vulkan.cpp
```diff
@@ -57,7 +62,7 @@ NVAPI_FUNCTION NvAPI_Vulkan_InitLowLatencyDevice(HANDLE vkDevice, HANDLE* signal
         return NoImplementation(n);
     }
```
jp7677: May be it works to move all this to `NvapiVulkanLowLatencyDevice::GetOrCreate()` and decide there if we even have to create the NoopBackend when the LL2 backend could not be created. 
I think that would also make `GetBackend()` obsolete.

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -13,51 +14,95 @@ namespace dxvk {
...
+        VkSemaphore m_semaphore{};
+        PFN_vkDestroySemaphore m_vkDestroySemaphore{};
+
+      private:
+        static std::unique_ptr<Vk> m_vk;
+        static std::unordered_map<VkDevice, std::unique_ptr<NvapiVulkanLowLatencyDevice>> m_nvapiDeviceMap;
+        static std::mutex m_mutex;
+    };
+
+    class NvapiVulkanLowLatencyNoopDevice : public NvapiVulkanLowLatencyDevice {
```
jp7677: May be `NvapiVulkanLowLatencyFakeVkReflexDevice` or `NvapiVulkanLowLatencyFakeDevice`? This also reflects the name of the environment variable.

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -13,51 +14,95 @@ namespace dxvk {
...
+      public:
+#define PFN_PARAM(proc) PFN_##proc proc
+        [[nodiscard]] explicit NvapiVulkanLowLatencyNoopDevice(
+            VkDevice device,
+            VkSemaphore semaphore,
+            PFN_PARAM(vkDestroySemaphore),
+            PFN_PARAM(vkSignalSemaphore));
+#undef PFN_PARAM
+
+        [[nodiscard]] virtual VulkanLowLatencyBackend GetBackend() const override;
```
jp7677: `virtual` should not be needed in override functions.

src/nvapi/nvapi_vulkan_low_latency_device.h
jp7677: Also `final`?

src/nvapi_private.h
```diff
@@ -69,4 +70,9 @@
 // clang-format on
 
 #define NVAPI_QUERY_INTERFACE extern "C" void* __cdecl
-#define NVAPI_FUNCTION extern "C" NvAPI_Status __cdecl
\ No newline at end of file
+#define NVAPI_FUNCTION extern "C" NvAPI_Status __cdecl
+
+enum VulkanLowLatencyBackend {
```
jp7677: I would add this enum to `nvapi_vulkan_low_latency_device.h` (or `nvapi_vulkan_low_latency_device.cpp` if this becomes needed only withing `nvapi_vulkan_low_latency_device.cpp`)

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -13,51 +14,95 @@ namespace dxvk {
...
+        [[nodiscard]] virtual VkResult Sleep(uint64_t value) override;
+        virtual void GetLatencyTimings(std::span<NV_VULKAN_LATENCY_RESULT_PARAMS_V1::vkFrameReport, 64> frameReports) override;
+        [[nodiscard]] virtual bool SetLatencyMarker(uint64_t frameID, NV_VULKAN_LATENCY_MARKER_TYPE marker) override;
+        virtual void QueueNotifyOutOfBand(VkQueue queue, NV_VULKAN_OUT_OF_BAND_QUEUE_TYPE queueType) override;
+
+      private:
+        PFN_vkSignalSemaphore m_vkSignalSemaphore{};
+    };
+
+    class NvapiVulkanLowLatency2Device : public NvapiVulkanLowLatencyDevice {
```
jp7677: Some other ideas for the name: `NvapiVulkanLowLatencyVkNvLl2Device` or even `NvapiVulkanLowLatencyVkReflexLayerDevice` or `NvapiVulkanLowLatencyLayerDevice`

src/util/util_log.h
```diff
@@ -124,6 +124,19 @@ namespace dxvk::log {
         inline std::string ngx_dlss_override_set_state_params(NV_NGX_DLSS_OVERRIDE_SET_STATE_PARAMS* p) {
             return str::format("{version=", p->version, ",processIdentifier=", p->processIdentifier, ",feature=", p->feature, ",feedbackMask=0x", std::hex, p->feedbackMask, "}");
         }
+
+        inline std::string_view vk_ll_backend(VulkanLowLatencyBackend backend) {
```
jp7677: An alternative would be to add a `name` function to the backends.

src/nvapi/nvapi_vulkan_low_latency_device.cpp
```diff
@@ -123,83 +134,99 @@ namespace dxvk {
         if (node.empty())
             return false;
 
-        auto& lowLatencyDevice = node.mapped();
-
-        lowLatencyDevice.m_vkDestroySemaphore(lowLatencyDevice.m_device, lowLatencyDevice.m_semaphore, nullptr);
+        auto lowLatencyDevice = std::move(node.mapped());
+        // the destructor will take care of semaphore cleanup
```
jp7677, line 80: somewhat indirect, but also elegant ;)

src/nvapi/nvapi_vulkan_low_latency_device.cpp
```diff
@@ -123,83 +134,99 @@ namespace dxvk {
...
+        PFN_PARAM(vkSignalSemaphore))
+        : NvapiVulkanLowLatencyDevice(device, semaphore, vkDestroySemaphore),
+          // Without the layer we'll pretend that Reflex is happening so that apps don't get a pink tint.
+          PFN_INIT(vkSignalSemaphore) {}
+
+    VulkanLowLatencyBackend NvapiVulkanLowLatencyNoopDevice::GetBackend() const {
+        return VulkanLowLatencyBackend::Noop;
+    }
+
+    NvBool NvapiVulkanLowLatencyNoopDevice::GetLowLatencyMode() const {
```
jp7677: I would be OK to have the implementation of those simple getters directly in the header file

```c++
[[nodiscard]] NvBool GetLowLatencyMode() const override { return NV_FALSE; }
```

src/nvapi/nvapi_vulkan_low_latency_device.cpp
```diff
@@ -44,32 +41,32 @@ namespace dxvk {
...
-        if (!vkSetLatencySleepModeNV || !vkLatencySleepNV || !vkGetLatencyTimingsNV || !vkSetLatencyMarkerNV || !vkQueueNotifyOutOfBandNV) {
-            // VK_NV_low_latency2 was not requested -> our Vulkan Reflex layer is not present -> grab vkSignalSemaphore to fake that reflex is working
-            //
-            // The app should have requested either the Vulkan 1.2 or the VK_KHR_timeline_semaphore extension
-            // We'll query whichever is available
-            if (!vkSignalSemaphore) {
-                VK_GET_DEVICE_PROC_ADDR(vkSignalSemaphoreKHR);
-                vkSignalSemaphore = vkSignalSemaphoreKHR;
-            }
+        VulkanLowLatencyBackend backend;
```
jp7677: May be `implementation` is somewhat more fitting.

jp7677 (commented):
Finally did a review pass with some thoughts and suggestions. Looks cool and I think we can even move more things into `NvapiVulkanLowLatencyDevice`

src/nvapi/nvapi_vulkan_low_latency_device.cpp
```diff
@@ -44,32 +41,32 @@ namespace dxvk {
...
 
-            if (!vkSignalSemaphore)
-                return {nullptr, VK_ERROR_EXTENSION_NOT_PRESENT};
+        if (vkSetLatencySleepModeNV && vkLatencySleepNV && vkGetLatencyTimingsNV && vkSetLatencyMarkerNV && vkQueueNotifyOutOfBandNV) {
```
jp7677: Something like a `TryCreate()` within each backend/implementation could also work, that way the probing of the VK function requirements could move into the corresponding backends/implementations.

src/nvapi_private.h
Saancreed: It's here because I needed it for the utility formatting thingy. But maybe it's not needed after all, we'll see.

src/nvapi/nvapi_vulkan_low_latency_device.cpp
Saancreed: I'll probably do this for `GetBackend` / `GetImplementation`, but I'm not convinced about other methods just yet.

jp7677: It’s just a thought, no preference from my side.

src/nvapi_vulkan.cpp
Saancreed: Do we want to have any kind of log statements in `NvapiVulkanLowLatencyDevice` though? This lets me avoid them.

jp7677: I think it's fine to also move the log statements. We already have a few setup related log statements in e.g. `NvapiAdapter::Initialize()`.

src/nvapi/nvapi_vulkan_low_latency_device.cpp
Saancreed: "Backend" made sense to me the same way Ninja or Make are backends for Meson. I went with the rename, but I don't like "implementation" any more than "backend"; I've made too many typos in "implementation" in my career already :sweat_smile:

PR #378, opened 2026-07-02
nvapi-d3d12: Allows NVK in GetGraphicsCapabilities

marysaka:
Saw this while doing some tracing with DLSS on NVK, might not matter much but lets be consistent here.


Saancreed:
Sounds good, but one of the tests will have to be updated to account for this. Check the Windows CI output to see which one.

marysaka:
Should be all good now

Saancreed (approved):

PR #379, opened 2026-07-03
Improve logging and some smaller things

jp7677:
Some stuff across the board (sorry for being lazy for not splitting this into multiple PR's). Up for discussion.

- First few commits are trivial.
- Log GetRaytracingCaps: Usefull to get an idea if requirements are met
- Report SHFL support for RTX Path Tracing Sample: I like this sample (version 1.7 for now) for short validation. It could be useful to support it without much configuration.
- Prevent OMM log spam: Probably slightly controversial but up for discussion. I went for `HRESULT` for now just for being consistent. I don't mind changing this back when at some point looking more closely to our logging mechanism.  Not having a good HRESULT for not-supported might be a motivation though to keep NvAPI_Status in NvapiAsConverter, but just move the status code logging out of it.


src/util/util_pso_extension.h
```diff
@@ -5,15 +5,15 @@
...
-            {5, "NV_PSO_SET_SHADER_EXTENSION_SLOT_AND_SPACE"},
-            {6, "NV_PSO_VERTEX_SHADER_EXTENSION"},
-            {7, "NV_PSO_DOMAIN_SHADER_EXTENSION"},
-            {9, "NV_PSO_HULL_SHADER_EXTENSION"},
+            MAP_ENUM_VALUE(NV_PSO_RASTER_EXTENSION),
+            MAP_ENUM_VALUE(NV_PSO_REQUEST_FASTGS_EXTENSION),
+            MAP_ENUM_VALUE(NV_PSO_GEOMETRY_SHADER_EXTENSION),
+            MAP_ENUM_VALUE(NV_PSO_ENABLE_DEPTH_BOUND_TEST_EXTENSION),
+            MAP_ENUM_VALUE(NV_PSO_EXPLICIT_FASTGS_EXTENSION),
+            MAP_ENUM_VALUE(NV_PSO_SET_SHADER_EXTNENSION_SLOT_AND_SPACE),
```
Saancreed: Consider using the non-typo'ed name 😄

src/nvapi_d3d12.cpp
```diff
@@ -29,7 +29,16 @@ NVAPI_FUNCTION NvAPI_D3D12_IsNvShaderExtnOpCodeSupported(ID3D12Device* pDevice,
     if (!device)
         return NoImplementation(n, alreadyLoggedNoImplementation);
 
-    *pSupported = env::isD3d12NvShaderExtnEnabled() && device->IsNvShaderExtnOpCodeSupported(opCode);
+    if (env::isNvidiaRtxPathTracingSample()) {
+        if (opCode == NV_EXTN_OP_SHFL) {
+            // NVIDIA RTX Path Tracing queries NV_EXTN_OP_SHFL for detecting HLSL extension UAV support
+            log::info("Reporting NV_EXTN_OP_SHFL HLSL support for NVIDIA RTX Path Tracing Sample");
+            *pSupported = true;
```
Saancreed: This way we won't check if vkd3d-proton side supports related `ID3D12DeviceExt` at all. Maybe we should call it with `OP_REORDER_THREAD` as a workaround instead?

Saancreed (commented):
Two more things I've noticed recently that we can take care of while we're at it:

* Typo in https://github.com/jp7677/dxvk-nvapi/blob/e4acf4a297a7b3cc80cc1ebc47380d589fdc0cb6/src/nvapi_d3d12.cpp#L870
* https://github.com/jp7677/dxvk-nvapi/blob/e4acf4a297a7b3cc80cc1ebc47380d589fdc0cb6/src/nvapi/nvapi_d3d12_graphics_command_list.h#L32 should be named something like `m_supportsLaunchCubinShaderEx`; it's actually unrelated to 64-bit cubin support.

src/util/util_pso_extension.h
jp7677: ah nice, didn't even see that one. The beauty of vertical select in idea :)

src/nvapi_d3d12.cpp
jp7677: yeah, that's a better alternative.

jp7677:
* https://github.com/jp7677/dxvk-nvapi/blob/e4acf4a297a7b3cc80cc1ebc47380d589fdc0cb6/src/nvapi/nvapi_d3d12_graphics_command_list.h#L32
> should be named something like `m_supportsLaunchCubinShaderEx`; it's actually unrelated to 64-bit cubin support.

What about `m_supportsCubinSMem`, based on https://github.com/HansKristian-Work/vkd3d-proton/pull/1640

Saancreed:
That does sound better, yeah.

src/util/util_env.cpp
```diff
@@ -181,10 +181,24 @@ namespace dxvk::env {
         return false;
     }
 
+    bool isNvidiaRtxPathTracingSample() {
+        static bool alreadyLogged = false;
+        if (getExecutableName() == std::string("Rtxpt.exe")) {
+            if (!std::exchange(alreadyLogged, true))
+                log::info(str::format("NVIDIA RTX Path Tracing Sample (", getExecutableName(), ") detected"));
```
Saancreed:
```suggestion
                log::info(str::format("NVIDIA RTX Path Tracing Sample (Rtxpt.exe) detected"));
```

We already know the name, and this saves us some unnecessary allocations.

src/nvapi/nvapi_as_convert.cpp
```diff
@@ -148,22 +143,21 @@ namespace dxvk {
         d3d12Inputs.DescsLayout = D3D12_ELEMENTS_LAYOUT_ARRAY;
         d3d12Inputs.pOpacityMicromapArrayDesc = &arrayDesc;
 
-        if (NvAPI_Status status = OmmArrayBuildFlagsToD3d12(d3d12Inputs.Flags, nvapiInputs.flags); status != NVAPI_OK)
+        if (HRESULT status = OmmArrayBuildFlagsToD3d12(d3d12Inputs.Flags, nvapiInputs.flags); status != S_OK)
```
Saancreed: Do we want to use `if (auto status = …` in such cases or not?

src/nvapi_d3d12.cpp
```diff
@@ -867,13 +897,13 @@ NVAPI_FUNCTION NvAPI_D3D12_SetCreatePipelineStateOptions(ID3D12Device5* pDevice,
...
-    static const NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS supprotedNvapiFlags =
+    static constexpr NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS supportedNvapiFlags =
         NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS_ENABLE_OMM_SUPPORT;
-    if (pState->flags & ~supprotedNvapiFlags)
+    if (pState->flags & ~supportedNvapiFlags)
         return NotSupported(n);
 
     if (pState->flags & NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS_ENABLE_OMM_SUPPORT)
-        flags = (D3D12_VK_EXT_PIPELINE_CREATION_STATE_FLAG)(flags | D3D12_VK_EXT_PIPELINE_CREATION_STATE_FLAGS_ENABLE_OMM_SUPPORT);
+        flags = static_cast<D3D12_VK_EXT_PIPELINE_CREATION_STATE_FLAG>(flags | D3D12_VK_EXT_PIPELINE_CREATION_STATE_FLAGS_ENABLE_OMM_SUPPORT);
```
Saancreed, line 905:
```suggestion
        flags = static_cast<decltype(flags)>(flags | D3D12_VK_EXT_PIPELINE_CREATION_STATE_FLAGS_ENABLE_OMM_SUPPORT);
```

The fact that the cast is required at all is stupid imho, but feel free to keep the current version if you prefer it more explicit.

src/nvapi_d3d12.cpp
```diff
@@ -867,13 +897,13 @@ NVAPI_FUNCTION NvAPI_D3D12_SetCreatePipelineStateOptions(ID3D12Device5* pDevice,
         return NotSupported(n);
 
     D3D12_VK_EXT_PIPELINE_CREATION_STATE_FLAG flags = D3D12_VK_EXT_PIPELINE_CREATION_STATE_FLAGS_NONE;
-    static const NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS supprotedNvapiFlags =
+    static constexpr NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS supportedNvapiFlags =
```
Saancreed: Can be `auto` if we want it, in both lines I think?

Saancreed (commented):
> Prevent OMM log spam: Probably slightly controversial but up for discussion. I went for HRESULT for now just for being consistent. I don't mind changing this back when at some point looking more closely to our logging mechanism. Not having a good HRESULT for not-supported might be a motivation though to keep NvAPI_Status in NvapiAsConverter, but just move the status code logging out of it.

Now that I think about it, I'd say I prefer having that NOT_SUPPORTED as an option. Can we change the type back, but leave the logging in the new place? I don't really find the inconsistency of `HRESULT` vs `NvAPI_Status` that big of a problem. VkReflex has `VkResult` in a few places too, so it's already pretty inconsistent… just a matter of using the best type for the job at hand.

src/nvapi/nvapi_as_convert.cpp
jp7677: ah yeah, removed that one here.

src/nvapi_d3d12.cpp
jp7677: I would prefer the explicit version.

src/nvapi/nvapi_as_convert.cpp
Saancreed: Ah, direct return is even better if we can afford it, but there are a few other cases of `if (NvAPI_Status status = …` in this file.

Saancreed (approved):
👍, but see https://github.com/jp7677/dxvk-nvapi/pull/379#discussion_r3521045100.

jp7677:
> 👍, but see [#379 (comment)](https://github.com/jp7677/dxvk-nvapi/pull/379#discussion_r3521045100).

Sry, had to hurry a bit. Should be good now. 

> Now that I think about it, I'd say I prefer having that NOT_SUPPORTED as an option. Can we change the type back, 
> but leave the logging in the new place? I don't really find the inconsistency of HRESULT vs NvAPI_Status that big of a > problem. VkReflex has VkResult in a few places too, so it's already pretty inconsistent… just a matter of using the 
> best type for the job at hand.

Yeah agree, having seen both versions now, HRESULT vs NvAPI_Status doesn't matter but NOT_SUPPORTED does look better. That said, the comment in the headers reads `Requested feature is not supported in the selected GPU`, which is not exactly how we are using it, but well ... ;)

PR #380, opened 2026-07-04
util: Logging improvements

jp7677:
This might avoid a few cycles. No big overhaul, just (hopefully) some small improvements.

### DIFF
--- a/src/nvapi.cpp
+++ b/src/nvapi.cpp
@@ -14,7 +14,7 @@ static auto initializationMutex = std::mutex{};
 using namespace dxvk;
 
 NVAPI_FUNCTION NvAPI_EnumLogicalGPUs(NvLogicalGpuHandle nvGPUHandle[NVAPI_MAX_LOGICAL_GPUS], NvU32* pGpuCount) {
-    static constexpr auto n = __func__;
+    static constexpr auto n = FUNC;
 
     if (log::tracing())
         log::trace(n, log::fmt::ptr(nvGPUHandle), log::fmt::ptr(pGpuCount));
@@ -34,7 +34,7 @@ NVAPI_FUNCTION NvAPI_EnumLogicalGPUs(NvLogicalGpuHandle nvGPUHandle[NVAPI_MAX_LO
 }
 
 NVAPI_FUNCTION NvAPI_EnumPhysicalGPUs(NvPhysicalGpuHandle nvGPUHandle[NVAPI_MAX_PHYSICAL_GPUS], NvU32* pGpuCount) {
-    static constexpr auto n = __func__;
+    static constexpr auto n = FUNC;
 
     if (log::tracing())
         log::trace(n, log::fmt::ptr(nvGPUHandle), log::fmt::ptr(pGpuCount));
@@ -54,7 +54,7 @@ NVAPI_FUNCTION NvAPI_EnumPhysicalGPUs(NvPhysicalGpuHandle nvGPUHandle[NVAPI_MAX_
 }
 
 NVAPI_FUNCTION NvAPI_EnumTCCPhysicalGPUs(NvPhysicalGpuHandle nvGPUHandle[NVAPI_MAX_PHYSICAL_GPUS], NvU32* pGpuCount) {
-    static constexpr auto n = __func__;
+    static constexpr auto n = FUNC;
 
     if (log::tracing())
         log::trace(n, log::fmt::ptr(nvGPUHandle), log::fmt::ptr(pGpuCount));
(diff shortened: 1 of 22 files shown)


src/nvofapi.cpp
```diff
@@ -41,11 +41,14 @@ NVOFAPI_FUNCTION OFSessionInit(NvOFHandle hOf, const NV_OF_INIT_PARAMS* initPara
         return ErrorGeneric(n);
     }
 
-    return nvOF->InitSession(initParams);
+    if (nvOF->InitSession(initParams) != NV_OF_SUCCESS)
+        return ErrorGeneric(n);
+
+    return Success(n);
```
Saancreed, line 47: Right, we don't really use more specific error codes anyway…

src/nvapi.cpp
```diff
@@ -14,7 +14,7 @@ static auto initializationMutex = std::mutex{};
 using namespace dxvk;
 
 NVAPI_FUNCTION NvAPI_EnumLogicalGPUs(NvLogicalGpuHandle nvGPUHandle[NVAPI_MAX_LOGICAL_GPUS], NvU32* pGpuCount) {
-    constexpr auto n = __func__;
+    static constexpr auto n = __func__;
```
Saancreed: One potential problem I still see with this is being typed as `const char*` string literal, then implicitly converted to `std:string_view` every time something like `Ok(n)` is called. Are compilers smart enough to figure out that this is a string literal and immediately know its length, or is this going to perform `strlen` every time?

If the latter. maybe something like `static constexpr std::string_view n = std::string_view{__func__};` could improve this?

src/util/util_log.h
```diff
@@ -145,14 +151,14 @@ namespace dxvk::log {
...
+        log::write(TRACE, name);
     }
 
     template <typename... Args>
-    void trace(const std::string& name, const Args&... args) {
+    void trace(const std::string_view name, const Args&... args) {
         std::stringstream stream;
         append(stream, args...);
-        log::write("trace", str::format(name, " (", stream.str(), ")"));
+        log::write(TRACE, str::format(name, " (", stream.str(), ")"));
```
Saancreed, line 162: There would be even more to gain by deferring this until we know that trace log level is enabled, but it might be kinda awkward to achieve.

src/util/util_log.cpp
```diff
@@ -5,34 +5,36 @@
...
     static std::mutex fileStreamMutex;
 
-    void print(const std::string& logMessage) {
-        auto line = logMessage + '\n'; // Do not flush buffers
+    void print(const std::string_view logMessage) {
+        auto line = str::format(logMessage, '\n'); // Do not flush buffers
         if (wineDbgOutput)
             wineDbgOutput(line.c_str());
         else
             std::cerr << line;
```
Saancreed, line 27: We could be a bit smarter here and do something like

```suggestion
        if (wineDbgOutput) {
            char buffer[1024]; // arbitrary length
            auto i = logMessage.copy(buffer, sizeof(buffer) - 2);
            buffer[i] = '\n';
            buffer[i + i] = 0;
            wineDbgOutput(buffer);
        } else {
            std::cerr << logMessage << '\n';
        }
```

src/nvapi.cpp
jp7677: To be on the safe side, I've added exactly this, packaged now in our own `FUNC`  marco.

src/util/util_log.h
jp7677: I don't really mind the penalty when tracing, so I would prefer to leave this for now.

src/util/util_log.cpp
jp7677: OK to keep the simple `str::format` version? I wanted to primarily avoid extra cycles when no logging or with logging on info. With info we tried our best to avoid repeatable log statements, so this should be only become relevant when tracing. In that case I don't really mind the penalty.

Saancreed: This would still save us allocating a copy of the message on every line of output, but sure, we can revisit this later if needed.

src/util/util_log.h
Saancreed: Okay, but keep in mind that this isn't a problem only when tracing. There are tens of `log::info(str::format(…` in our code, mostly in `util_statuscode.h`, that will allocate unconditionally, but wouldn't have to if we had some kind of `log::info_format(…)` instead.

Admittably, my comment wasn't placed in the most fortunate spot for this remark, but the issue is wider than that.

Saancreed (approved):
Oops, forgot to approve. But see latest comments for some future work.

src/util/util_log.h
jp7677: I see those and agree to handle this at some point with a bigger rework. For now the `alreadyLogged*` mechanics should avoid hitting those on hot paths since, as far as I'm aware, all those currently end with `Ok(n, alreadyLoggedOk)`. But yeah. surely to be continued.

PR #381, opened 2026-07-09
nvapi: Fix an adapter registry initialization race.

shaunren:
`NvAPI_Initialize()` currently allocates and assigns `nvapiAdapterRegistry` before the registry had finished initializing, allowing other NVAPI entrypoints to observe an uninitialized non-null adapter registry.

This fixes a crash in Elite Dangerous when using the EDHM mod, where `NvAPI_SYS_GetDriverAndBranchVersion()` races with `NvAPI_Initialize()` and calls `nvapiAdapterRegistry->GetFirstAdapter()` before the registry is initialized.

### DIFF
--- a/src/nvapi.cpp
+++ b/src/nvapi.cpp
@@ -375,12 +375,12 @@ NVAPI_FUNCTION NvAPI_Initialize() {
     if (!resourceFactory)
         resourceFactory = std::make_unique<NvapiResourceFactory>();
 
-    nvapiAdapterRegistry = std::make_unique<NvapiAdapterRegistry>(*resourceFactory);
-    if (!nvapiAdapterRegistry->Initialize()) {
-        nvapiAdapterRegistry.reset();
+    auto registry = std::make_unique<NvapiAdapterRegistry>(*resourceFactory);
+    if (!registry->Initialize()) {
         --initializationCount;
         return NvidiaDeviceNotFound(n);
     }
+    nvapiAdapterRegistry = std::move(registry);
 
 #if _WIN64
     SetNgxDebugOptions(); // NGX is 64-bit only


jp7677 (approved):
Thanks, makes sense and elegant solution.

Saancreed (approved):
Curious, does NvAPI_Initialize have any unwritten thread safety guarantees? Application calling any other NvAPI_ function before Initialize returns sounds like a bug…

jp7677:
> Curious, does NvAPI_Initialize have any unwritten thread safety guarantees? Application calling any other NvAPI_ function before Initialize returns sounds like a bug…

Thinking about this further, may be the same logic should also be applied to `NvAPI_Unload`. I remember this being mentioned at some point. Something like this (which is ofc not a solution for avoiding unload while another function is being called at the same time, but merely changes timing):

```diff
diff --git a/src/nvapi.cpp b/src/nvapi.cpp
index a5f79d3..8a4519c 100644
--- a/src/nvapi.cpp
+++ b/src/nvapi.cpp
@@ -346,8 +346,10 @@ NVAPI_FUNCTION NvAPI_Unload() {
     if (initializationCount == 0)
         return ApiNotInitialized(n);
 
-    if (--initializationCount == 0)
-        nvapiAdapterRegistry.reset();
+    if (--initializationCount == 0) {
+        auto registry = std::move(nvapiAdapterRegistry);
+        registry.reset();
+    }
 
     return Ok(n);
 }
```

(The `reset` not being strictly needed)

(@shaunren I hope you are ok with slightly widening the scope of the discussion of your PR.)

jp7677:
From what I read, `reset()` does the replacement (with `nullptr`) first and then deletes the old instance, so first doing `move` wouldn't make any difference.

Anyway, thanks for the PR, merged!

PR #383, opened 2026-07-12
nvapi-vulkan: Refactor VkReflex implementation into separate backends (2)

jp7677:
Based on #377 . This one avoids virtual functions.


src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -4,60 +4,106 @@
...
             PFN_PARAM(vkLatencySleepNV),
             PFN_PARAM(vkGetLatencyTimingsNV),
             PFN_PARAM(vkSetLatencyMarkerNV),
-            PFN_PARAM(vkQueueNotifyOutOfBandNV),
-            PFN_PARAM(vkSignalSemaphore));
+            PFN_PARAM(vkQueueNotifyOutOfBandNV));
 #undef PFN_PARAM
 
-        [[nodiscard]] bool IsLayerPresent() const;
         [[nodiscard]] VkSemaphore GetSemaphore() const;
```
Saancreed: `GetSemaphore` can be moved to base class, and even have the implementation inlined in the header file.

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -4,60 +4,106 @@
 #include "./nvapi_resource_factory.h"
 
 namespace dxvk {
+
+    enum LowLatencyDeviceType : uint8_t {
```
Saancreed: I think preferred even the `Implementation` in the name over `Type`, the latter is just the most generic word possible 🙈

Can we make it scoped (`enum class`)? Shouldn't we specify in the enum's name that it's _Vulkan_ Reflex?

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -4,60 +4,106 @@
...
+        static std::mutex m_mutex;
+    };
+
+    class NvapiVulkanLowLatency2LayerDevice final : public NvapiVulkanLowLatencyDevice {
+      public:
+        [[nodiscard]] static std::pair<std::unique_ptr<NvapiVulkanLowLatency2LayerDevice>, VkResult> TryCreate(Vk* vk, VkDevice device);
 
 #define PFN_PARAM(proc) PFN_##proc proc
-        [[nodiscard]] explicit NvapiVulkanLowLatencyDevice(
+        [[nodiscard]] NvapiVulkanLowLatency2LayerDevice(
```
Saancreed: No longer `explicit`?

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -4,60 +4,106 @@
...
+#undef PFN_PARAM
+        [[nodiscard]] LowLatencyDeviceType GetType() const { return m_type; }
+        virtual ~NvapiVulkanLowLatencyDevice() { m_vkDestroySemaphore(m_device, m_semaphore, nullptr); }
+
+      protected:
+        VkDevice m_device{};
+        VkSemaphore m_semaphore{};
+
+      private:
+        LowLatencyDeviceType m_type;
```
Saancreed:
```suggestion
        LowLatencyDeviceType m_type{};
```

Although the constructor always takes care of it, so…

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -4,60 +4,106 @@
...
+
+    enum LowLatencyDeviceType : uint8_t {
+        LowLatency2,
+        VkReflexFake
+    };
+
     class NvapiVulkanLowLatencyDevice {
       public:
+#define PFN_PARAM(proc) PFN_##proc proc
+        explicit NvapiVulkanLowLatencyDevice(
```
Saancreed: Should we also delete copy and move constructors?

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -4,60 +4,106 @@
...
+      protected:
+        VkDevice m_device{};
+        VkSemaphore m_semaphore{};
+
+      private:
+        LowLatencyDeviceType m_type;
+        PFN_vkDestroySemaphore m_vkDestroySemaphore{};
+    };
+
+    class NvapiVulkanLowLatencyDeviceFactory {
```
Saancreed: Perhaps the factory could be moved into it's own pair of header+implementation files? Though I'm not sure how much it would improve clarity here, but I'm a little confused about why it's in the middle of `Device` classes.

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -4,60 +4,106 @@
...
+
+#define PFN_PARAM(proc) PFN_##proc proc
+        [[nodiscard]] explicit NvapiVulkanLowLatencyFakeDevice(
+            VkDevice device,
+            VkSemaphore semaphore,
+            PFN_PARAM(vkDestroySemaphore),
+            PFN_PARAM(vkSignalSemaphore));
+#undef PFN_PARAM
+
+        [[nodiscard]] VkSemaphore GetSemaphore() const;
```
Saancreed: Same here, base class implementation should be sufficient.

src/nvapi_vulkan.cpp
```diff
@@ -2,10 +2,23 @@
...
+            ret static_cast<NvapiVulkanLowLatency2LayerDevice*>(lowLatencyDevice)->__VA_ARGS__; \
+            break;                                                                              \
+        case VkReflexFake:                                                                      \
+            ret static_cast<NvapiVulkanLowLatencyFakeDevice*>(lowLatencyDevice)->__VA_ARGS__;   \
+            break;                                                                              \
+        default:                                                                                \
+            __builtin_unreachable();                                                            \
+    }
+#define LOW_LATENCY_DEVICE_INVOKE0(lowLatencyDevice, ...) LOW_LATENCY_DEVICE_INVOKE_(, lowLatencyDevice, __VA_ARGS__)
+#define LOW_LATENCY_DEVICE_INVOKE1(ret, lowLatencyDevice, ...) LOW_LATENCY_DEVICE_INVOKE_(ret =, lowLatencyDevice, __VA_ARGS__)
```
Saancreed: An idea: maybe this macro could be moved to `NvapiVulkanLowLatencyDevice`? Then the base class would have _non-virtual_ methods for each one we'd like to dispatch, and we'd be able to have the implementation of each one simply return the result of direct call, and I could we could get away with just one macro for both `void` and non-`void` methods. Like so:

```cpp
#define INVOKE(call)                                                            \
    switch (m_type) /* can read directly instead of calling GetType */ {        \
        case LowLatencyDeviceType::LowLatency2:                                 \
            return static_cast<NvapiVulkanLowLatency2LayerDevice*>(this)->call; \
        case LowLatencyDeviceType::VkReflexFake:                                \
            return static_cast<NvapiVulkanLowLatencyFakeDevice*>(this)->call;   \
        default:                                                                \
            __builtin_unreachable();                                            \
    }

…

VkResult NvapiVulkanLowLatencyDevice::LatencySleep(uint64_t value) {
    INVOKE(LatencySleep(value))
}

…

void NvapiVulkanLowLatencyDevice::QueueNotifyOutOfBand(VkQueue queue, NV_VULKAN_OUT_OF_BAND_QUEUE_TYPE queueType) {
    INVOKE(QueueNotifyOutOfBand(queue, queueType))
}
```

src/nvapi_private.h
```diff
@@ -61,6 +62,7 @@
 #if defined(__GNUC__)
 #define _ReturnAddress() __builtin_return_address(0);
 #elif defined(_MSC_VER)
+#define __builtin_unreachable __assume(false);
```
Saancreed: I think this should be

```suggestion
#define __builtin_unreachable() __assume(false);
```

but I'm also wondering if we shouldn't do the same thing https://en.cppreference.com/cpp/utility/unreachable does and add `!defined(__clang__)` to this.

Fingers crossed we get to switch to C++23 one day.

src/nvapi/nvapi_vulkan_low_latency_device.h
jp7677: I guess you mean like this?

```c++
NvapiVulkanLowLatencyDevice(const NvapiVulkanLowLatencyDevice&) = delete;
NvapiVulkanLowLatencyDevice& operator=(const NvapiVulkanLowLatencyDevice&) = delete;
```

Saancreed: Yes. Move constructors (and move-assignments) too.

src/nvapi_vulkan.cpp
jp7677: Cool idea, from what I'm seeing though, unfortunately this won't work like this because the derived classes are not yet known at this point.

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -4,60 +4,94 @@
...
       public:
-        static bool Initialize(NvapiResourceFactory& resourceFactory);
-        static void Reset();
+#define PFN_PARAM(proc) PFN_##proc proc
+        explicit NvapiVulkanLowLatencyDevice(LowLatencyDeviceImplementation type, VkDevice device, VkSemaphore semaphore, PFN_PARAM(vkDestroySemaphore))
+            : m_device(device), m_semaphore(semaphore), m_type(type), m_vkDestroySemaphore(vkDestroySemaphore) {}
+#undef PFN_PARAM
+        // Prevent default construction and copy semantics.
+        NvapiVulkanLowLatencyDevice() = delete;
+        NvapiVulkanLowLatencyDevice(const NvapiVulkanLowLatencyDevice&) = delete;
```
Saancreed, line 24: Move constructor too I suppose…

```suggestion
        NvapiVulkanLowLatencyDevice(const NvapiVulkanLowLatencyDevice&) = delete;
        NvapiVulkanLowLatencyDevice(NvapiVulkanLowLatencyDevice&&) = delete;
```

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -4,60 +4,94 @@
...
+        NvapiVulkanLowLatencyDevice& operator=(const NvapiVulkanLowLatencyDevice&) = delete;
+        NvapiVulkanLowLatencyDevice& operator=(NvapiVulkanLowLatencyDevice&&) = delete;
+
+        [[nodiscard]] LowLatencyDeviceImplementation GetType() const { return m_type; }
```
Saancreed: Nitpicking on the naming:

```suggestion
        [[nodiscard]] LowLatencyDeviceImplementation GetImplementation() const { return m_implementation; }
```

src/nvapi_vulkan.cpp
Saancreed: It does work if the implementation is in CPP file. At which point, all the types are fully known.

jp7677: Doh :)

Saancreed: Demo here: https://github.com/jp7677/dxvk-nvapi/pull/377/changes/e6415632293afaf973947df51b883361e264459d

Saancreed (approved):

PR #385, opened 2026-07-17
Add alternative VkReflex implementation using legacy VK_NV_low_latency ext

Saancreed:
Should make the layer obsolete on sufficiently new drivers and winevulkan…

TODO: add more tests

### DIFF
--- a/src/nvapi/nvapi_vulkan_low_latency_device.cpp
+++ b/src/nvapi/nvapi_vulkan_low_latency_device.cpp
@@ -6,6 +6,8 @@ namespace dxvk {
 
 #define INVOKE(call)                                                            \
     switch (m_implementation) {                                                 \
+        case LowLatencyDeviceImplementation::LowLatencyLegacy:                  \
+            return static_cast<NvapiVulkanLowLatencyLegacyDevice*>(this)->call; \
         case LowLatencyDeviceImplementation::LowLatency2:                       \
             return static_cast<NvapiVulkanLowLatency2LayerDevice*>(this)->call; \
         case LowLatencyDeviceImplementation::VkReflexFake:                      \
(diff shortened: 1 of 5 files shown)


src/nvapi/nvapi_vulkan_low_latency_device.cpp
```diff
@@ -69,6 +71,115 @@ namespace dxvk {
...
+    std::pair<std::unique_ptr<NvapiVulkanLowLatencyLegacyDevice>, VkResult> NvapiVulkanLowLatencyLegacyDevice::TryCreate(Vk* vk, VkDevice device) {
+        VK_GET_DEVICE_PROC_ADDR(vkCreateSemaphore);
+        VK_GET_DEVICE_PROC_ADDR(vkDestroySemaphore);
+
+        if (!vkCreateSemaphore || !vkDestroySemaphore) {
+            log::info("Initializing Vulkan Low-Latency with VkNvLowLatencyLegacy implementation failed: device does not appear to support semaphores?!");
+            return {nullptr, VK_ERROR_INCOMPATIBLE_DRIVER};
+        }
+
+        // If the Vulkan Reflex Layer is present it will enable VK_NV_low_latency2 which will let us query
```
wcampbell-nv: This comment is stale, these functions come from VK_NV_low_latency. Games using the Vulkan Reflex NvAPIs should already request that, the one exception I've found is Red Dead Redemption 2.

wcampbell-nv (commented):
The core implementation looks good to me, though I didn't look at the tests.

src/nvapi/nvapi_vulkan_low_latency_device.cpp
Saancreed: Right, copy-paste derp, will fix :sweat_smile:

src/nvapi/nvapi_vulkan_low_latency_device.h
```diff
@@ -43,6 +44,48 @@ namespace dxvk {
...
+
+        [[nodiscard]] NvBool GetLowLatencyModeImpl();
+        [[nodiscard]] VkResult SetLatencySleepModeImpl(std::nullptr_t);
+        [[nodiscard]] VkResult SetLatencySleepModeImpl(bool lowLatencyMode, bool lowLatencyBoost, uint32_t minimumIntervalUs);
+        [[nodiscard]] VkResult LatencySleepImpl(uint64_t value);
+        void GetLatencyTimingsImpl(std::span<NV_VULKAN_LATENCY_RESULT_PARAMS_V1::vkFrameReport, 64> frameReports);
+        [[nodiscard]] bool SetLatencyMarkerImpl(uint64_t frameID, NV_VULKAN_LATENCY_MARKER_TYPE marker);
+        void QueueNotifyOutOfBandImpl(VkQueue queue, NV_VULKAN_OUT_OF_BAND_QUEUE_TYPE queueType);
+
+        ~NvapiVulkanLowLatencyLegacyDevice() { m_vkShutdownLatencyDeviceLegacyNV(m_device); }
```
jp7677:
```suggestion
        ~NvapiVulkanLowLatencyLegacyDevice() override { m_vkShutdownLatencyDeviceLegacyNV(m_device); }
```

tests/nvapi_vulkan.cpp
```diff
@@ -109,252 +151,275 @@ TEST_CASE("Vulkan methods succeed", "[.vulkan]") {
         REQUIRE(NvAPI_Vulkan_InitLowLatencyDevice(vkDevice.get(), reinterpret_cast<HANDLE*>(&result)) == NVAPI_NOT_SUPPORTED);
     }
 
-    SECTION("InitLowLatencyDevice and DestroyLowLatencyDevice returns OK") {
+    SECTION("VkReflex with LowLatency2Layer") {
```
jp7677:
```suggestion
    SECTION("VkReflex with LowLatencyLegacy") {
```

jp7677 (commented):
The implementation fits nicely! Do you plan to extend the tests as part of this PR?

Saancreed:
> Do you plan to extend the tests as part of this PR?

Yes, although I got distracted by other stuff. I will add some over the coming days.

oscarbg:
> Should make the layer obsolete

@Saancreed just curious and maybe stupid question: but what's special in using the new "legacy VK_NV_low_latency ext" vs existing VK_NV_low_latency2 that makes the layer obsolete?
i.e. I understand no need to install/use the layer for getting Reflex working on Vulkan apps supporting it via NVAPI or "NvLowLatencyVk.dll"

EDIT: also I understand new winevulkan (as added in 357) but "sufficiently new drivers"? so existing drivers supporting VK_NV_low_latency aren't enough?

Saancreed:
> > Should make the layer obsolete
> 
> @Saancreed just curious and maybe stupid question: but what's special in using the new "legacy VK_NV_low_latency ext" vs existing VK_NV_low_latency2 that makes the layer obsolete? i.e. I understand no need to install/use the layer for getting Reflex working on Vulkan apps supporting it via NVAPI or "NvLowLatencyVk.dll"

@oscarbg Two things. One, titles integrating Vulkan Reflex SDK already [enable](https://github.com/NVIDIA-RTX/Streamline/blob/v2.12.0/source/plugins/sl.reflex/reflexEntry.cpp#L274) this [extension](https://github.com/NVIDIAGameWorks/dxvk-remix/blob/remix-1.5.2/src/dxvk/dxvk_extensions.h#L347), with one known exception of RDR2 which will still need some very smol layer that only enables the extension and does nothing else.

The other one is that legacy entrypoints don't need `VkSwapchainKHR` to be passed to them, so we don't have to keep track of swapchains that applications create anymore, or depend on invalid Vulkan usage just to get to the Unix side of winevulkan.

> EDIT: also I understand new winevulkan (as added in 357) but "sufficiently new drivers"? so existing drivers supporting VK_NV_low_latency aren't enough?

Winevulkan 357 and drivers supporting revision 2 of VK_NV_low_latency are both required. For example, 610.43.03 driver released earlier this month supports only revision 1 and therefore doesn't match the requirements.

mati865:
To be fair 610.43.03 is just a hotfix to a quite dated driver release, and revision 2 of the extension was released a [bit less than a month ago](https://docs.vulkan.org/refpages/latest/refpages/source/VK_NV_low_latency.html#_version_history).
So, currently this can be read as a "future driver release", without specific details.

Saancreed:
Yeah, you're right.

That said. there's nonzero chance that the latest Vulkan Developer/Beta driver, 595.44.11, will work. Just… don't look too closely at what revision of the ext it claims to support.

tests/nvapi_vulkan.cpp
```diff
@@ -109,252 +151,468 @@ TEST_CASE("Vulkan methods succeed", "[.vulkan]") {
...
+                setSleepModeParams.bLowLatencyBoost = false;
+                REQUIRE(NvAPI_Vulkan_SetSleepMode(vkDevice.get(), &setSleepModeParams) == NVAPI_ERROR);
+
+                REQUIRE(NvAPI_Vulkan_GetSleepStatus(vkDevice.get(), &getSleepStatusParams) == NVAPI_OK);
+                REQUIRE(getSleepStatusParams.bLowLatencyMode == true);
+
+                REQUIRE(NvAPI_Vulkan_DestroyLowLatencyDevice(vkDevice.get()) == NVAPI_OK);
+            }
+
+            SECTION("GetSleepStatus with unknown struct version returns incompatible-struct-version") {
```
jp7677: Just having one version of those `... with unknown struct version returns incompatible-struct-version` / `.. with current struct version returns not incompatible-struct-version` would be sufficient since it's not related to the used implementation.

jp7677 (approved):
LGTM and feel free to ignore my comment and let this land.

tests/nvapi_vulkan.cpp
Saancreed: An idea would be to extract this to a separate section, as the version check should kick in even before we check if the device was initialized, and with which backing implementation.

jp7677: yeah having a separate section would be nice

PR #386, opened 2026-07-30
meta: Update README for VkReflex legacy

Saancreed:
Thinking about it, maybe I could hack our layer enough to just force-enable `VK_NV_low_latency` if revision 2 is supported as an option specifically for RDR2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ This implementation currently offers entry points for supporting the following f
 
 - NVIDIA DLSS for Vulkan, by supporting the relevant adapter information by querying from [Vulkan](https://www.vulkan.org/).
 - NVIDIA DLSS for D3D11 and D3D12, by querying from Vulkan and forwarding the relevant calls into DXVK / VKD3D-Proton.
-- NVIDIA Reflex, by forwarding the relevant calls into either DXVK / VKD3D-Proton or its custom Vulkan layer.
+- NVIDIA Reflex, by forwarding the relevant calls into either DXVK / VKD3D-Proton or a supported Vulkan implementation.
 - Several NVAPI D3D11 extensions, among others `SetDepthBoundsTest` and `UAVOverlap`, by forwarding the relevant calls into DXVK.
 - NVIDIA PhysX, by supporting entry points for querying PhysX capabilities.
 - Several GPU topology related methods for adapter and display information, by querying from DXVK and Vulkan.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,10 @@ While originally being developed for usage with Unreal Engine 4, most notably fo
...
+In order to support Vulkan flavor of NVIDIA Reflex, one of two backing implementations can be used:
+
+- A Vulkan driver that supports revision 2 of [`VK_NV_low_latency`](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_NV_low_latency.html) device extension and winevulkan 1.4.357 or newer. This relies on the application enabling this extension on its own, which notably prevents Red Dead Redemption 2 from working using this method out of the box. Going forward, this will be the preferred choice.
+- An [additional Vulkan layer](./layer) is which intercepts (technically invalid) Vulkan calls made by DXVK-NVAPI, enriches them with extra data, and forwards to a Vulkan driver that supports [`VK_NV_low_latency2`](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_NV_low_latency2.html) device extension. This layer will _replace_ request to enable `VK_NV_low_latency` with itself. This implementation is soft-deprecated and may be removed once support for the first one is widespread enough.
```
jp7677: `This implementation is soft-deprecated and may be removed once support for the first one is widespread enough.` I would leave that out for the readme (but add that for the next release notes) since the first point already mentions what the way forward is.

jp7677 (approved):
Reads good! See suggestion.

README.md
Saancreed: Ok, done.

README.md
```diff
@@ -23,7 +23,10 @@ While originally being developed for usage with Unreal Engine 4, most notably fo
...
+In order to support Vulkan flavor of NVIDIA Reflex, one of two backing implementations can be used:
+
+- A Vulkan driver that supports revision 2 of [`VK_NV_low_latency`](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_NV_low_latency.html) device extension and winevulkan 1.4.357 or newer. This relies on the application enabling this extension on its own, which notably prevents Red Dead Redemption 2 from working using this method out of the box. Going forward, this will be the preferred choice.
+- An [additional Vulkan layer](./layer) is which intercepts (technically invalid) Vulkan calls made by DXVK-NVAPI, enriches them with extra data, and forwards to a Vulkan driver that supports [`VK_NV_low_latency2`](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_NV_low_latency2.html) device extension. This layer will _replace_ request to enable `VK_NV_low_latency` with itself.
```
jp7677: Sry, just spotted a typo which I missed before: `is which intercepts` -> `which intercepts`. Additionally I think `requests` instead of `request` reads better.

Saancreed: Oops, my bad. Fixed the first one, no strong opinion about the second one but I changed it too.

PR #387, opened 2026-08-04
github: Use ming-action v8

jp7677:
No need to use master anymore, Wine dependencies are resolved. As a side effect, no AUR packages in the build process anymore. So essentially dropping the last 4 commits from https://github.com/misyltoad/arch-mingw-github-action/commits/master/

While at it, update catch2 and also declare a function static.

### DIFF
--- a/.github/workflows/build-test-package-linux.yml
+++ b/.github/workflows/build-test-package-linux.yml
@@ -32,14 +32,14 @@ jobs:
         echo "VERSION=${VERSION}" >> $GITHUB_ENV
 
     - name: Lint source code
-      uses: misyltoad/arch-mingw-github-action@617caabb7a2f459c4e6dca76e0eb4974d7dc20b2
+      uses: misyltoad/arch-mingw-github-action@v8
       with:
         command: |
           meson setup --cross-file "./build-win64.txt" -Denable_tests=True build/lint
           ninja clang-format-check -C build/lint
 
     - name: Build and run tests
-      uses: misyltoad/arch-mingw-github-action@617caabb7a2f459c4e6dca76e0eb4974d7dc20b2
+      uses: misyltoad/arch-mingw-github-action@v8
       with:
         command: |
           ./package-release.sh "${{ env.VERSION }}" build --enable-tests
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test-package-linux.yml
```diff
@@ -32,14 +32,14 @@ jobs:
         echo "VERSION=${VERSION}" >> $GITHUB_ENV
 
     - name: Lint source code
-      uses: misyltoad/arch-mingw-github-action@617caabb7a2f459c4e6dca76e0eb4974d7dc20b2
+      uses: Joshua-Ashton/arch-mingw-github-action@v8
```
Saancreed:
```suggestion
      uses: misyltoad/arch-mingw-github-action@v8
```

.github/workflows/build-test-package-linux.yml
```diff
@@ -32,14 +32,14 @@ jobs:
...
       with:
         command: |
           meson setup --cross-file "./build-win64.txt" -Denable_tests=True build/lint
           ninja clang-format-check -C build/lint
 
     - name: Build and run tests
-      uses: misyltoad/arch-mingw-github-action@617caabb7a2f459c4e6dca76e0eb4974d7dc20b2
+      uses: Joshua-Ashton/arch-mingw-github-action@v8
```
Saancreed:
```suggestion
      uses: misyltoad/arch-mingw-github-action@v8
```

Saancreed (approved):
LGTM but see the comments.

.github/workflows/build-test-package-linux.yml
jp7677: I tried that initially, but seems that v8 of that action has only been published with the old name (See https://github.com/jp7677/dxvk-nvapi/actions/runs/30940218085/job/92096402766)

Saancreed: I think that's because you had a typo, `misyltoad/arch-mingw-github-action@8` instead of `misyltoad/arch-mingw-github-action@v8` (missing `v`).

jp7677: Derp... , thanks for not being fooled by my reasoning ;)

Issue #233, opened 2024-12-02
Requesting exposure of SMP/MVP in dxvk-nvapi


DanFraserUK:
Hello!

First go at submitting an issue here, hopefully I can explain this well.  I am far from technically experienced enough on this subject to go into detail and pick out certain code :(

iRacing is a game that implements SMP/MVP (simultaneous multi-projection, multi view projection - these are a pain to google).  It's VR _related_ but the specific purpose in this regard is for the triple monitor functionality, where on top of the functional Render View Mode that wraps the game across angled monitors this SMP/MVP option in iRacing assists with rendering process time with having 3 "cameras"/viewports, very similar to VR rendering.

[An announcement (?) link](https://forums.developer.nvidia.com/t/nvidia-smp-assist-api-for-vr-programming/148663)

It would _greatly_ assist with presenting iRacing being functional on Linux!

Looking through the API pages on nvidia.com and https://github.com/NVIDIA/nvapi NV_SMP_ASSIST* is the stuff not in dxvk-nvapi.

I hope you can help!!

Saancreed:
Hi, considering that this feature looks like something that would have to be done with most of the heavy lifting on DXVK side of things (just like we "expose" Multi Draw Indirect and similar extensions by forwarding the relevant calls to DXVK/vkd3d-proton), I asked Philip and he said that implementing such a thing would be (paraphrasing) a huge pain and a very invasive change that would require months of work just for one or two games to benefit from it, and therefore is not worth the cost.

sp82:
Hi, I came here to ask the same question for the same reason, iracing!
Saancreed are you sure it's require too much work? Vulkan expose already the multiview functionality, can it be used to implement this feature?

sp82:
I find out that in the specifications of the kronos vulkan multiview extension (https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_multiview.html)  linked to the page that nvidia explain nvapi multiview: https://developer.nvidia.com/blog/turing-multi-view-rendering-vrworks
90% are identical in implementation.

Saancreed:
As per my previous comment, there is little-to-nothing dxvk-nvapi can do until DXVK supports this thing. Feel free to file a feature request there, but don't be surprised when the answer is "no". DLSS support is already enough of a headache, so I doubt anyone over there would be willing to add and maintain something as… exotic as this.

DanFraserUK:
[...]

PR #388, opened 2026-08-05
nvapi: Forward NVAPI multiview calls to DXVK

DanFraserUK:
This PR is to enable SMP/MVP support in iRacing by exposing the SMP capability to iRacing and also forwards the calls from iRacing to DXVK.

As far as I can tell iRacing is the only game to use SMP/MVP for triple screen gaming.  These changes only affect users with an NVIDIA card of the Turing generation or newer and nothing changes for any users not using this feature either with an incompatible graphics card or SMP simply being disabled in game.

Multiple test runs and the DXVK_HUD and Mangohud data collected from them are in this graphic:

smp-triple-screen

(performance changes are as expected when enabling SMP, FPS should increase significantly over SMP off)

https://github.com/doitsujin/dxvk/pull/5828 is necessary for a reviewer to be aware of as both the DXVK and dxvk-nvapi changes need to work together.

This work is the result of my issue #233 18 months ago.  I understood and accepted the reasoning given there as it was a very fair comment to make.  It did inspire me to work on doing it myself though!  I have been spearheading an unofficial campaign to ask for iRacing to reintroduce some form of support regarding Linux with the most recent path asking for the EOS Linux module to be built.  iRacing hasn't pushed back on anticheat being the reason they won't let users play on Linux but instead pointing to "third party APIs, hardware support" (etc).  This is part of my response by looking to close off excuses and narrow down their replies to simply being "we won't do it because of cheaters" or them running dry of excuses.

I'm happy to do further testing that may be needed.


jp7677:
Hi, thanks for the PR. Getting this landed obviously depends on how the discussion on DXVK side goes. This is indeed a lot of code for just a single title with a very specific setup. But if it lands there, cool to add our side as well.

A few generals notes from briefly looking at the changes:
- some of the comments are a bit off or don't add anything additionally to the actual code
- the fallback path is supposed for DXVK versions that don't have the needed interfaces or features? In that case, I don't think we have to prepare for that. Just returning no-implementation when requirements are not given should be ok.
- we have a lot of tests, mostly as a safety net for future refactorings (I don't trust myself), so the new entry points in this PR should also be added to the tests. Feel free to ask about the test suite setup, it can be a bit overwhelming.

DanFraserUK:
Hi,

I'm going to say the same as my comment on the DXVK side, apologies for dumping something so big and so specific on you like this.  My knowledge of the processes for this side of things is certainly lacking.

As with the DXVK comment, this PR is included in my decision to go away and work on improving the PRs as now it is clear both of you have made some very valid points about how this has been presented and I need to resolve them.

For the concerns from both of you about the changes being very specific to not only one game but also the unique architecture being used I have already completed a quick test using lavapipe to do the rendering and aside from the completely hilarious software rendering speed of 45-50s per frame, that worked.  That was done by installing lavapipe, editing launch options and a temporary code edit to force Turing detection and continue anyway.


I have released a custom proton tool for users to test with these changes and knowing people have downloaded it, I am taking the no news is good news approach as iRacing users are very vocal if something doesn't work, but don't say anything when it does!

So with the lavapipe result, there is very good potential for me to resolve the device agnostic issue in DXVK and your quick skim over is already noted down to review those parts.  Those were part of the mistakes of me being so heavily focused on the single task to complete the goal.

I hope you can understand I am very passionate about the end goal of showing iRacing that their game can successfully work on Linux but now I know things need to be somewhat different compared to merely just slapping code on a website, any help in understanding and completing what *should* be done will be greatly appreciated!  I want to get this right!

Thank you for your time!

jp7677:
All good. Having things in a working state and then refining / splitting it up is not a wrong approach.

I started this whole project to just get a few more fps in Assetto Corsa Competizione on Linux/NV, so I certainly understand the passion :)

PR #389, opened 2026-08-23
github: Drop arch-mingw-github

jp7677:
This pulls in less packages, thus faster, and avoids AUR.

### DIFF
--- a/.github/workflows/build-test-package-linux.yml
+++ b/.github/workflows/build-test-package-linux.yml
@@ -11,8 +11,12 @@ on:
 jobs:
   build-test-package-linux:
     runs-on: ubuntu-latest
+    container: docker.io/library/archlinux:base-devel
 
     steps:
+    - name: Install build tooling
+      run: pacman -Syu --needed --noconfirm git clang mingw-w64-gcc meson ninja wine
+
     - name: Checkout code
       uses: actions/checkout@v7
       with:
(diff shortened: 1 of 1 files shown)


Saancreed (approved):
Makes sense, so 👍 from me.

Or, we could just drop the action and install the deps ourselves. All we need is `archlinux:base-devel` and `pacman -S git meson mingw-w64-gcc ninja`. Plus maybe `clang-format` for linting and `wine` for running tests. [This is how I do this in wine-nvml](https://github.com/Saancreed/wine-nvml/blob/a1a7d8fac0e054122cbaf9205c9bee33f5831015/.github/workflows/build.yaml#L22-L23), I'd consider it somewhat simpler than our current setup.

WinterSnowfall:
> Or, we could just drop the action and install the deps ourselves. All we need is `archlinux:base-devel` and `pacman -S git meson mingw-w64-gcc ninja`. Plus maybe `clang-format` for linting and `wine` for running tests. [This is how I do this in wine-nvml](https://github.com/Saancreed/wine-nvml/blob/a1a7d8fac0e054122cbaf9205c9bee33f5831015/.github/workflows/build.yaml#L22-L23), I'd consider it somewhat simpler than our current setup.

Probably best to just do this IMHO. I've only created that fork as a stop gap measure and can't guarantee it will be around for all eternity. dxvk/vkd3d-proton will most likely gravitate towards the same approach of installing the deps directly on the base image in the near future.

WinterSnowfall:
FYI, not sure whether you are using git anywhere within the CI flows, but if you are it's probably regressed now and needs adding the GitHub workspace as a safe directory: https://github.com/doitsujin/dxvk/pull/5905

jp7677:
> FYI, not sure you are using git anywhere within the CI flows, but if you are it's probably regressed now and needs adding the GitHub workspace as a safe directory: [doitsujin/dxvk#5905](https://github.com/doitsujin/dxvk/pull/5905)

Thanks a lot for the heads up, addressed in https://github.com/jp7677/dxvk-nvapi/pull/397 with an additional step that fails when git wont work with us. So no more silent drops.

PR #390, opened 2026-08-30
nvapi-d3d12: Correct m_nvapiDeviceMap invalidation

mkualquiera:
This PR fixes a crash related to ReShade. 

My understanding of what happened before is:
1. The application calls and ReShade intercepts `CreateCommandList`, grabs the real `NvapiD3d12GraphicsCommandList*` and returns a proxy to the application.
2. `m_nvapiDeviceMap` ends up mapping the proxy object's address to the real object pointer. 
3. The game drops the proxy object, which also drops the real object. But the entry remains in `m_nvapiDeviceMap`.
4. Eventually ReShade creates a new proxy object that happens to be on the same address as the earlier dropped one, but which wraps a different command list.
5. The next time `m_nvapiDeviceMap` is accessed, it yields the old `NvapiD3d12GraphicsCommandList*` that was actually freed. Dereferencing it for a vtable finally results in a crash.

The fix is to tighten the cache logic a bit. It moves the QueryInterface up as well, but I believe this should be fine since the expensive part is `probeInterfaceChain`, `GetDevice`, etc. 

This has indeed fixed my crashes in Cyberpunk 2077 and Kingdom Come: Deliverance II when using ReShade + RenoDX DLSS5 addon.

Thanks.

### DIFF
--- a/src/nvapi/nvapi_d3d12_command_queue.cpp
+++ b/src/nvapi/nvapi_d3d12_command_queue.cpp
@@ -11,15 +11,20 @@ namespace dxvk {
     }
 
     NvapiD3d12CommandQueue* NvapiD3d12CommandQueue::GetOrCreate(ID3D12CommandQueue* commandQueue) {
-        std::scoped_lock lock{m_mutex};
-
-        auto itF = m_nvapiDeviceMap.find(commandQueue);
-        if (itF != m_nvapiDeviceMap.end())
-            return &itF->second;
-
+        // Same non-owning-key hazard as NvapiD3d12GraphicsCommandList::GetOrCreate; see there.
         Com<ID3D12CommandQueueExt> commandQueueExt;
         if (FAILED(commandQueue->QueryInterface(IID_PPV_ARGS(&commandQueueExt))))
             return nullptr;
+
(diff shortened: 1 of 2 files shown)


jp7677:
Thanks for the PR. You've touched a valid point here, the caching is indeed somewhat fragile but up until now it did the job :). Give us some time to think about how to solve this properly.

Saancreed:
@mkualquiera Could you please check if https://github.com/jp7677/dxvk-nvapi/tree/d3d12-private-data-storage also resolves your issue? We'd ideally get rid of the cache maps altogether, and our best chance at it is probably D3D12 private data store. I'd just prefer if we didn't exchange one set of issues for another if we can help it.

mkualquiera:
@Saancreed I've checked the code statically and I believe it does indeed solve the issue. Seems like a better solution too. Given that my patch works for me I'm honestly a bit lazy to actually build this and test it with the games though haha. Unless you do want me to, but I think it looks good. Thanks!

jp7677:
This has been resolved differently in https://github.com/jp7677/dxvk-nvapi/pull/391
Thanks for investigation that lead to this path!

PR #391, opened 2026-09-02
Store D3D12 wrappers in private data

Saancreed:
Should be more robust than caching stuff in maps. But, this prevents us from storing anything fancy like `NvapiAsConverter` there, so I've made it `thread_local` as a workaround. The fact that it doesn't store anything list-specific makes this… not much of a problem.

Plus fixes one naming derp I caught while working on this.

Supersedes https://github.com/jp7677/dxvk-nvapi/pull/390.

### DIFF
--- a/src/nvapi/nvapi_d3d12_command_queue.cpp
+++ b/src/nvapi/nvapi_d3d12_command_queue.cpp
@@ -2,34 +2,45 @@
 #include "../util/com_pointer.h"
 
 namespace dxvk {
-    std::unordered_map<ID3D12CommandQueue*, NvapiD3d12CommandQueue> NvapiD3d12CommandQueue::m_nvapiDeviceMap;
+    std::atomic<LONGLONG> NvapiD3d12CommandQueue::m_resetTimestamp;
     std::mutex NvapiD3d12CommandQueue::m_mutex;
 
     void NvapiD3d12CommandQueue::Reset() {
-        std::scoped_lock lock{m_mutex};
-        m_nvapiDeviceMap.clear();
+        LARGE_INTEGER count;
+        QueryPerformanceCounter(&count);
+        m_resetTimestamp.store(count.QuadPart, std::memory_order_release);
     }
 
-    NvapiD3d12CommandQueue* NvapiD3d12CommandQueue::GetOrCreate(ID3D12CommandQueue* commandQueue) {
+    std::optional<NvapiD3d12CommandQueue> NvapiD3d12CommandQueue::GetOrCreate(ID3D12CommandQueue* commandQueue) {
+        static constexpr GUID NvapiD3d12CommandQueueGuid = {0x3579ec19, 0x2741, 0x488a, {0xbb, 0x39, 0x5c, 0x03, 0xcd, 0x89, 0x5e, 0x2a}};
+
+        NvapiD3d12CommandQueue nvapiCommandQueue{nullptr};
+        UINT size = sizeof(nvapiCommandQueue);
+
+        auto result = commandQueue->GetPrivateData(NvapiD3d12CommandQueueGuid, &size, &nvapiCommandQueue);
+        if (SUCCEEDED(result) && size == sizeof(nvapiCommandQueue) && nvapiCommandQueue.m_creationTimestamp > m_resetTimestamp.load(std::memory_order_acquire))
+            return nvapiCommandQueue;
+
+        size = sizeof(nvapiCommandQueue);
         std::scoped_lock lock{m_mutex};
 
-        auto itF = m_nvapiDeviceMap.find(commandQueue);
-        if (itF != m_nvapiDeviceMap.end())
-            return &itF->second;
+        result = commandQueue->GetPrivateData(NvapiD3d12CommandQueueGuid, &size, &nvapiCommandQueue);
+        if (SUCCEEDED(result) && size == sizeof(nvapiCommandQueue) && nvapiCommandQueue.m_creationTimestamp > m_resetTimestamp.load(std::memory_order_acquire))
+            return nvapiCommandQueue;
 
         Com<ID3D12CommandQueueExt> commandQueueExt;
         if (FAILED(commandQueue->QueryInterface(IID_PPV_ARGS(&commandQueueExt))))
-            return nullptr;
+            return std::nullopt;
 
-        auto [itI, inserted] = m_nvapiDeviceMap.emplace(
-            std::piecewise_construct,
-            std::forward_as_tuple(commandQueue),
-            std::forward_as_tuple(commandQueueExt.ptr()));
+        LARGE_INTEGER count;
+        QueryPerformanceCounter(&count);
 
-        if (!inserted)
-            return nullptr;
+        nvapiCommandQueue = NvapiD3d12CommandQueue{commandQueueExt.ptr()};
+        nvapiCommandQueue.m_creationTimestamp = count.QuadPart;
 
-        return &itI->second;
+        commandQueue->SetPrivateData(NvapiD3d12CommandQueueGuid, sizeof(nvapiCommandQueue), &nvapiCommandQueue);
+
+        return nvapiCommandQueue;
     }
 
     NvapiD3d12CommandQueue::NvapiD3d12CommandQueue(ID3D12CommandQueueExt* vkd3dCommandQueue)
--- a/src/nvapi/nvapi_d3d12_command_queue.h
+++ b/src/nvapi/nvapi_d3d12_command_queue.h
@@ -8,7 +8,7 @@ namespace dxvk {
 
       public:
         static void Reset();
-        [[nodiscard]] static NvapiD3d12CommandQueue* GetOrCreate(ID3D12CommandQueue* device);
+        [[nodiscard]] static NvapiD3d12CommandQueue* GetOrCreate(ID3D12CommandQueue* commandQueue);
 
         explicit NvapiD3d12CommandQueue(ID3D12CommandQueueExt* vkd3dCommandQueue);
 
--- a/src/nvapi/nvapi_d3d12_command_queue.h
+++ b/src/nvapi/nvapi_d3d12_command_queue.h
@@ -4,20 +4,21 @@
 #include "../interfaces/vkd3d-proton_interfaces.h"
 
 namespace dxvk {
-    class NvapiD3d12CommandQueue {
+    class NvapiD3d12CommandQueue final {
 
       public:
         static void Reset();
-        [[nodiscard]] static NvapiD3d12CommandQueue* GetOrCreate(ID3D12CommandQueue* commandQueue);
+        [[nodiscard]] static std::optional<NvapiD3d12CommandQueue> GetOrCreate(ID3D12CommandQueue* commandQueue);
 
         explicit NvapiD3d12CommandQueue(ID3D12CommandQueueExt* vkd3dCommandQueue);
 
         [[nodiscard]] HRESULT NotifyOutOfBandCommandQueue(D3D12_OUT_OF_BAND_CQ_TYPE type) const;
 
       private:
-        static std::unordered_map<ID3D12CommandQueue*, NvapiD3d12CommandQueue> m_nvapiDeviceMap;
+        static std::atomic<LONGLONG> m_resetTimestamp;
         static std::mutex m_mutex;
 
+        LONGLONG m_creationTimestamp{};
         ID3D12CommandQueueExt* m_vkd3dCommandQueue{};
     };
 }
--- a/src/nvapi/nvapi_d3d12_device.cpp
+++ b/src/nvapi/nvapi_d3d12_device.cpp
@@ -102,7 +102,8 @@ namespace dxvk {
 
         m_supportsNvxBinaryImport = vkd3dDevice->GetExtensionSupport(D3D12_VK_NVX_BINARY_IMPORT);
         m_supportsNvxImageViewHandle = vkd3dDevice->GetExtensionSupport(D3D12_VK_NVX_IMAGE_VIEW_HANDLE);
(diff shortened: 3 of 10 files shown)


src/nvapi/nvapi_d3d12_command_queue.cpp
```diff
@@ -2,34 +2,45 @@
 #include "../util/com_pointer.h"
 
 namespace dxvk {
-    std::unordered_map<ID3D12CommandQueue*, NvapiD3d12CommandQueue> NvapiD3d12CommandQueue::m_nvapiDeviceMap;
+    std::atomic<LONGLONG> NvapiD3d12CommandQueue::m_resetTimestamp;
     std::mutex NvapiD3d12CommandQueue::m_mutex;
 
     void NvapiD3d12CommandQueue::Reset() {
```
jp7677, line 8: Looks like we didn't reset everything yet on `UnLoad` in https://github.com/jp7677/dxvk-nvapi/blob/master/src/nvapi/nvapi_adapter_registry.cpp#L14-L17 , I guess we should do that now.

src/nvapi/nvapi_d3d12_command_queue.h
```diff
@@ -4,20 +4,21 @@
 #include "../interfaces/vkd3d-proton_interfaces.h"
 
 namespace dxvk {
-    class NvapiD3d12CommandQueue {
+    class NvapiD3d12CommandQueue final {
 
       public:
         static void Reset();
-        [[nodiscard]] static NvapiD3d12CommandQueue* GetOrCreate(ID3D12CommandQueue* device);
+        [[nodiscard]] static std::optional<NvapiD3d12CommandQueue> GetOrCreate(ID3D12CommandQueue* device);
```
jp7677: Another old naming derp (device -> queue).

src/nvapi/nvapi_d3d12_command_queue.cpp
```diff
@@ -2,34 +2,45 @@
...
+        LARGE_INTEGER count;
+        QueryPerformanceCounter(&count);
+        m_resetTimestamp.store(count.QuadPart, std::memory_order_release);
     }
 
-    NvapiD3d12CommandQueue* NvapiD3d12CommandQueue::GetOrCreate(ID3D12CommandQueue* commandQueue) {
+    std::optional<NvapiD3d12CommandQueue> NvapiD3d12CommandQueue::GetOrCreate(ID3D12CommandQueue* commandQueue) {
+        static constexpr GUID NvapiD3d12CommandQueueGuid = {0x3579ec19, 0x2741, 0x488a, {0xbb, 0x39, 0x5c, 0x03, 0xcd, 0x89, 0x5e, 0x2a}};
+
+        NvapiD3d12CommandQueue queue{nullptr};
```
jp7677: may be `nvapiCommandQueue` ?

src/nvapi/nvapi_d3d12_device.cpp
```diff
@@ -43,26 +45,36 @@ namespace dxvk {
         return m_cubin64bitSupportAvailable.emplace(false);
     }
 
-    NvapiD3d12Device* NvapiD3d12Device::GetOrCreate(ID3D12Device* device) {
+    std::optional<NvapiD3d12Device> NvapiD3d12Device::GetOrCreate(ID3D12Device* device) {
+        static constexpr GUID NvapiD3d12DeviceGuid = {0x0266efbf, 0x6dc7, 0x4017, {0xb4, 0x21, 0x3d, 0x93, 0xce, 0xd0, 0x90, 0x64}};
+
+        NvapiD3d12Device dev{nullptr};
```
jp7677: may be `nvapiDevice`?

src/nvapi/nvapi_d3d12_graphics_command_list.cpp
```diff
@@ -4,38 +4,53 @@
...
+        LARGE_INTEGER count;
+        QueryPerformanceCounter(&count);
+        m_resetTimestamp.store(count.QuadPart, std::memory_order_release);
     }
 
-    NvapiD3d12GraphicsCommandList* NvapiD3d12GraphicsCommandList::GetOrCreate(ID3D12GraphicsCommandList* commandList) {
+    std::optional<NvapiD3d12GraphicsCommandList> NvapiD3d12GraphicsCommandList::GetOrCreate(ID3D12GraphicsCommandList* commandList) {
+        static constexpr GUID NvapiD3d12GraphicsCommandListGuid = {0x1d409261, 0xd613, 0x4ef4, {0x86, 0x10, 0x25, 0x0e, 0x8b, 0x3b, 0x56, 0x56}};
+
+        NvapiD3d12GraphicsCommandList list{nullptr};
```
jp7677: May be `nvapiCommmandList`?

src/nvapi_d3d12.cpp
jp7677, line 652: `!device.has_value()` ?

tests/mocks/d3d12_mocks.h
```diff
@@ -3,14 +3,71 @@
 #include "../../src/nvapi_private.h"
 #include "../../src/interfaces/vkd3d-proton_interfaces.h"
 
+struct GuidHash {
+    std::size_t operator()(const GUID& guid) const {
+        uint64_t data4;
+        std::memcpy(&data4, guid.Data4, sizeof(data4));
+#define HASH(x) std::hash<decltype(x)>{}(x)
```
jp7677: I guess I would have written this out for readability:

```c++
        return std::hash<decltype(GUID::Data1)>{}(guid.Data1)
            ^ std::hash<decltype(GUID::Data2)>{}(guid.Data2)
            ^ std::hash<decltype(GUID::Data3)>{}(guid.Data3)
            ^ std::hash<decltype(data4)>{}(data4);
``` 

But cool to keep your version if you prefer.

tests/mocks/d3d12_mocks.h
```diff
@@ -3,14 +3,71 @@
...
+    }
+};
+
+struct GuidEquals {
+    bool operator()(const GUID& lhs, const GUID& rhs) const {
+        return !std::memcmp(&lhs, &rhs, sizeof(GUID));
+    }
+};
+
+#define DXVK_NVAPI_PRIVATE_DATA_IMPL                                                                            \
```
jp7677: What about using inline instead of a macro? Something line this?

```c++
inline std::unordered_map<GUID, std::vector<std::byte>, GuidHash, GuidEquals> m_privateData;
inline std::mutex m_privateDataMutex;

inline HRESULT STDMETHODCALLTYPE GetData(REFGUID guid, UINT* pDataSize, void* pData) {
    if (!pDataSize)
// ...
}

inline HRESULT STDMETHODCALLTYPE SetData(REFGUID guid, UINT DataSize, const void* pData) {
    auto ptr = static_cast<const std::byte*>(pData);
// ..
}

```

and then 

```c++
HRESULT STDMETHODCALLTYPE GetPrivateData(REFGUID guid, UINT* pDataSize, void* pData) override { return GetData(guid, pDataSize, pData); }
HRESULT STDMETHODCALLTYPE SetPrivateData(REFGUID guid, UINT DataSize, const void* pData) override { return SetData(guid, DataSize, pData); }
    
```

src/nvapi/nvapi_d3d12_device.cpp
```diff
@@ -77,6 +89,9 @@ namespace dxvk {
 
     NvapiD3d12Device::NvapiD3d12Device(ID3D12DeviceExt* vkd3dDevice)
         : m_vkd3dDevice(static_cast<ID3D12DeviceExt5*>(vkd3dDevice)) { // NOLINT(*-pro-type-static-cast-downcast)
+        if (!vkd3dDevice)
+            return;
```
jp7677, line 93: Thanks. I guess we should also change

```c++
bool NvapiD3d12Device::SetCreatePipelineStateFlagsNVAPI(D3D12_VK_EXT_PIPELINE_CREATION_STATE_FLAG pipeline_state_flags) {
        return m_vkd3dDevice->SetCreatePipelineStateFlagsNVAPI(pipeline_state_flags);
    }
```

to 

```c++
bool NvapiD3d12Device::SetCreatePipelineStateFlagsNVAPI(D3D12_VK_EXT_PIPELINE_CREATION_STATE_FLAG pipeline_state_flags) {
        return m_vkd3dDevice && m_vkd3dDevice->SetCreatePipelineStateFlagsNVAPI(pipeline_state_flags);
    }
```

other functions do look safe.

jp7677 (commented):
Thanks, very nice improvement. See comments for some suggestions.

src/nvapi_d3d12.cpp
Saancreed: To be fair, I was aiming for minimal diffs. As you probably know, they are equivalent, but there are _quite a few_ places where previously pointers, now optionals, are checked for nullptr/nullopt using `!expression` syntax. Should I change them all for clarity, or do we keep using the conversion-to-bool rule?

tests/mocks/d3d12_mocks.h
Saancreed: I don't like repeating myself :sweat_smile:

tests/mocks/d3d12_mocks.h
Saancreed: Inline where? Our three mock object types don't have any common base type. Or are you suggesting creating one?

(Admittably, I didn't pay too much attention to this file, clangd behaves like if it was dying whenever I make any edits there. I'm happy enough it compiles as is.)

src/nvapi/nvapi_d3d12_device.cpp
Saancreed: In theory there's also `VerifyOpacityMicromapArrayNVAPI`, which should probably do at least a nullptr check, and likely even an interface version, check before the call.

Saancreed: I kind of already had that as https://github.com/jp7677/dxvk-nvapi/commit/d7673dd9a50024d60f21d8b4a265837aae9aac5c in my branch, might as well cherry-pick it.

tests/mocks/d3d12_mocks.h
Saancreed: Went with a dedicated `PrivateDataStore` class that's now a member of each mock interface.

tests/mocks/d3d12_mocks.h
Saancreed: How about a generic lambda?

src/nvapi_d3d12.cpp
jp7677: No need to change all and I'm also ok with keeping conversion-to-bool to bool here, we are not really consistent anyway in that regard. I just find `has_value()` more readable with `std::optional`, and when touching that function, might be an opportunity to tweak that. That said, if you prefer to change all, please feel free to do so.

tests/mocks/d3d12_mocks.h
jp7677: Yeah that looks cool, thanks!

tests/mocks/d3d12_mocks.h
jp7677: I had locally tried with a single one and that worked well (for the tests). But `PrivateDataStore` looks also cool, thanks!

tests/mocks/d3d12_mocks.h
```diff
@@ -3,14 +3,79 @@
 #include "../../src/nvapi_private.h"
 #include "../../src/interfaces/vkd3d-proton_interfaces.h"
 
+struct GuidHash {
```
jp7677: `GuidHash` and `GuidEquals` could be part of `PrivateDataStore`.

tests/mocks/d3d12_mocks.h
```diff
@@ -3,14 +3,79 @@
...
+        if (data.size() > size)
+            return DXGI_ERROR_MORE_DATA;
+
+        std::memcpy(pData, data.data(), data.size());
+
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE SetPrivateData(const GUID& guid, UINT DataSize, const void* pData) {
+        auto ptr = reinterpret_cast<const std::byte*>(pData);
```
jp7677: `static_cast`

tests/mocks/d3d12_mocks.h
```diff
@@ -3,14 +3,79 @@
...
 class D3D12Vkd3dDeviceMock final : public trompeloeil::mock_interface<ID3D12Vkd3dDevice> {
+    PrivateDataStore m_store;
+
     MAKE_MOCK2(QueryInterface, HRESULT(REFIID, void**), override);
     MAKE_MOCK0(AddRef, ULONG(), override);
     MAKE_MOCK0(Release, ULONG(), override);
-    IMPLEMENT_MOCK3(GetPrivateData);
-    IMPLEMENT_MOCK3(SetPrivateData);
+
+    virtual HRESULT STDMETHODCALLTYPE GetPrivateData(const GUID& guid, UINT* pDataSize, void* pData) override {
```
jp7677: `virtual` is not needed (the mocks are final).

src/nvapi/nvapi_d3d12_device.cpp
jp7677: I kind of like the bools that describe a certain capability instead of using the tier numbers later on. What about bringing `m_supportsNvShaderExtn`  back and introducing something along the line of `m_supportsCreatePipelineStateFlagsNVAPI`

Saancreed: Did that, but I chose a name that describes a feature over just this particular method.

src/nvapi_d3d12.cpp
Saancreed: Might be something to leave on a TODO list for later.

jp7677 (approved):
Very nice!

PR #394, opened 2026-09-08
Update NVAPI headers to R615

Saancreed:
And recognize the setting previously seen only in Streamline code.

There's no new NVML header yet as far as I can tell :(


jp7677 (approved):
Thanks!

PR #396, opened 2026-09-17
nvapi_d3d: implement NvAPI_D3D_SetReflexSync

Hyphaed:
## Problem

`NvAPI_D3D_SetReflexSync` is unimplemented. Any title or middleware that calls
it (Reflex's low-latency sync path, distinct from `NvAPI_D3D_SetSleepMode`)
gets no low-latency behavior on this translation layer, silently.

## Cause

The function was never added to `nvapi_d3d.cpp` or wired into the
`nvapi_interface.cpp` dispatch table, even though the numeric interface ID
(`0xb9f6faff`) already exists in the vendored `nvapi_interface.h` table.

## Fix

Implements `NvAPI_D3D_SetReflexSync` following the exact structure of the
existing `NvAPI_D3D_SetSleepMode`:

- Standard argument/version/registry validation, matching every other
  function in the file.
- Resolves the device's `NvapiD3dLowLatencyDevice` and calls
  `SetLatencySleepMode` with the enable/disable flags, the same call
  `SetSleepMode` already makes.
- `bEnable`/`bDisable` map to the low-latency sleep-mode toggle. The
  struct's dynamic Multi-Frame-Generation pacing fields
  (`vblankIntervalUs`, `timeInQueueUs`, `timeInQueueUsTarget`,
  `fgMultiplier`, `dfgMaxMultiplier`, `dfgTargetFps`) are logged once (info
  level) as received-but-not-driving-anything: there's no public Vulkan
  mechanism in this translation layer to drive NVIDIA's proprietary Dynamic
  MFG pacing, so honoring only the low-latency toggle and saying so
  explicitly seemed better than silently dropping the rest or faking
  support.
- No-ops cleanly (`Ok`) when both `bEnable` and `bDisable` are unset,
  instead of forwarding a call that changes nothing.

## Testing

Added to `tests/nvapi_d3d.cpp`, following the file's existing three-tier
pattern (null-device / no-implementation / success), plus one test asserting
`SetLatencySleepMode` is *not* called when only the unsupported MFG-pacing
fields are set (`FORBID_CALL`), so the "logged but doesn't fake-trigger the
real call" behavior above doesn't regress silently.

Built both `dxvk-nvapi-x86_64` and `dxvk-nvapi-i386` under the project's own
mingw cross-toolchain (zero errors) and confirmed the new log-message string
is present in the stripped, installed `nvapi64.dll` via `strings`.

## Out of scope

Execution-time verification against a real game calling this specific
function hasn't been done yet, this PR is compile- and unit-test-verified,
not run against a live `SetReflexSync` caller. Flagging that explicitly
rather than implying more coverage than exists.


Saancreed:
Hello,

> Any title or middleware that calls it (Reflex's low-latency sync path, distinct from `NvAPI_D3D_SetSleepMode`) gets no low-latency behavior on this translation layer, silently.

Citation needed. It would be the case, if we had an implementation that returns `NVAPI_OK` but doesn't do that the call is expected to do. Which… appears to be what the code in this PR is doing.

> Execution-time verification against a real game calling this specific function hasn't been done yet — this PR is compile- and unit-test-verified, not run against a live `SetReflexSync` caller. Flagging that explicitly rather than implying more coverage than exists.

So, what even is the point of this? You want us to accept a dubious function implementation, for which there are no indicators that it's even slightly correct, and you can't even tell if it helps anything at all?

Hyphaed:
I was checking the injection of DLLS5 Neural Rendering into 007 First Light, and found it as a fix/improvement to do before kept testing...  

Found while testing the OptiScaler DLSS5-NR injection setup (workflow/state/PROGRESS.md) against 007 First Light (IO Interactive, Glacier engine), run under Proton with PROTON_ENABLE_NVAPI=1 + dxvk-nvapi as the NVAPI translation layer. The NVAPI log from a real session showed 29,239 SetReflexSync/SetDynamicMFGParams failures in ~6 minutes, the game retry-storming because the call kept coming back unimplemented.


Why it's useful
- Maps bEnable/bDisable onto the same SetLatencySleepMode path SetSleepMode already drives, real, working low-latency toggle, not a stub.
- The struct's Dynamic MFG pacing fields (vblankIntervalUs, fgMultiplier, etc.) have no public Vulkan hook (VK_NV_low_latency2 only exposes sleep mode / markers / latency queries), logged once as received-but-inert instead of faked.
- Stops the retry storm; games polling this call no longer burn cycles hitting a guaranteed failure.

Yet, I haven't retested again

I also have some other tests to go trough...  related to Proton an specific optimizations for  i9-14900KF hybrid P/E-core scheduling (PROTON_CPU_TOPOLOGY, SCHED_BATCH fallback, E-core, steering)

Don't know when I will be back to that... I'm busy those days... when I opened the PR I was just bored, testing DLSS5 Neural Rendering injection + possible optimizations related to gaming on my specific hardware topology while keeping in consideration logics should be dynamic and intelligent to adapt to other hardware

Saancreed:
But why does this code do precisely what you state the problem to be? You have stated that middlewares will silently get unexpected behavior. It's obviously not the case if 29239 failures were correctly observed and even logged. With your version, the application will see success being returned from the NVAPI call, but it is not possible to fully implement the desired behavior with what extensions we have in Vulkan today.

Returning success while discarding some parts of the request will almost certainly break more stuff than it will help with. It's precisely why we currently don't have any implementation for functions like `NvAPI_D3D12_SetFlipConfig`, even though we know that Dynamic MFG wants to use it. And, by extension, why this code is not acceptable in its current form.

PR #397, opened 2026-09-19
github: Set working directory as safe for git in container

jp7677:
Also add an extra step that prints that info (and a bit more). Additionally, that step fails when git denies. 

Follow up from https://github.com/jp7677/dxvk-nvapi/pull/389#issuecomment-5694906885 and thus https://github.com/doitsujin/dxvk/pull/5905


.github/workflows/build-llvm.yml
```diff
@@ -42,6 +42,9 @@ jobs:
         test -e ./Include || mv ./include ./Include
         test -e ./Lib || mv ./lib ./Lib
 
+    - name: Print build-in version
```
Saancreed: Phrasing is a little awkward, How about

```suggestion
    - name: Print build version
```

or

```suggestion
    - name: Print current version
```

? Same below.

Saancreed (approved):

